Issue #50, opened 2023-10-18
Arweave Storage type


alexcos20:
Write a storage class for arweave storage type,
Needs ARWEAVE_GATEWAY gateway defined

Issue #42, opened 2023-10-18
IPFS Storage type


alexcos20:
Write a storage class for ipfs storage type,

Depending on the config, based an ENV var, it should:

- [ ]  use an ipfs server
- [ ]  use a ipfs gateway

Issue #34, opened 2023-10-18
URL Storage type


alexcos20:
Write a storage class for url type

PR #94, opened 2023-11-06
Storage base class

### LINKED ISSUE #35: Storage interface
As Ocean Node will have to deal with multiple storage types, we need a structure for that.

Have a base storage class, that exposes standard functions for all storage types, and it's initialized with a file object  (see https://docs.oceanprotocol.com/developers/storage)

As Ocean Node works only with streams, this class needs to support them

Once we have this, we can start writing storage classes for each type

mariacarmina:
Fixes #35  #34 #42 #50 .

Changes proposed in this PR:

- create the Storage base class
- create & implement children classes
- create tests


mariacarmina:
Commented tests in favour of mocha framework.

paulo-ocean (changes requested):
Hi @mariacarmina 
couple of comments
You have 4 common functions (please check my previous comments on this :-)  ) :

```
getFile():
validate(): 
getDownloadUrl():
getReadableStream():
```
This could be easily an interface. And your subclasses (ArweaveStorage, IpfsStorage,  UrlStorage) could just implement this interface. There is no really point in having the 4 "dummy" implementations on the main "Storage" class

Also, your ` getFile()` implementations are checking for the correct instanceOf 
Example:
```
getFile(): UrlFileObject {
    if (this.getFile() instanceof UrlStorage) {
      return this.getFile()
    }
    throw new Error(`Invalid storage type for this method`)
  }
```
Not really sure what you're trying to do, IMO, it looks a bit confusing.. but if you want to check the correct file type, why not do it on the constructor itself. Seems to me your getFile() its calling itself and looking for a different thing than the `file` type passed on the constructor ?

mariacarmina:
> Hi @mariacarmina couple of comments You have 4 common functions (please check my previous comments on this :-) ) :
> 
> ```
> getFile():
> validate(): 
> getDownloadUrl():
> getReadableStream():
> ```
> 
> This could be easily an interface. And your subclasses (ArweaveStorage, IpfsStorage, UrlStorage) could just implement this interface. There is no really point in having the 4 "dummy" implementations on the main "Storage" class
> 
> Also, your ` getFile()` implementations are checking for the correct instanceOf Example:
> 
> ```
> getFile(): UrlFileObject {
>     if (this.getFile() instanceof UrlStorage) {
>       return this.getFile()
>     }
>     throw new Error(`Invalid storage type for this method`)
>   }
> ```
> 
> Not really sure what you're trying to do, IMO, it looks a bit confusing.. but if you want to check the correct file type, why not do it on the constructor itself. Seems to me your getFile() its calling itself and looking for a different thing than the `file` type passed on the constructor ? I'm a bit confused here

I did the validation in the constructor for children classes, I'll remove from `getFile()`. I can create a separate interface for those methods and make `Storage` abstract.

paulo-ocean:
hi @mariacarmina , looks better, but i think there are a few things that can be improved :-)
- You don't need to have `getFile()` implemented on all subclasses just to call` super.getFile()`.
One concrete implementation on the abstract class, would be enough.. the others are not really necessary.

- Your `async getReadableStream(): Promise<Readable>` implementations are pretty much the same on all subclasses. Does an abstract method makes sense here if they are all the same?, probably you don't need an abstract method for this as well
- Another aspect is the` process.env` vars. If they are required, i think its better to check them on the constructor, and log/throw error if they are missing (see how its done for private key and rpcs, for instance)
thanks

mariacarmina:
> * Double check your imports, for instance, when importing from a relative path we might need to add the file extensions (.js). Otherwise the ESM modules do not load (try npm run build & npm run start) and you'll see..
>   An example:` import { getFileFromURL } from '../core/downloadHandler'` does not work, it needs to be:
>   `import { getFileFromURL } from '../core/downloadHandler.js'`
>   thanks

Hello @paulo-ocean, dowloadHandler is a .ts file
Screenshot 2023-11-14 at 14 23 52

mariacarmina:
> * Your `async getReadableStream(): Promise<Readable>` implementations are pretty much the same on all subclasses. Does an abstract method makes sense here if they are all the same?, probably you don't need an abstract method for this as well

@alexcos20 mentioned in previous comments that this method need to throw a specific error depending on the storage type and should not be implemented only in the parent class and have conditional instances like. I think we can leave this method as abstract for the moment if @alexcos20 has not anything to add.

mariacarmina:
> * Another aspect is the` process.env` vars. If they are required, i think its better to check them on the constructor, and log/throw error if they are missing (see how its done for private key and rpcs, for instance)

I think `validate()` can include `process.env`, because is also called inside constructor and we can delete the validation condition from `getDownloadUrl()`. Makes sense?

paulo-ocean:
> > * Double check your imports, for instance, when importing from a relative path we might need to add the file extensions (.js). Otherwise the ESM modules do not load (try npm run build & npm run start) and you'll see..
> >   An example:` import { getFileFromURL } from '../core/downloadHandler'` does not work, it needs to be:
> >   `import { getFileFromURL } from '../core/downloadHandler.js'`
> >   thanks
> 
> Hello @paulo-ocean, dowloadHandler is a .ts file Screenshot 2023-11-14 at 14 23 52

yes, but not when you build it :-) we had this conversation about node v20, relative paths, and ES modules a couple of weeks ago

paulo-ocean:
@mariacarmina  your tests still break :-(, similar to what was happening to Illia on his branch... probably related with imports..
Do we really need all these types spread to multiple files. I think @alexcos20 also mentioned something about this
```
import { ArweaveFileObject } from '../../@types/arweaveFileObject'
import { IpfsFileObject } from '../../@types/ipfsFileObject'
import { UrlFileObject } from '../../@types/urlFileObject'
```
tkx

mariacarmina:
> > > * Double check your imports, for instance, when importing from a relative path we might need to add the file extensions (.js). Otherwise the ESM modules do not load (try npm run build & npm run start) and you'll see..
> > >   An example:` import { getFileFromURL } from '../core/downloadHandler'` does not work, it needs to be:
> > >   `import { getFileFromURL } from '../core/downloadHandler.js'`
> > >   thanks
> > 
> > 
> > Hello @paulo-ocean, dowloadHandler is a .ts file Screenshot 2023-11-14 at 14 23 52
> 
> yes, but not when you build it :-) we had this conversation about node v20, relative paths, and ES modules a couple of weeks ago, so i would not rely on that


After `npm run build` command. I see that there are both extensions.
Screenshot 2023-11-14 at 16 20 32

mariacarmina:
> @mariacarmina your tests still break :-(, similar to what was happening to Illia on his branch... probably related with imports.. Do we really need all these types spread to multiple files. I think @alexcos20 also mentioned something about this
> 
> ```
> import { ArweaveFileObject } from '../../@types/arweaveFileObject'
> import { IpfsFileObject } from '../../@types/ipfsFileObject'
> import { UrlFileObject } from '../../@types/urlFileObject'
> ```
> 
> IMO, these types/interfaces could be on the same file, as they relate tkx

I can group the interfaces in the same file.

paulo-ocean:
> > > > * Double check your imports, for instance, when importing from a relative path we might need to add the file extensions (.js). Otherwise the ESM modules do not load (try npm run build & npm run start) and you'll see..
> > > >   An example:` import { getFileFromURL } from '../core/downloadHandler'` does not work, it needs to be:
> > > >   `import { getFileFromURL } from '../core/downloadHandler.js'`
> > > >   thanks
> > > 
> > > 
> > > Hello @paulo-ocean, dowloadHandler is a .ts file Screenshot 2023-11-14 at 14 23 52
> > 
> > 
> > yes, but not when you build it :-) we had this conversation about node v20, relative paths, and ES modules a couple of weeks ago, so i would not rely on that
> 
> After `npm run build` command. I see that there are both extensions. Screenshot 2023-11-14 at 16 20 32

do you see the `downloadHandler.ts` ? i don't :-)

mariacarmina:
> > > > > * Double check your imports, for instance, when importing from a relative path we might need to add the file extensions (.js). Otherwise the ESM modules do not load (try npm run build & npm run start) and you'll see..
> > > > >   An example:` import { getFileFromURL } from '../core/downloadHandler'` does not work, it needs to be:
> > > > >   `import { getFileFromURL } from '../core/downloadHandler.js'`
> > > > >   thanks
> > > > 
> > > > 
> > > > Hello @paulo-ocean, dowloadHandler is a .ts file Screenshot 2023-11-14 at 14 23 52
> > > 
> > > 
> > > yes, but not when you build it :-) we had this conversation about node v20, relative paths, and ES modules a couple of weeks ago, so i would not rely on that
> > 
> > 
> > After `npm run build` command. I see that there are both extensions. Screenshot 2023-11-14 at 16 20 32
> 
> do you see the `downloadHandler.ts` ? i don't :-)

No, it is only `downloadHandler.d.ts`, I refactored the import locally.

paulo-ocean (approved):
Overall LGTM
your 
`static getStorageClass(file: any): UrlStorage | IpfsStorage | ArweaveStorage {`
could be just
`static getStorageClass(file: any): Storage {`
as everything is extending that, and we probably should expect some error handling on the stream request, but i don't think its a deal breaker at this moment :+1:

src/components/storage/index.ts
```diff
@@ -0,0 +1,144 @@
...
+
+  abstract validate(): [boolean, string]
+
+  abstract getDownloadUrl(): string
+
+  getFile(): any {
+    return this.file
+  }
+
+  async getReadableStream(): Promise<Readable> {
```
alexcos20, line 24: can we move this out of Storage class?
Not all storage types will act as an url (think sql, azure blob, etc)
Even if we repeat ourselfs, for now I would have it in each of URL/IPFS/Arweave and deal with that later  (When we will have native IPFS connection, we don't need to use the gateway)

mariacarmina: Ok, so I'll move this function for the moment in each child class and remove from the parent class (`Storage`). Thanks.

mariacarmina:
> Overall LGTM your `static getStorageClass(file: any): UrlStorage | IpfsStorage | ArweaveStorage {` could be just `static getStorageClass(file: any): Storage {` as everything is extending that, and we probably should expect some error handling on the stream request, but i don't think its a deal breaker at this moment 👍

For the moment, we cannot do that, because `async getReadableStream` will be specific only in the children classes from @alexcos20 comment.

alexcos20 (approved):

PR #101, opened 2023-11-08
Storing logs in database

### LINKED ISSUE #67: Store logs in database
- [ ] define logs mapping
- [ ] use one collection to store them all
- [ ] each component has its own tag(component)
- [ ] prune logs after x time, configurable from the env config

### LINKED ISSUE #69: Expose logs endpoint
The endpoint exposes the logs stored in the database 
Parameters:
```
{
"component": "provider",
"logLevel": "info/error/critical",
"startDate": "736784367",
"endDate": "8784", //Optional, if not available take the current time
"compressed": true/false
}
```

Response: 
File content (zip or not) with accurate headers

jamiehewitt15:
Fixes #67 
Fixes #69 
Fixes: #109 

Changes proposed in this PR:

- Storing logs in database
- Creating a custom transport for storing the logs in the Typsense database
- Updating the schema to include the logs
- Adding methods for retrieving individual logs and multiple logs
- Adding HTTP routes to retrieve the logs
- Adding tests to check that the logs are successfully saved in the database.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,7 +140,7 @@ npm run check-nonce
 
 (Purpose: for checking nonce tracking flow. This last one requires DB up and running)
 
-## Log Retrieval
+## Log Retrieval using HTTP
 
 ### Get Multiple Logs
 
@@ -212,3 +212,29 @@ GET /log/123456789
 ```
 
 If the log with the given ID is not found, you will receive a `404 Not Found` response. For server errors, you will receive a `500 Internal Server Error` response.
+
+## Log Retrieval Using Script
+
+The logging system provides a convenient way to retrieve logs via a command-line script. The script is capable of fetching logs with various filters, such as start time, end time, maximum number of logs, module name, and log level.
+
+**Usage**
+You can call the script directly from your command line with optional parameters to filter the logs. The parameters are as follows:
+
+- `API_URL`: The URL of the logs API endpoint. Defaults to http://localhost:8000.
+- `START_TIME`: The start time for the logs you want to retrieve. Defaults to 24 hours before the current time.
+- `END_TIME`: The end time for the logs you want to retrieve. Defaults to the current time.
+- `MAX_LOGS`: The maximum number of logs to retrieve. Defaults to 100.
+- `MODULE_NAME`: The specific module name to filter the logs. Optional.
+- `LEVEL`: The specific log level to filter the logs. Optional.
+
+**Example Without Parameters (Uses Defaults):**
+
+```bash
+npm run logs
+```
+
+**Example With Specific Parameters:**
+
+```
+npm run logs http://localhost:8000 "2023-11-01T00:00:00Z" "2023-11-30T23:59:59Z" 50 "http" "info"
+```
--- a/package.json
+++ b/package.json
@@ -21,7 +21,8 @@
     "test:unit:storage": "npm run mocha -- 'tests/unit/storage.test.ts'",
     "test:integration": "npm run mocha -- 'tests/integration/**/*.test.ts'",
     "test:unit:cover": "nyc --report-dir coverage/unit npm run test:unit",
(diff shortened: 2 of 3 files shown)


alexcos20 (changes requested):
Can we add module name as well?

paulo-ocean (commented):
Any particular reason why we're just using console.log() here on this class ? 
like
console.log(`Schema for '${this.schema.name}' collection already exists.`)
If we agreed that we should avoid these on other places,  i think we certainly we should avoid it here as well ..

paulo-ocean:
hey @jamiehewitt15 
i think it could be interesting that we could make this part (on index.ts) :
```
const customLogTransport = new CustomOceanNodesTransport({ dbInstance: dbconn })
  logger.addTransport(customLogTransport)
```
Somehow available as a function that we could call from other places (or maybe it have it exported on some other place)..
So, we could add this new transport on other loggers, and not only that one on index.ts..
Right now, all the other "loggers" instances on other files, are only logging to the console transport... if we could construct this db transport from from other places, it would be very easy to put something we want... logging into console and DB at same

paulo-ocean:
hi @jamiehewitt15 

I was not referring to export this logger (this is associated with HTTP component) , although there is nothing wrong in export it
```
**export const logger**: CustomNodeLogger = getCustomLoggerForModule(
  LOGGER_MODULE_NAMES.HTTP,
  LOG_LEVELS_STR.LEVEL_INFO, // Info level
  defaultConsoleTransport // console only Transport
)
```
but actually having something like:
```
export function getCustomDBTransport(dbconn: Database) {
  return new CustomOceanNodesTransport({ dbInstance: dbconn })
}
```
(not exactly sure what would be the best file to put this, but..)

so you could, for instance just do (on index.ts) :

```
...
if (config.hasIndexer) indexer = new OceanIndexer(dbconn)
  if (config.hasProvider) provider = new OceanProvider(dbconn)
  **const customLogTransport = getCustomDBTransport(dbconn)
  logger.addTransport(customLogTransport)**
...
```
And for instance on any other component (lets take the P2P index as example)
where we have another custom logger:

```
export const P2P_CONSOLE_LOGGER: CustomNodeLogger = getCustomLoggerForModule(
  LOGGER_MODULE_NAMES.P2P,
  LOG_LEVELS_STR.LEVEL_INFO,
  defaultConsoleTransport
)
```
we could get that custom transport and add it to the P2P_CONSOLE_LOGGER transports array, together with the defaultConsoleTransport . Same thing on other files , that have custom loggers as well.....

alexcos20 (approved):

src/components/database/index.ts
```diff
@@ -233,16 +233,102 @@ export class IndexerDatabase {
...
+    return (async (): Promise<LogDatabase> => {
+      this.provider = new Typesense(convertTypesenseConfig(this.config.url))
+      try {
+        await this.provider.collections(this.schema.name).retrieve()
+      } catch (error) {
+        if (error instanceof TypesenseError && error.httpStatus === 404) {
+          try {
+            await this.provider.collections().create(this.schema)
+          } catch (creationError) {
+            // logger.log(
```
mariacarmina, line 252: Shouldn't we log the error here?

mariacarmina (commented):
Just a minor Q. Overall, looks good.

paulo-ocean (approved):
LGTM :+1: 
maybe on future iterations we can also have a logger instance on the database module, as we're not logging much of the typesense possible error stuff

jamiehewitt15:
Thanks for the feedback @mariacarmina and @paulo-ocean, you're both right we should have a lot more logs in all of the different components. I'll create a separate PR for that as this one is already getting quite big  (issue here: https://github.com/oceanprotocol/ocean-node/issues/127)

Issue #48, opened 2023-10-18
Core:  Database functionality


alexcos20:
Build on top of oceanprotocol/ocean-node#29, this will implement database functionality for all components & modules.

TO DO:

- [ ] implement CRUD for DDOs
- [ ] implement CRUD for indexer internal stats (last_indexed_block, etc)

- [ ] implement CRUD for user nonces

PR #102, opened 2023-11-08
Feature nonce tracking

### LINKED ISSUE #83: Nonce tracking
When an user interacts with a node, for some actions, it needs to identify itself.
In order to that, user will send a signed message and a nonce.

User's actions:
   - create a message = String(nonce)
   - sign it
   - when calling ocean-node endpoints that do required auth, add nonce and signature as parameters

Ocean-node will check:
 - user's nonce >  previous stored nonce
 - signer of message matches user's address (consumer parameter)
 
  

Example:
GET /endpoint?param1=x&param2=y&nonce=1&signature=0x123&consumer=0x456

TO DO:
- [x]  implement nonce related function in core db  (getNonce(address) ,setNonce(address,newNonce)
- [x]  create a core function nonce(address), that returns stored nonce for an address  ( 0 if not found)
- [x]  expose nonce command to http
- [x]  expose nonce command to libp2p
- [x]  create a core function:  check_nonce(consumer,nonce,signature) that:
     - checks nonce
     - updates new nonce in db
     - returns true/false (+ error message if needed)

paulo-ocean:
Fixes #83 .

Changes proposed in this PR:

- core db functions (get, set and update nonce) - eventually will refactored due to #48 
- expose nonce (get) command to http and p2p
```
 {
    "command":"nonce",
    "address": "0x8F292046bb73595A978F4e7A131b4EBd03A15e8a"
    "node"?
}
```
- check nonce (validate against previous nonce + check signature (client signed nonce) + update new nonce)
- there is a script to test the flow (under `helpers/scripts/checkNonceTracking`) ( npm run check-nonce ) and unit tests under` tests/units/nonce.spec.ts` . However, due to issues with jest (and imports), they are mutually exclusive (to run unit tests we need to adjust some imports) and these tests are not much useful at the moment cause we can´t import everything we need.

So, for now better just do` npm run check-nonce` , until we sort the test frameworks issues

Obs: Main core functionality is handled on `components/core/nonceHandler.ts`


src/components/core/nonceHandler.ts
```diff
@@ -82,11 +83,12 @@ export async function getNonce(address: string): Promise<P2PCommandResponse> {
 async function updateNonce(address: string, nonce: number): Promise<NonceResponse> {
```
mariacarmina, line 83: Before updating the nonce, can we check if the nonce is valid. In the previous logic we had a check before updating the nonce. Are we still keeping the logic of comparing the previous nonce with the actual nonce @alexcos20 ? If yes, I think we can include a check in this function as well. Thanks.

mariacarmina: Ok, I understand, from my side I would've called `checkNonce` inside `updateNonce`, because `checkNonce` should only do validation for current nonce (as from the name of the function), not including the updating the nonce part as well. That's my suggestion regarding this approach.

paulo-ocean: Sorry, but i have to disagree :-)
The functions names expose exactly what they do.
From the requirements #83 , we need to expose the nonce (`getNonce()`) on HTTP/P2P, and have a function `checkNonce()` to validate the nonce/signature and eventually update in on DB (this function will be called by any endpoint that requires this verification). 
Those are the only 2 functions exposed to other modules..
You either get the current nonce (called maybe from a client) OR
you call some endpoint that requires nonce/signature validation...
In this scenario, you check the given nonce, validate it against previous one and signature and update it on DB, returning the status, exactly as it is on the requirements

mariacarmina: Ok, based on the requirements from the issue, you reflected the functionality accordingly. But from my point of view, the issue specs should be written like this:
`checkNonce` should only do the validation part, not eventually updating in the db. I don't think that the spec will be modified for this aspect, otherwise all good from my side.

mariacarmina: Based on Slack discussion, can we rename `checkNonce` into `checkAndUpdateNonce`? Since `checkNonce` also includes the updating part.

alexcos20 (approved):

bogdanfazakas (approved):
lgtm, we can merge it as it is, also if not handled in the pr with refactors maybe worth checking the the lint warnings from the changed files even if them were not introduced in the pr(like unused imports etc)

PR #103, opened 2023-11-09
Indexer core

### LINKED ISSUE #21: Indexer core
Indexer core is a critical component of Ocean Node, and it's responsible of crawling several chains in the same time , and index different events and then process them.

To Do:

- [ ]  create config and take list of :  RPC,chainId,chunkSize from env for every network
- [ ]  for each network, start a working thread  (use nodejs Working Threads)

Each thread should:
 - [ ]  should detect the block used to deploy the contracts from addresses.json 
 - [ ]  should store the last indexed block. if none, use above
 - [ ]  check network height, if > than last indexed block, process remaining blocks in chunks.  If something fails, reduce chunk size to 1/2 and retry

Process function:
 - [ ]  based on event, use a specific processor  (each event will be detailed in a separate issue)
 - [ ]  if something fails or processing of an event takes more that than a couple of secs, insert that event to a retry queue and move to next one
 
 Development flow:  have a branch called "Indexer core" and then small PR into that branch.   Once all ticks are checked, will merge that branch into develop

 (Note: check out https://github.com/oceanprotocol/aquarius/blob/main/aquarius/events/events_monitor.py, it has pretty much the same principles)

bogdanfazakas:
Fixes #21 .

Changes proposed in this PR:

- [x] create config and take list of : RPC,chainId,chunkSize from env for every network,
  need to add **chunkSize**
- [x] for each network, start a working thread (use nodejs Working Threads)
- [x] should detect the block used to deploy the contracts from addresses.json
- [x] should store the last indexed block. if none, use above
   need db interaction to save the last indexed block and get the last indexed block
- [x] check network height, if > than last indexed block, process remaining blocks in chunks. If something fails, reduce chunk size to 1/2 and retry
- [x] based on event, use a specific processor
- [x] if something fails or processing of an event takes more that than a couple of secs, insert that event to a retry queue and move to next one


src/components/Indexer/crawlerThread.ts
```diff
@@ -0,0 +1,82 @@
...
+  let startBlock =
+    lastIndexedBlock && lastIndexedBlock > deployedBlock
+      ? lastIndexedBlock
+      : deployedBlock
+
+  INDEXER_LOGGER.logMessage(
+    `network: ${rpcDetails.network} Start block ${startBlock} network height ${networkHeight}`
+  )
+
+  if (networkHeight > startBlock) {
```
alexcos20, line 43: what happends once we processed all blocks?
Thread should be sleeping, and from time to time get the new height and process more blocks.

From your code(let me know if I'm wrong) looks like it's going to process once and then stop.

alexcos20 (changes requested):

src/components/Indexer/utils.ts
```diff
@@ -0,0 +1,143 @@
...
+  return deployedBlock
+}
+
+export const getNetworkHeight = async (provider: ethers.Provider) => {
+  const networkHeight = await provider.getBlockNumber()
+
+  return networkHeight
+}
+
+export const processBlocks = async (
```
alexcos20, line 45: also, process block by block is going to be a pain
we need to filter a chunk of blocks by events, using provider.getLogs

src/components/Indexer/crawlerThread.ts
bogdanfazakas: good catch, thanks forgot to update the heigth in a timeout, will push that

src/components/Indexer/utils.ts
```diff
@@ -0,0 +1,143 @@
...
+  LOGGER_MODULE_NAMES.DATABASE,
+  LOG_LEVELS_STR.LEVEL_INFO,
+  defaultConsoleTransport
+)
+
+export const getDeployedContractBlock = async (network: number) => {
+  let deployedBlock: number
+  const addressFile = JSON.parse(
+    // eslint-disable-next-line security/detect-non-literal-fs-filename
+    fs.readFileSync(
```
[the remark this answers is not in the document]
bogdanfazakas: got it so we should add an address.json file and fallback to that , i think aqua is doing something similar atm

paulo-ocean (changes requested):
Hi @bogdanfazakas 
Just a couple of suggestions as i think you could make a little bit better use of our custom logger :-)
for instance, when you create the custom Logger you could specify the module as 'Indexer, rather than 'Database'
```
export const INDEXER_LOGGER: CustomNodeLogger = getCustomLoggerForModule(
  LOGGER_MODULE_NAMES.DATABASE, // INDEXER
  LOG_LEVELS_STR.LEVEL_INFO,
  defaultConsoleTransport
)
```
And when you're logging, you could set the log level directly by calling '`log(level,message,includeModuleName)`' // include module name
instead of always setting log level before and after the message, like:

```
INDEXER_LOGGER.loggerOptions.level = LOG_LEVELS_STR.LEVEl_ERROR
INDEXER_LOGGER.logMessage('Error retrieving last indexed block')
INDEXER_LOGGER.loggerOptions.level = LOG_LEVELS_STR.LEVEL_INFO
```

thanks :+1:

bogdanfazakas:
> Hi @bogdanfazakas Just a couple of suggestions as i think you could make a little bit better use of our custom logger :-) for instance, when you create the custom Logger you could specify the module as 'Indexer, rather than 'Database'
> 
> ```
> export const INDEXER_LOGGER: CustomNodeLogger = getCustomLoggerForModule(
>   LOGGER_MODULE_NAMES.DATABASE, // INDEXER
>   LOG_LEVELS_STR.LEVEL_INFO,
>   defaultConsoleTransport
> )
> ```
> 
> And when you're logging, you could set the log level directly by calling '`log(level,message,includeModuleName)`' // include module name instead of always setting log level before and after the message, like:
> 
> ```
> INDEXER_LOGGER.loggerOptions.level = LOG_LEVELS_STR.LEVEl_ERROR
> INDEXER_LOGGER.logMessage('Error retrieving last indexed block')
> INDEXER_LOGGER.loggerOptions.level = LOG_LEVELS_STR.LEVEL_INFO
> ```
> 
> thanks 👍

thanks for the hints, will update

src/components/Indexer/utils.ts
```diff
@@ -0,0 +1,131 @@
...
+  return networkHeight
+}
+
+export const processBlocks = async (
+  provider: ethers.Provider,
+  startIndex: number,
+  count: number
+): Promise<ProcessingEvents> => {
+  try {
+    const blockLogs = await provider.getLogs({
```
alexcos20, line 48: We need to add all topics that we are looking for, otherwise we will get all logs, from all contracts on the network :)

src/utils/address.ts
```diff
@@ -0,0 +1,76 @@
+export const addressFile = {
```
alexcos20, line 1: why don't you "npm i @oceanprotocol/contracts" ? :)

bogdanfazakas: aa ok i think missundertood the initial comment 🤦 , like not to use ocean-contracts 😄 
will update :)

src/components/Indexer/utils.ts
bogdanfazakas: got it, finally, so we should filter a chunk of blocks with topics that are hashes of the events we want to catch 😃

src/utils/address.ts
alexcos20: we will need contracts anyway for events, provider functionality, etc.. so let's have it now

alexcos20 (approved):

paulo-ocean (approved):
alright :+1:

PR #104, opened 2023-11-09
Branch setup mocha tests

paulo-ocean:
Fixes # .

Changes proposed in this PR:

- add mocha as test framework

It seems to be working properly. We have failing tests, that we will have to fix,
but the issues with the import of the modules apparently are gone.
Needs a little more check, but i think we're on the right track

UPDATE: these tests are all passing now

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -45,21 +45,22 @@
         "winston-transport": "^4.6.0"
       },
       "devDependencies": {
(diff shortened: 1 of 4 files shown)


mariacarmina (approved):
LGTM

bogdanfazakas (approved):
lgtm, i've run the tests suite also locally and all seems to be fine

PR #116, opened 2023-11-15
Feature findDDO

### LINKED ISSUE #107: GetDDO Broadcast

paulo-ocean:
Fixes #107.

Changes proposed in this PR:

- Send command 'findDDO' to the network, given the identifier
- Using cache and/or DHT content routing we get info about a specific DDO & providers serving it

- The node previously store and advertise whatever is serving to the network (could be on startup or when indexing locally) `storeAndAdvertiseDDOS(list: any[])`
-  Node keeps a cached version (the most updated one) of each DDO he is aware of (and which provider serves it)

- If node has cached version (and still fresh) it responds with that (cache expires after 5 min)
- If cache id 'old' or it does not have any info locally it queries the network. It calls` getProvidersForDid(did)`, processes list of results (sends command `getDDO` on each), and updates cache, keeping only the most recent one

- Node can respond with a list of known providers for the DDO, ordered by update time (up to 5 max, where 1st on the list is always the most recent), or just with the cached version (which is always the newest one he knows about).
If node knows nothing and does not get any info from other nodes either, it just sends an empty list []

example request:
```
{
    "command": "findDDO",
    "id": "did:op:fa0e8fa9550e8eb13392d6eeb9ba9f8111801b332c8d2345b350b3bc66b379d5"
}
(as with other commands, we can also use the optional "node" param, to proxy requests)
```
example response:
```
[
    {
        "id": "did:op:fa0e8fa9550e8eb13392d6eeb9ba9f8111801b332c8d2345b350b3bc66b379d5",
        "lastUpdateTx": "0xceb617f13a8db82ba9ef24efcee72e90d162915fd702f07ac6012427c31ac952",
        "lastUpdateTime": "2022-12-30T08:40:06Z",
        "provider": "16Uiu2HAkuYfgjXoGcSSLSpRPD6XtUgV71t5RqmTmcqdbmrWY9MJo"
    }
]
```
Obs: if you set the env var 
`export LOAD_INITIAL_DDOS=true`
the node will load the DDO json examples, inside the **/data** folder, on startup.. and will store/advertise them to the network


alexcos20 (approved):
awesome!
Might need some small cleanup after indexer is merged, but will do that in another issue

src/components/P2P/index.ts
```diff
@@ -11,21 +11,16 @@ import {
...
   handlePeerDiscovery,
   handlePeerDisconnect,
-  handlePeerJoined,
-  handlePeerLeft,
-  handleSubscriptionCHange,
   handleProtocolCommands,
   handleDirectProtocolCommand
 } from './handlers.js'
 
 // import { encoding } from './connection'
```
jamiehewitt15, line 18: Is there any particular reason that we are leaving these comments here? There are also more on lines 3 and 4 that could be removed.

src/components/P2P/index.ts
```diff
@@ -440,12 +453,59 @@ export class OceanP2P extends EventEmitter {
...
+        } catch (e) {
+          console.log(e)
+        }
+      })
+      if (count > 0) {
+        this._ddoDHT.updated = new Date().getTime()
+      }
+      return count === list.length
+    } catch (err) {
+      console.log(err)
```
jamiehewitt15, line 505: It would be better to use the Logger class rather than `console.log` that way the logs will also saved in the database. The same also applies to all of the other `console.log` lines in this PR

jamiehewitt15:
It would be good to see some tests added for this, was there a reason why you didn't add any?

src/components/core/ddoHandler.ts
```diff
@@ -24,3 +39,286 @@ export async function handleGetDdoCommand(
...
+    )
+  }
+  return false
+}
+/**
+ * Get the DDO list from list of available providers (including self)
+ * @param task the findDDO command task
+ * @returns filtered list
+ */
+export async function findDDO(task: FindDDOCommand): Promise<P2PCommandResponse> {
```
jamiehewitt15, line 72: This function is over 200 lines long, it will be easier to maintain if it is shorter and easier to read. The functions defined within this (such as `sink`, `sortedResults` & `fnTimeout`) can be moved out to keep the findDDO function shorter and more manageable. Also consider if all of these functions need to be defined within this file, something like `fnTimeout` could be moved to a utils file as it might get used elsewhere

paulo-ocean: hi @jamiehewitt15 ,  
thanks
i definitely agree its long, but unfortunately what you're saying cannot be applied here.
None of that could be easily moved out.
- If you move the `sink`  out, how do you process the different streamed responses and the corresponding final API response? 

- The timeout is to avoid that we got stuck somewhere in the loop if there is any issue/delay with any of the remote providers, and still reply a valid response if that happens.. What do you want to put on utils file exactly? We would still need to return a meaningful response for this specific API call

- Same for the `sortedResults`. It sorts this specific list, by lastUpdateTime .. What you you put on utils? A generic sort function? but each sort will be different, depending on the context. So in this case you would be passing around specific data from the API call and make it generic as an utility?

paulo-ocean: fyi, move the sort list out

jamiehewitt15: Ok, yeah that's fair enough, I don't think we should create loads of extra work here

jamiehewitt15 (approved):
Nice, looks good to me

Issue #24, opened 2023-10-10
Command: query


AnaLoznianu:
This will allow searching through existing DDOs stored on a node

TO DO:

- [ ]  implement core function search (it has to search only in collections defined by #19, AND NOT in node private collections (like users noces, indexer, etc)
- [ ]  expose it on http interface
- [ ]  expose it on p2p interface

Command description:

{
command: "query",
node: "NodeID", // optional, if missing, it refers to current node
query: Object
}

PR #121, opened 2023-11-20
PR for Command: query

an7e:
PR for #24

Changes proposed in this PR:

- added command query for ddo
- added query handler
- added search method in DdoDatabase


paulo-ocean (commented):
hi @an7e 
how about about start adding some logs man :-) ?
i mean, the flow is very straightforward and we either get a successful response or an error response, but we're not tracking much about the requests. If we need to debug some stuff, at a later stage, we don't have much to inspect .. 
for instance we don't even know what was the original 'query' and if the request fails at DB level, we don't know exactly what was the error 'catched'  :-) 
Example: Was a search(..) error or was just that the DB was down?   :-)
tkx

src/components/database/index.ts
```diff
@@ -24,6 +25,17 @@ export class DdoDatabase {
     })() as unknown as DdoDatabase
   }
 
+  async search(query: Record<string, any>) {
+    try {
+      return await this.provider
+        .collections(this.schemas[0].name)
```
alexcos20, line 31: can you add all schemas here? (a simple loop)
I know that we only have one, but it will make our life easier when we add more, so we don't need to edit this anymore

paulo-ocean (changes requested):
You forgot to update **SUPPORTED_PROTOCOL_COMMANDS**
on utils/constants.ts
```
export const SUPPORTED_PROTOCOL_COMMANDS: string[] = [
  PROTOCOL_COMMANDS.DOWNLOAD_URL,
  PROTOCOL_COMMANDS.ECHO,
  PROTOCOL_COMMANDS.GET_DDO,
  PROTOCOL_COMMANDS.NONCE
]
```
Your command is still not supported :-)

As another side note, what is exactly a valid 'query' ? i see you have ` search(query as TypesenseSearchParams)` at db level. but if i was calling this API without any knowledge of the DB or typesense (and specially without any logs), as an external developer.. how do i figure out what i'm doing wrong or what exactly should my query look like? :-)

For instance, if the search query needs to have a mandatory parameter  **'q'** , you could do that validation sooner, at the `validateCommandAPIParameters` instead of going deeper into the DB level
 
thanks

an7e:
> You forgot to update **SUPPORTED_PROTOCOL_COMMANDS** on utils/constants.ts
> 
> ```
> export const SUPPORTED_PROTOCOL_COMMANDS: string[] = [
>   PROTOCOL_COMMANDS.DOWNLOAD_URL,
>   PROTOCOL_COMMANDS.ECHO,
>   PROTOCOL_COMMANDS.GET_DDO,
>   PROTOCOL_COMMANDS.NONCE
> ]
> ```
> 
> Your command is still not supported :-)
> 
> As another side note, what is exactly a valid 'query' ? i see you have ` search(query as TypesenseSearchParams)` at db level. but if i was calling this API without any knowledge of the DB or typesense (and specially without any logs), as an external developer.. how do i figure out what i'm doing wrong or what exactly should my query look like? :-)
> 
> For instance, if the search query needs to have a mandatory parameter **'q'** , you could do that validation sooner, at the `validateCommandAPIParameters` instead of going deeper into the DB level
> 
> thanks

@paulo-ocean thanks for noticing :)
about errors, don't worry, because the typesense driver itself has a ton of logs, so when there is a typesense error, we will see it
about validation, I didn't do it on purpose for the sake of simplicity, but it will be done later when we deal with chacl

src/components/database/index.ts
an7e: I added a simple loop, of course it is good to do the same for other methods, but then there is a problem of identification and validation, for example, in the create method in which of the collections should be created an object

paulo-ocean (approved):
LGTM :+1:

alexcos20 (approved):

bogdanfazakas (approved):

PR #122, opened 2023-11-21
OceanNode class

### LINKED ISSUE #117: OceanNode class
Following a Slack discussion [here](https://oceanprotocol.slack.com/archives/C06435G1AN4/p1700064452682359)

TO DO:
 - create OceanNode class
 - make main instantiate a new OceanNode class

mariacarmina:
Fixes #117  .

Changes proposed in this PR:

- created OceanNode class
- renamed OceanNode interface to OceanNodeType
- tests


mariacarmina:
Need to add the test with the new class at least for `status` command.

mariacarmina:
> hi @mariacarmina , from a quick look i think we broke the nodes :-(. It does not run anymore, since there is no entry point.
> 
> AFAIK, the idea was just to make main() return an instance of OceanNode, that we would still would export as a singleton from that module (even if we could create other instances for testing). 
> 
> But we don't have a singleton anymore.. and instead we have multiple instances (several places were we do `new OceanNode(...config)`. But that does not work, as we're basically creating a new node instance on every handler.
> 
> thanks
> 
> 

Hello @paulo-ocean, thanks for feedback. The flow should look like this:
- create OceanNode class that includes config, P2P node, indexer, provider
- delete OceanNode interface from types directory
- make `main` function (outside the OceanNode class) return instance of class OceanNode which is afterwards exported like a singleton.

Are these bullet points correct for the goal of this task?

Thanks

paulo-ocean:
IMO, i have the feeling we might be overdoing it with the OOP approach, trying to encapsulate everything into classes we're also loosing flexibility 
as an example, before i could easily understand that an OceanNode could have a P2P component, an Indexer and a Provider (all optional)
```
export interface OceanNode {
  node: OceanP2P | null
  indexer: OceanIndexer | null
  provider: OceanProvider | null
}
```
Now i see a class that has the 3 (even if they can be null)
```
export class OceanNode {
  private config: OceanNodeConfig
  private node: OceanP2P
  private provider: OceanProvider
  private indexer: OceanIndexer
  public constructor(config: OceanNodeConfig) {
    this.config = config
  }
```
just my 2 cents

paulo-ocean (approved):
LGTM, but as we brainstormed yesterday we might need to re-write our handlers so they become classes with access to the node (passed on the constructor). Having the handlers code inside the node would result on a HUGE file hard to maintain. So keeping them on their own files is probably the best.. 

For that, we need the context and... 
The bind() and/or call() tricks might work for a while... but as we already saw on the nonceHandler... if we have multiple functions that need to access DB there are already  troubles, cause the singleton is gone... and its hard to pass the (this) context around, ...

And, IMO, having this type of call everywhere can become problematic to handle pretty soon, :
```
response = await handleDownloadURLCommand.call(this, task)
      break
    case PROTOCOL_COMMANDS.GET_DDO:
      response = await handleGetDdoCommand.call(this, task)
      break
    case PROTOCOL_COMMANDS.QUERY:
      response = await handleQueryCommand.call(this, task)
      break
    case PROTOCOL_COMMANDS.NONCE:
      response = await getNonce.call(this, task.address)
      break
    case PROTOCOL_COMMANDS.STATUS:
      response = await handleStatusCommand.call(this, task)
      break
    case PROTOCOL_COMMANDS.FIND_DDO:
      response = await findDDO.call(this, task)
      break
```
Besides, using this approach.. we will still not be able to test our handler functions directly, without spreading the bind(), call() first everywhere, to have the right node context inside...

So, IMO, the handlers should be classes, that share and implement a common interface, like `handle(), handleCommand()` or whatever name we want...

and if we for instance, we want the `ddoHandler` to handle a command..
we just do something like: 
` new DDOHandler(node).handle(command)` passing the node to the constructor..
and we solve all our issues of loosing/accessing context inside the handlers

That said, IMO, i think this could be a separate issue... should be a quick refactor anyway

mariacarmina:
For the nonceHandler is problematic for `checkNonce` and that's why I put the db global for this handler.
Option 2: To pass oceanNode as parameter to handlers as well (but again it takes more refactoring)

IMO, the Command should represent an abstract class and have children `QueryCommand`, `FindDDOCommand`, `StatusCommand`, `GetNonceCommand` etc. and each class to override `handle()` method accordingly. Just another idea of how we can refactor the code. If we choose that command to be an abstract class, I'll do it in a follow-up issue this sprint. I don't think it will take that long to do so.

Thoughts? @paulo-ocean @alexcos20

paulo-ocean:
LGTM,
you just forgot something on checkNonceTracking.ts :-)
```
const config = await getConfig()
const oceanNode = new OceanNode(config)
**const dbconn = await new Database(config.dbConfig)
const p2pNode = new OceanP2P(dbconn, config)
oceanNode.setOceanNode(p2pNode, null, null, dbconn)**
```
(the oceanNode does not have P2P, indexer or Provider) ...

but here only P2P was needed (without that it fails)

Apart from that seems to be working fine :+1:

alexcos20 (approved):
lgtm

src/OceanNode.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export class OceanNode {
+  private config: OceanNodeConfig
+  private node: OceanP2P
+  private provider: OceanProvider
+  private indexer: OceanIndexer
+  private db: Database
+  public constructor(config: OceanNodeConfig) {
+    this.config = config
+  }
```
an7e, line 15: `config: OceanNodeConfig` -> can be replaced on `private config: OceanNodeConfig`
no need to reassign
next
we can put all the initialisation of all the modules here and remove `setOceanNode`

src/OceanNode.ts
```diff
@@ -0,0 +1,48 @@
...
+    newIndexer: OceanIndexer,
+    newProvider: OceanProvider,
+    newDbConn: Database
+  ): void {
+    this.node = newNode
+    this.indexer = newIndexer
+    this.provider = newProvider
+    this.db = newDbConn
+  }
+}
```
an7e, line 48: we can move it to constructor, but first we need to make some changes to getConfig
The basic idea is as follows - a config can have or not have a module, which is reflected through such options as for example hasP2P=true
if hasP2P=true we create a module if false we don't create a module

src/components/P2P/handleProtocolCommands.ts
```diff
@@ -66,25 +66,25 @@ export async function handleProtocolCommands(connection: any) {
...
       case PROTOCOL_COMMANDS.NONCE:
-        response = await getNonce(task.address)
+        response = await getNonce(this, task.address)
         break
       case PROTOCOL_COMMANDS.STATUS:
         response = await handleStatusCommand(task)
         break
       case PROTOCOL_COMMANDS.FIND_DDO:
-        response = await findDDO(task)
+        response = await findDDO(this, task)
```
an7e, line 87: all these and other handlers can be converted into classes and replacre all code with matrix
for example, we create a base class 
```
class baseHandler {
   command: COMMAND
   node: OceanP2P
   constructor(node: OceanP2P) {
        this.node= node;
    }
   handleCommand(task: Task): void {
        console.log(this.command);
    }
}

then inherit all other

class ddoHandler extends baseHandler {
   command: DDO_COMMAND
    handleCommand(task: Task): void {
        // do some cool things with task because now you have this.node from class
    }
}

next we can create register for handlers in OceanP2P
somewhere in constructor

handlersClasses = [ddoHandler, WhatEverHandler, ...];
handlers = {}
handlersClasses.forEach(handlersClass=> {
    const handlerInstance = new handlersClass(this)
    handlers[handlerInstance.command] = handlerInstance
})

then in sendToSelf remove handleDirectProtocolCommand
const task = JSON.parse(message)
amd simply call this.handlers[task.command].handleCommand(task)

```

mariacarmina: Yes, but this is already put in another issue that describes your code, which we decided not to include in the POC. Here is the issue https://github.com/oceanprotocol/ocean-node/issues/130.

src/index.ts
```diff
@@ -63,71 +60,58 @@ function loadInitialDDOS(): any[] {
...
+    req.oceanNode = oceanNode
+    next()
+  })
+  app.use(
+    '/docs',
+    swaggerUi.serve,
+    swaggerUi.setup(undefined, {
+      swaggerOptions: {
+        url: '/swagger.json'
+      }
```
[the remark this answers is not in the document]
mariacarmina: It was part from `main` function from `index.ts`, but I can add it to OceanNode class.

src/OceanNode.ts
an7e: @mariacarmina
Yes and it will remain like that, but we can add logic like
if (config.hasP2P) {
  this.node = new OceanP2P(config)
}
to get rid of setOceanNode method
Anyway this is just a suggestion so that you don't have to run through the code later and initialise everything everywhere like here
const config = await getConfig()
const oceanNode = new OceanNode(config)
const dbconn = new Database(config.dbConfig)
const p2pNode = new OceanP2P(dbconn, config)
oceanNode.setOceanNode(p2pNode, null, null, dbconn)

if we are going to use new OceanNode(config) often, and we will do it for all tests usually, i guess, this boilerplate will become frequent

src/index.ts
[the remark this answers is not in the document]
an7e: @mariacarmina 
Before we add it, we need to think about how we could use it, maybe somewhere inside the app there are references to This or similar and hiding the app inside might allow us to use context or super to reach out and call db or whatever where we need it.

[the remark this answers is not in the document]
an7e: @mariacarmina My point is that if you don't see the benefit of it, then there's no point in doing it and you can ignore my suggestion and skip this :)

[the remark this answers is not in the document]
mariacarmina: I think for the moment it should stick with `index.ts` file, not with OceanNode, because as it is right now I need to move the logic from `main()` into OceanNode and I think those routes need to belong for the moment (and also for POC) inside `index.ts`. This is more like a nice-to-have and can be put in a separate issue, too.

src/OceanNode.ts
mariacarmina: The config received as parameter is from `new OceanNode(config)` and I set the `private config: OceanNodeConfig` inside the constructor from the class with the received one from instantiating a new OceanNode object. What did you mean by this?

an7e:
```
public constructor(config: OceanNodeConfig) {
    this.config = config
  }

can be simply replaced like

public constructor(private config: OceanNodeConfig) {}
```

src/OceanNode.ts
an7e: @mariacarmina Yeah, but now I'm thinking it might not be so easy at first glance and we need some advice from @alexcos20 if it worth to do it here

mariacarmina: I have done it in the latest commit, but I've encountered `ERROR: null` again on the unit tests. The good news is that at least the node starts.

an7e (approved):

PR #135, opened 2023-11-29
Handled MetadataState event.

### LINKED ISSUE #46: Indexer:  handle MetadataState event
If metadataState is changing, we need to make sure that we update our ddo.

Depending on state changes, we need to handle this accordingly

- active:   no change if it was active, restore DDO otherwise
- EOL:   mark it as EOL
- Deprecated: mark it
- Revoked:  soft delete it  (keep only id , nft address, and what we need in order to restore it if because active again)
-

See https://docs.oceanprotocol.com/developers/ddo-specification#state

mariacarmina:
Fixes #46  .

Changes proposed in this PR:

- 
- 
-


src/components/Indexer/index.ts
```diff
@@ -96,6 +101,51 @@ export class OceanIndexer {
...
+      if ('nft' in ddo && ddo.nft.state !== event.data.metadataState) {
+        ddo.nft.state = event.data.metadataState
+      } else {
+        // Still update until we validate and polish schemas for DDO.
+        // But it should update ONLY if first condition is met.
+        // Check https://github.com/oceanprotocol/aquarius/blob/84a560ea972485e46dd3c2cfc3cdb298b65d18fa/aquarius/events/processors.py#L663
+        ddo.nft = {
+          state: event.data.metadataState
+        }
+      }
```
bogdanfazakas, line 134: I would not add all this logic to the indexer, but keep in the event_processor and in the index just to call the database method. I would advocate for this for 2 reasons, one keep the indexer main thread as simple as possible and two consider that when running on multiple networks all this logic would be processed in the respective network thread.

bogdanfazakas (changes requested):

bogdanfazakas:
Also i think it's good that you added lots of logs but maybe we should remove the duplicated ones, so maybe worth taking a look and clear some of the logs, this optional tho

src/components/Indexer/index.ts
mariacarmina: I need the DDO database in processMetadataStateEvent (if we want to move it inside `eventProcessor`) to extract the DDO data, because those checks are based on the specific DDO. As far as I've seen there the db ref is only in indexer.

bogdanfazakas: yep sure we can recreate the instance in the working thread easily, but we initially decided to treat the db as a singleton that's we had this worker - main_thread exchange of events to use the db since we cannot pass object with prototypes to the worker

mariacarmina: Ok, I'll recreate the db instance in eventProcessor

bogdanfazakas: tbh not sure what's the best way of doing it, maybe keeping the necessary logic in the indexer main thread I'd say

bogdanfazakas: with the prev comment i just tried to explain to Paulo the reason behind this message exchange and why we don't have db connection in the working threads

paulo-ocean: i miss the OceanNode singleton :-( we need to have it back

mariacarmina:
> Also i think it's good that you added lots of logs but maybe we should remove the duplicated ones, so maybe worth taking a look and clear some of the logs, this optional tho

There are no duplicates as far as I saw, but I filtered some of them. If you still want to delete some, please specify here which. Thanks!

src/components/Indexer/index.ts
```diff
@@ -44,6 +46,9 @@ export class OceanIndexer {
         if (event.method === EVENTS.METADATA_CREATED) {
           this.saveDDO(event.network, event.data)
         }
+        if (event.method === EVENTS.METADATA_STATE) {
+          this.storeState(event.network, event.data)
```
mariacarmina, line 50: To avoid duplicate code, we can use `saveDDO`, because now, it is the same logic and we can adapt the logs. If we decide that index.ts should only update in db, I propose to use `saveDDO`. What do you think @bogdanfazakas @paulo-ocean ?

paulo-ocean: if its the same logic, than we do not need 2 different functions for sure.. maybe refactor both into just 1 with a more generic name (that would be suitable for both)
and pass the event.method as a 3rd parameter , so we could distinguish the original event and therefore the logs

bogdanfazakas: i did something similar for metadata-update you can take a look @mariacarmina but the direction is right

paulo-ocean: @mariacarmina 
better double check this line:
`if (event.method === EVENTS.METADATA_CREATED || EVENTS.METADATA_STATE) {`
i'm not sure that's exactly what you wanted

mariacarmina: Yes, I've missed the event.method :). On my way to fix it. Thanks!

bogdanfazakas (commented):
Still not sure if we need that new instance of database in the event processor, in the PR we are using that to get the get the DDO and the other instance from indexer main thread to save or update the DDO.
So i think we need to decide how we are going fwd with the database instance singleton or not and keep just one approach, in this case if singleton use the db only from indexer main thread otherwise do anything related to the db in the working threads.

mariacarmina:
> Still not sure if we need that new instance of database in the event processor, in the PR we are using that to get the get the DDO and the other instance from indexer main thread to save or update the DDO. So i think we need to decide how we are going fwd with the database instance singleton or not and keep just one approach, in this case if singleton use the db only from indexer main thread otherwise do anything related to the db in the working threads.

As @paulo-ocean mentioned, this shouldn't be a show stopper. I decided to instantiate a new Database object to the event processor (for the moment I put only in metadata state, but we can declare it globally in that file as config was). If not, I've seen this message of yours @bogdanfazakas:

> maybe keeping the necessary logic in the indexer main thread I'd say

Can you give more details about this approach, please?
In other words, for the moment I want to see if this change with database instance is more like an enhancement issue or it will actually break the functionality for POC. Can I have an answer to this statement, please? Because if it's not impacting the functionality for POC (and we decide it that way), maybe we can put these details to a separate issue to tackle it afterwards. What do you think @paulo-ocean @alexcos20 @bogdanfazakas @jamiehewitt15 ?

tests/integration/indexer.test.ts
```diff
@@ -137,4 +137,19 @@ describe('Indexer stores a new published DDO', () => {
...
+
+  delay(100000)
+
+  it('should get the updated state', async () => {
+    const result = await nftContract.getMetaData()
+    const resolvedDDO = await waitToIndex(assetDID, database)
+    // Expect the result from contract
+    expect(resolvedDDO.nft.state).to.equal(parseInt(result[2].toString()))
+  })
 })
```
jamiehewitt15, line 155: Can you add a couple of extra tests here to check that it handles invalid input gracefully: 
1. an invalid setMetaDataState update, test that you get the right rejection response. 
2. call getMetaData on an asset that isn't indexed, testing that you get the right not found response.
3. Attempt to update the metadata with another account and check that it was not allowed.

jamiehewitt15:
I agree with paulo's comment on the DB, creating a new instance shouldn't necessarily be a showstopper

> not going into details here, but if the issue is having 'DB access', i think that should not be a show stopper. We can always get the DB config and create a new Database instance if needed, anywhere. Its always the same DB everywhere, anyway

tests/integration/indexer.test.ts
mariacarmina: I think you are proposing to test more the contracts functionality, not necessary the indexer part. Those tests might be included in the contracts repo. I can add more `expect` statements regarding DDO structure instead.

mariacarmina: For usecase 2, it does not throw the error per-se (see https://github.com/oceanprotocol/ocean-node/actions/runs/7127773665/job/19408379978?pr=135#step:12:1153), because the error of not found was caught before in the image below. The error log with Not found 404 can be seen in the logs provided in the link that I provided for the CI run. I suggest to skip it for the moment, since the main logic that we were interested was that contract call corresponds with DDO structure stored by Indexer.
Screenshot 2023-12-07 at 14 00 46

mariacarmina: This is what I've seen in the console:
`Request /collections/ddo_v0.1/documents/did:op:86d375f360f203cd734eb9b0f07728302b827e7716a6e041c94bae45c73781e2: Request to Node localhost was made. Response Code was 404.`

For sure if we want to improve the implementation we should tackle in a separate issue since my PR does not cover necessary this scope (it is not related to MetadataState event). What do you think?

PR #137, opened 2023-12-01
PR for APIs backward compatible

### LINKED ISSUE #133: Cleanup and make ocean-node APIs backward compatible

an7e:
Fixes #133 

Changes proposed in this PR:

- added service router
- added api methods

please see the task description to see the full api list here #133


src/components/httpRoutes/index.ts
```diff
@@ -15,3 +18,5 @@ httpRoutes.use(getProvidersForDidRoute)
 httpRoutes.use(broadcastCommandRoute)
 httpRoutes.use(directCommandRoute)
 httpRoutes.use(logRoutes)
+httpRoutes.use('/api/service/', providerRoutes)
```
paulo-ocean, line 21: isn't the path `/api/service**s**/` ? :-)

src/components/httpRoutes/provider.ts
```diff
@@ -0,0 +1,91 @@
...
+import { getNonce } from '../core/nonceHandler.js'
+import { streamToString } from '../../utils/util.js'
+import { Readable } from 'stream'
+import { encrypt } from '../../utils/crypt.js'
+import { calculateFee } from '../core/feesHandler.js'
+import { DDO } from '../../@types/DDO/DDO'
+
+export const providerRoutes = express.Router()
+
+providerRoutes.post('/encrypt', async (req, res) => {
```
paulo-ocean, line 11: Shouldn't we use the handler functions we already have?  for instance, turning the incoming request into the equivalent "directCommand" task ?
Most of these, we implemented as directCommands, so maybe it makes some sense to try to reuse some of that handler logic?

We are also not logging anything, neither the errors (just sending 500 error back) , and not validating the inputs (which we already have on the directCommands)
I think its important to look at this in more detail, since we are going to rely on the CLI for testing, and pretty much we will just use these legacy endpoints

This does not apply to all endpoints, but for encrypting, getting fees, nonce, findDDO, etc, we already have some handler functions that might be handy as they have more checks already implemented, 
my 2 cents
tkx

src/components/httpRoutes/aquarius.ts
```diff
@@ -0,0 +1,109 @@
...
+import { Readable } from 'stream'
+import { handleGetDdoCommand } from '../core/ddoHandler.js'
+import { PROTOCOL_COMMANDS } from '../../utils/constants.js'
+import { handleQueryCommand } from '../core/queryHandler.js'
+
+export const aquariusRoutes = express.Router()
+
+aquariusRoutes.get('/assets/ddo/:did', async (req, res) => {
+  try {
+    const { did } = req.params
```
alexcos20, line 12: check if did is passed, otherwise send proper http error code

src/components/httpRoutes/aquarius.ts
```diff
@@ -0,0 +1,109 @@
...
+      res.status(result.status.httpStatus).send(result.status.error)
+    }
+  } catch (error) {
+    res.status(500).send('Internal Server Error')
+  }
+})
+
+aquariusRoutes.get('/assets/metadata/:did', async (req, res) => {
+  try {
+    const { did } = req.params
```
alexcos20, line 31: check if did is passed

src/components/httpRoutes/provider.ts
alexcos20: I agree as well.  http routes shoud have no command logic, only some params checking for proper error msg

src/index.ts
```diff
@@ -96,6 +96,7 @@ logger.addTransport(customLogTransport)
 // global
 oceanNode.setOceanNode(node, indexer, provider, dbconn)
 if (config.hasHttp) {
+  app.use(express.raw())
```
alexcos20, line 99: out of curiosity, why do we need this?

src/components/httpRoutes/index.ts
an7e: lol missed that

src/components/httpRoutes/aquarius.ts
an7e: but i passed
```
if (result.stream) {
      const ddo = JSON.parse(await streamToString(result.stream as Readable))
      res.json(ddo)
    } else {
      res.status(result.status.httpStatus).send(result.status.error) <= HERE
    }
```
or did I misunderstand?

src/index.ts
an7e: to be able to get data in req.body
```
providerRoutes.post('/encrypt', async (req, res) => {
  try {
    const data = Uint8Array.from(req.body)
    const encryptedData = await encrypt(data, 'ECIES')
    if (encryptedData) {
      res.send(encryptedData)
    } else {
      res.status(400).send()
    }
  } catch (error) {
    res.status(500).send('Internal Server Error')
  }
})
```
without it 
`const data = Uint8Array.from(req.body)`
will be empty

src/components/httpRoutes/provider.ts
paulo-ocean: Not sure if i'm following you here:
Your encrypHandler does:
```
// do encrypt magic
 const encryptedData = await encrypt(blobData, task.encryptionType)
```
```
which is the same thing you're doing on the provider route:
const encryptedData = await encrypt(data, 'ECIES')
```
What's the difference here? It looks the same to me.
you could still use handleEncryptCommand(
{... }
)

src/components/httpRoutes/aquarius.ts
```diff
@@ -0,0 +1,117 @@
...
+      res.json(queryResult)
+    } else {
+      res.status(result.status.httpStatus).send(result.status.error)
+    }
+  } catch (error) {
+    res.status(500).send('Internal Server Error')
+  }
+})
+
+aquariusRoutes.get('/aquarius/state/ddo', async (req, res) => {
```
paulo-ocean, line 75: This route is not correct:
should be only `/state/ddo` not` /aquarius/state/ddo` because you already have: `/api/aquarius`  on all these routes before... like this it becomes `/api/aquarius/aquarius/state/ddo`

paulo-ocean (changes requested):

src/components/httpRoutes/aquarius.ts
```diff
@@ -0,0 +1,117 @@
...
+      command: PROTOCOL_COMMANDS.GET_DDO
+    })
+    if (result.stream) {
+      const ddo = JSON.parse(await streamToString(result.stream as Readable))
+      res.json(ddo)
+    } else {
+      res.status(result.status.httpStatus).send(result.status.error)
+    }
+  } catch (error) {
+    res.status(500).send('Internal Server Error')
```
paulo-ocean, line 29: you should have logs on the errors at least, these are calls are throwing error, and having only the 500 error does not help (when testing, i had to put a console.log to be able to understand the issue)
The problem is on `handleGetDdoCommand` and `handleQueryCommand`
when you're constructing the response stream with 
`Readable.from(string)` 
better use instead:
`new ReadableString(string)`
The function `streamToString(stream)` is complaining about the args

paulo-ocean (changes requested):

src/components/httpRoutes/aquarius.ts
```diff
@@ -0,0 +1,117 @@
...
+        query_by: 'nft.address'
+      }
+    }
+    const node = req.oceanNode.getP2PNode()
+    const result = await handleQueryCommand(node, {
+      query,
+      command: PROTOCOL_COMMANDS.QUERY
+    })
+    if (result.stream) {
+      const queryResult = JSON.parse(await streamToString(result.stream as Readable))
```
paulo-ocean, line 105: there's another issue here:
the returned `queryResult` object is an array, like `[{"facet_counts":[],"found":1,"hits":[{"document":{"`
so, the  check `queryResult.found` fails because it should be `queryResult[0].found` (same for the other references bellow)
and inside the `hits[0]` there is the `nft.state` is inside `document`
instead of `queryResult.hits[0].nft.state` should be something like:
`queryResult[0].hits[0].document.nft.state`
thanks

paulo-ocean (changes requested):

paulo-ocean:
btw, i used postman for testing all the routes.. once we fix these issues i think we're good...
It would be nice to have param checking and more logging, but i guess we can do that after POC

paulo-ocean (approved):
lgtm :+1:

alexcos20 (approved):
lgtm. if bugs are found, will fix them later
@an7e - can you open a PR later just for download ?

PR #140, opened 2023-12-05
WIP - Feature/ Index metadata update

### LINKED ISSUE #45: Indexer:  index MetadataUpdated event
Every time a DDO is updated, we need to track this

Event source:  https://github.com/oceanprotocol/contracts/blob/main/contracts/templates/ERC721Template.sol#L57C11-L66

The flow follows same pattern as #44, creating the asset if does not exists, otherwise update it

bogdanfazakas:
Fixes #45  .

Changes proposed in this PR:

- detect metdata update & index ddo

### DIFF
--- a/src/components/Indexer/eventProcessor.ts
+++ b/src/components/Indexer/eventProcessor.ts
@@ -14,7 +14,7 @@ export const INDEXER_LOGGER: CustomNodeLogger = getCustomLoggerForModule(
   defaultConsoleTransport
 )
 
-export const processMetadataCreatedEvent = async (
+export const processMetadataEvents = async (
   event: ethers.Log,
   chainId: number,
   provider: JsonRpcApiProvider
--- a/src/components/Indexer/index.ts
+++ b/src/components/Indexer/index.ts
@@ -41,8 +41,11 @@ export class OceanIndexer {
         if (event.method === 'store-last-indexed-block') {
           this.updateLastIndexedBlockNumber(event.network, event.data)
         }
-        if (event.method === EVENTS.METADATA_CREATED) {
-          this.saveDDO(event.network, event.data)
+        if (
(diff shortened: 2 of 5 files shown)


mariacarmina (commented):
Some suggestions/open points regarding processing MetadataUpdated from my side.

src/components/Indexer/eventProcessor.ts
```diff
@@ -26,34 +26,45 @@ export const INDEXER_LOGGER: CustomNodeLogger = getCustomLoggerForModule(
 
 const config = await getConfig()
 
-export const processMetadataCreatedEvent = async (
+export const processMetadataEvents = async (
```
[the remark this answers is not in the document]
bogdanfazakas: yep saw the checks in the aqua repo, they were not mentioned in the https://github.com/oceanprotocol/ocean-node/issues/45 and did't added them, but if we consider them mandatory for the POC we can add some of them for sure.

src/components/Indexer/crawlerThread.ts
```diff
@@ -31,15 +31,14 @@ function delay(ms: number) {
 }
 
 export async function proccesNetworkData(): Promise<void> {
+  let lastSavedBlock = lastIndexedBlock
   while (true) {
```
paulo-ocean, line 35: maybe instead of an infinite while() loop with a delay() we could do a setInterval() instead?
probably less intensive cause we're not locking inside the loop. Maybe something we could consider later
btw, we already had a sleep() function on utils :-)

mariacarmina (approved):
LGTM

src/components/Indexer/crawlerThread.ts
bogdanfazakas: thanks for raising this, should be good now

PR #141, opened 2023-12-05
Check and validate order transaction

### LINKED ISSUE #88: Check on-chain order
When an user want to download an asset, he/she must sent the txid proving that payment using datatoken was done.

That transaction can have two events:
 - OrderStarted   - when a payment was made, using new provider fees
 - OrderReused - whan an order was still valid, but provider fees were expired

We must validate tx send by user by:
 - checking if user is who it claims to be
 - check if datatoken belongs to nft
 - check if datatoken belongs to service
 - check if order is still valid
  - if user tx has OrderReused, fetch tx that contains OrderStarted
  

Reference:  https://github.com/oceanprotocol/contracts/blob/main/contracts/templates/ERC20Template.sol#L82-L97

jamiehewitt15:
Fixes #88

Changes proposed in this PR:

- Checking the user address
- Checking if the datatoken was deployed by the nft
- Checking if the datatoken is part of the right service
- checking if the order is still valid
- Testing a successful purchase with an orderStarted event
- Testing a successful purchase with an orderReuse event
- Testing that the transaction is rejected if the serviceIndex is invalid.
- Testing that the transaction is rejected if the user address is invalid.


package.json
```diff
@@ -85,6 +85,7 @@
     "eslint-plugin-promise": "^6.1.1",
     "mocha": "^10.2.0",
     "prettier": "^3.0.3",
-    "tsx": "^3.12.8"
+    "tsx": "^3.12.8",
+    "web3": "^4.2.2"
```
bogdanfazakas, line 89: do we really need web3 as a dependancy since we already have ethers ?

tests/testUtils.ts
```diff
@@ -0,0 +1,74 @@
+import { Database } from '../src/components/database/index.js'
+import { Signer } from 'ethers'
+import Web3 from 'web3'
+
+export const genericAsset = {
```
bogdanfazakas, line 5: we now have this in the data -> ddo.ts

tests/testUtils.ts
```diff
@@ -0,0 +1,74 @@
...
+      description: 'Download service',
+      files: '',
+      datatokenAddress: '0x0',
+      serviceEndpoint: 'http://172.15.0.4:8030',
+      timeout: 0
+    }
+  ]
+}
+
+export function sleep(ms: number) {
```
bogdanfazakas, line 36: there is also a util in the codebase that we can reuse see` import { sleep } from '../../src/utils/util'`

package.json
paulo-ocean, line 89: hi @jamiehewitt15 do we really web3 lib as well?
i mean, we already have ethers, not sure if we need both. Also we are only using it in 1 function on testUtils.ts while we already are using ethers in multiple places
thanks

paulo-ocean: yeah, just did the same comment :-)

jamiehewitt15: Yeah my thinking was the same initially but it was causing some problems. I tried it with ethers, with the function below but Ganache throws an error. Do you have any ideas on this?

```
signMessage error Error: unsupported operation (operation="personal_sign", info={ "error": { "code": -32700, "message": "The method personal_sign does not exist/is not available", "stack": "Error: The method personal_sign does not exist/is not available\n    at Executor.execute (/app/dist/node/1.js:2:284518)\n    at EthereumProvider._requestRaw (/app/dist/node/1.js:2:183642)\n    at Connector.<anonymous> (/app/dist/node/1.js:2:104154)\n    at Connector.handle (/app/dist/node/1.js:2:104577)\n    at /app/dist/node/1.js:2:8596" }, "payload": { "id": 33, "jsonrpc": "2.0", "method": "personal_sign", "params": [ "0x307862396665653837633366323630626631363339373838363336643337643237656631666431653231393864616533363030396265393934643635343432633139", "0xe2dd09d719da89e5a3d0f2549c7e24566e947260" ] } }, code=UNSUPPORTED_OPERATION, version=6.8.1)
    at makeError (file:///home/jamie/Desktop/ocean/ocean-node/node_modules/ethers/src.ts/utils/errors.ts:694:21)
    at JsonRpcProvider.getRpcError (file:///home/jamie/Desktop/ocean/ocean-node/node_modules/ethers/src.ts/providers/provider-jsonrpc.ts:971:20)
    at file:///home/jamie/Desktop/ocean/ocean-node/node_modules/ethers/src.ts/providers/provider-jsonrpc.ts:526:45
    at processTicksAndRejections (node:internal/process/task_queues:95:5) {
  code: 'UNSUPPORTED_OPERATION',
  operation: 'personal_sign',
  info: {
    error: {
      message: 'The method personal_sign does not exist/is not available',
      stack: 'Error: The method personal_sign does not exist/is not available\n' +
        '    at Executor.execute (/app/dist/node/1.js:2:284518)\n' +
        '    at EthereumProvider._requestRaw (/app/dist/node/1.js:2:183642)\n' +
        '    at Connector.<anonymous> (/app/dist/node/1.js:2:104154)\n' +
        '    at Connector.handle (/app/dist/node/1.js:2:104577)\n' +
        '    at /app/dist/node/1.js:2:8596',
      code: -32700
    },
    payload: {
      method: 'personal_sign',
      params: [Array],
      id: 33,
      jsonrpc: '2.0'
    }
  },
  shortMessage: 'unsupported operation'
}

```

```
import { Signer, Signature } from 'ethers'

export async function signMessage(
  message: string,
  signer: Signer
): Promise<{ v: string; r: string; s: string }> {
  try {
    const signedMessage = await signer.signMessage(message)
    const signature = Signature.from(signedMessage)

    return {
      v: signature.v.toString(),
      r: signature.r,
      s: signature.s
    }
  } catch (e) {
    console.error('signMessage error', e)
    throw new Error('Signing message failed')
  }
}

```

jamiehewitt15: `getSigner(0)` just get's the first account made available from ganache, it's not the zero address. It still gives the same `UNSUPPORTED_OPERATION` error whichever account is used.

paulo-ocean: so probably that's it, you cannot sign with it as you don't have the private key

jamiehewitt15: web3 is removed now

jamiehewitt15: web3 is removed now

src/utils/util.ts
```diff
@@ -10,3 +11,35 @@ export function getEventFromTx(txReceipt: { logs: any[] }, eventName: any) {
...
+        }
+      })
+      .map((log: any) => ({
+        ...contractInterface.parseLog(log),
+        log
+      }))
+
+    return events.length > 0 ? events : null
+  } catch (error) {
+    console.error('Error fetching event from transaction:', error)
```
bogdanfazakas, line 42: Lets use logger class here as well

bogdanfazakas (approved):
the logic looks good to me 👍

PR #145, opened 2023-12-06
Index order events

### LINKED ISSUE #125: Indexer: index orders event
Right now, we are using subgraph to index two events:
  - Order
  - OrderReuse
 
Let's have them in our indexer.

Source events:  https://github.com/oceanprotocol/contracts/blob/main/contracts/templates/ERC20Template.sol#L82-L97

Reference:  https://github.com/oceanprotocol/ocean-subgraph/blob/main/schema.graphql#L133-L184

mariacarmina:
Fixes #125  .

Changes proposed in this PR:

- index OrderStarted
- index OrderReused
- tests
- create orderSchema
- store orders in the db


mariacarmina:
needs integration tests

mariacarmina:
CURRENT PROBLEM:

`getERC721Address` smart contract function retruns a different NFT address -> returns a different DID checksum which cannot be identified in the DB.
For metadata events, `event.address` (which is == NFT address) is a different address (rather than `getERC721Address`) and stores the DDO with a certain checksum which does not correspond with the one generated in OrderStarted processor.

mariacarmina:
Here is the source code from the contracts which sets the `erc721Address` https://github.com/oceanprotocol/contracts/blob/main/contracts/templates/ERC20Template.sol#L268.

bogdanfazakas:
Maria is it enough to just update the stats.orders ? this is basically just replacing what we have in current aquarius, but from my understanding we should also have a new schema and store order informations that are stored in the subgraph at for the moment ?

src/components/Indexer/eventProcessor.ts
```diff
@@ -102,3 +115,102 @@ export const processMetadataStateEvent = async (
...
+      }
+    }
+    INDEXER_LOGGER.logMessage(`Found did ${did} for order starting on network ${chainId}`)
+    return ddo
+  } catch (err) {
+    INDEXER_LOGGER.log(LOG_LEVELS_STR.LEVEl_ERROR, `Error retrieving DDO: ${err}`, true)
+  }
+}
+
+export const processOrderReusedEvent = async (
```
bogdanfazakas, line 174: processOrderReusedEvent is so pretty much using same code as processOrderStartedEvent so think we can refactor a bit and get rid of as many code duplications as possible

mariacarmina:
> Maria is it enough to just update the stats.orders ? this is basically just replacing what we have in current aquarius, but from my understanding we should also have a new schema and store order informations that are stored in the subgraph at for the moment ?

From the last week discussion, I understood from @alexcos20 that the pricing (which was previously retrieved from subgraph into aquarius) will be tackled in the frontend component (which will interact directly with the smart contracts). @alexcos20 suggested that I should keep in DDO only orders counter.

bogdanfazakas:
> > Maria is it enough to just update the stats.orders ? this is basically just replacing what we have in current aquarius, but from my understanding we should also have a new schema and store order informations that are stored in the subgraph at for the moment ?
> 
> From the last week discussion, I understood from @alexcos20 that the pricing (which was previously retrieved from subgraph into aquarius) will be tackled in the frontend component (which will interact directly with the smart contracts). @alexcos20 suggested that I should keep in DDO only orders counter.

I remember this talk and wasn't it about the price from the stats? And we decided to not add it there, but that should have nothing to do with storing the actual order info and reused order info like we have in the subgraph. But let's see other opinions as well and we'll move fwd.

alexcos20 (changes requested):
We need to store orders in the database, and I'm not seeing that in the code.
Primary use case is to check if an existing order is still valid.
Orders != Price schema  (which indeed, price schemas can be fetched realtime from contracts)

tests/integration/indexer.test.ts
```diff
@@ -96,6 +122,11 @@ describe('Indexer stores a new published DDO', () => {
     const event = getEventFromTx(txReceipt, 'NFTCreated')
     nftAddress = event.args[0]
     assert(nftAddress, 'find nft created failed')
+    console.log('nftAddress for OrderStarted test: ', nftAddress)
```
bogdanfazakas, line 125: remove this

tests/data/orderSchema.ts
```diff
@@ -0,0 +1,18 @@
+import { TypesenseCollectionCreateSchema } from '../../src/@types'
+
+export const orderSchema: TypesenseCollectionCreateSchema = {
+  name: 'order',
+  enable_nested_fields: true,
+  fields: [
+    { name: 'type', type: 'string', enum: ['startOrder', 'reuseOrder'] },
```
bogdanfazakas, line 7: In the current subgraph we have reuses of an orders as a property within that order, is it ok to have them indexed have as separate entities with the current implementation ? will this break things thinking of backwards compatibility ?

src/components/Indexer/eventProcessor.ts
```diff
@@ -102,3 +115,138 @@ export const processMetadataStateEvent = async (
...
+
+    try {
+      const startOrder = await dbconn.order.retrieve(startOrderId)
+      if (!startOrder) {
+        INDEXER_LOGGER.logMessage(
+          `Detected OrderReused changed for order ${startOrderId}, but it does not exists.`
+        )
+        return
+      }
+      await dbconn.order.create(
```
bogdanfazakas, line 232: not really sure of how we are dealing with this kind of events that require the update 2 models in the database, in this case we are using the newly created instance from the save the order , and passing back the ddo to the main thread to store it using the original db connection.

src/components/Indexer/eventProcessor.ts
```diff
@@ -102,3 +115,138 @@ export const processMetadataStateEvent = async (
...
+      .digest('hex')
+  try {
+    const ddo = await dbconn.ddo.retrieve(did)
+    if (!ddo) {
+      INDEXER_LOGGER.logMessage(
+        `Detected OrderReused changed for ${did}, but it does not exists.`
+      )
+      return
+    }
+    INDEXER_LOGGER.logMessage(`Found did ${did} on network ${chainId}.`)
```
bogdanfazakas, line 221: lets delete this

src/components/Indexer/eventProcessor.ts
```diff
@@ -102,3 +115,138 @@ export const processMetadataStateEvent = async (
...
+  provider: JsonRpcApiProvider
+) => {
+  const receipt = await provider.getTransactionReceipt(event.transactionHash)
+  const iface = new Interface(ERC20Template.abi)
+  const eventObj = {
+    topics: receipt.logs[0].topics as string[],
+    data: receipt.logs[0].data
+  }
+  const decodedEventData = iface.parseLog(eventObj)
+  INDEXER_LOGGER.logMessage(`Decoded event data args: ${decodedEventData.args}`, true)
```
bogdanfazakas, line 131: we can delete this i guess since we have a better log bellow

tests/data/orderSchema.ts
mariacarmina: That was @alexcos20  proposal. I followed it accordingly.

mariacarmina:
> without the more detailed things like the order schema or how we work with the db

From what I've discussed with @alexcos20, the preferred schema was something that follows this model:
```json 
{
 id:   txID,
 type: 'startOrder' | 'reuseOrder'
 timestamp: block.timestamp,
 consumer,
 payer,
 startOrderId: (doar daca e de type reuse)
}
```
@alexcos20 can you confirm as well and review it?

alexcos20 (approved):

bogdanfazakas (approved):

PR #151, opened 2023-12-07
add default fee tokens if not configured

### LINKED ISSUE #142: Fee Token defaults
Right now, owner must define FEE_TOKEN in order to start the node.

Let's have some defaults:
 - if FEE_AMOUNT is not defined, make it 0
 - if FEE_TOKEN is not defined, make it equal to OCEAN address per each chain
 
 This will ease onboarding for non tech users

paulo-ocean:
Fixes #142 

Changes proposed in this PR:

- always have default values for FEE_TOKENS and FEE_AMOUNT, even if not configured
- refactor some related code
- common util function for getting artifacts addresses
- some related cleaning on the env. vars + logging

### DIFF
--- a/src/components/core/feesHandler.ts
+++ b/src/components/core/feesHandler.ts
@@ -370,7 +370,7 @@ export async function getProviderFeeToken(chainId: number): Promise<string> {
   const result = (await getConfiguration()).feeStrategy.feeTokens.filter(
     (token: FeeTokens) => Number(token.chain) === chainId
   )
-  return result.length ? result[0].token : null
+  return result.length ? result[0].token : ethers.ZeroAddress
 }
 
 /**
--- a/src/components/core/feesHandler.ts
+++ b/src/components/core/feesHandler.ts
@@ -4,7 +4,7 @@ import { Command, ICommandHandler, GetFeesCommand } from '../../utils/constants.
 import { DDO } from '../../@types/DDO/DDO'
 import { Service } from '../../@types/DDO/Service'
 import { AssetUtils } from '../../utils/asset.js'
-import { P2PCommandResponse } from '../../@types'
+import { OceanNodeConfig, P2PCommandResponse } from '../../@types'
 import { OceanP2P } from '../P2P/index'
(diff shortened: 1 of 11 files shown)


src/utils/address.ts
```diff
@@ -0,0 +1,97 @@
...
+        `${homedir}/.ocean/ocean-contracts/artifacts/address.json`,
+      'utf8'
+    )
+    return JSON.parse(data)
+  } catch (error) {
+    return null
+  }
+}
+// default Ocean token (only) addresses per chain
+export const OCEAN_TOKEN_ADDRESS_PER_CHAIN = {
```
alexcos20, line 22: no need for hardcoding, you can find them in address.json, see https://github.com/oceanprotocol/contracts/blob/main/addresses/address.json

alexcos20 (changes requested):

src/utils/address.ts
paulo-ocean: yeah, i missed that, we are already importing it in other place
thanks

src/components/core/feesHandler.ts
```diff
@@ -343,28 +347,28 @@ export async function getFees(task: GetFeesCommand): Promise<P2PCommandResponse>
...
  * @returns the token address
  */
-export function getProviderFeeToken(chainId: string): string | null {
-  const result = config.feeStrategy.feeTokens.filter(
-    (token: FeeTokens) => token.chain === chainId
+export async function getProviderFeeToken(chainId: number): Promise<string> {
+  const result = (await getConfiguration()).feeStrategy.feeTokens.filter(
+    (token: FeeTokens) => Number(token.chain) === chainId
   )
   return result.length ? result[0].token : null
```
alexcos20, line 373: you should return zero address (0x0000000000000000000000000000000000000000) instead of null

paulo-ocean: alright, thanks. just updated it

bogdanfazakas (approved):

alexcos20 (approved):
lgtm

PR #154, opened 2023-12-11
Remove enforcement for exporting ARWEAVE and IPFS gateway.

mariacarmina:
Fixes #.

Changes proposed in this PR:

- adds checks for ipfs and arwave and throw error if not supported and method is called for one of them
- adds the supported storage types in the status handler


alexcos20 (changes requested):
Not sure if this is enough.

How Ocean node will behave when somebody will request downloadFile, with an encrypted url using IPFS or ARWEAVE, and those envs are missing?
Also, it would be nice to expose this to status endpoint (ie:  supported storage: { ipfs:true,arweave:false} )

jamiehewitt15:
We could set defaults for each of them as fall back if they are not defined:
```
IPFS_GATEWAY="https://ipfs.oceanprotocol.com/"
ARWEAVE_GATEWAY="https://arweave.net/"
```

bogdanfazakas (changes requested):
maybe we should also addd more checks to the commands working with file and throw proper errors if one of ipfs and arware file is received and the node is not able to handle that kind of file manipulation

jamiehewitt15:
I'm just wondering why we don't use public gateways as a default? Instead of returning an error when they haven't been defined

alexcos20 (approved):
lgtm

paulo-ocean (approved):
lgtm

bogdanfazakas (approved):

PR #157, opened 2023-12-11
Handle downloads

### LINKED ISSUE #128: Replace downloadURL command with the download command
Command description:

{
  command: "download",
  node: "NodeB ID",  // optional, if missing, it refers to current node
  documentId,
  serviceId,
  transferTxId,
  fileIndex,
  nonce,
  consumerAddress,
  signature.
 }
Inputs:

 - documentId: String object containing document id (e.g. a DID)
 - serviceId: String, representing the list of `file` objects that describe each file in the dataset
 - transferTxId: Hex string -- the id of on-chain transaction for payment using datatokens 
 - fileIndex: integer, the index of the file from the files list in the dataset
 - nonce: String object, the nonce (either integer, either UTC timestamp format) required for download request
 - consumerAddress: String object containing consumer's address
 - signature: String object containg user signature (signed message).
     The signature is based on hashing the following string concatenation consisting of:
       documentId + nonce

This a glue of all issues in this epic, because the steps are the following:
 - check nonce & signature
 - check provider fees
 - check order tx
 - decrypt url
 - read file content & stream it  using Storage classes
 - check if we need to encrypt it using user's private key
 

By doing this, we are going to replace downloadURL command (which was designed only as a test helper)

jamiehewitt15:
Fixes #128

Changes proposed in this PR:

- Find the DDO and ensure it is in the right format. 
- check nonce & signature
- check provider fees
- check and validate the order transaction
- decrypt files URL
- Stream content to user


src/components/core/downloadHandler.ts
```diff
@@ -22,10 +30,99 @@ async function getFileFromURL(fileURL: string): Promise<any> {
...
+      `Invalid payment transaction: ${paymentValidation.message}`,
+      true
+    )
+    throw new Error(paymentValidation.message)
+  }
+
+  // 6. Proceed to download the file
+  return await handleDownloadURLCommand(node, {
+    command: PROTOCOL_COMMANDS.DOWNLOAD_URL,
+    url: ddo.services[task.serviceIndex].files,
```
alexcos20, line 117: you need to decrypt this first..

alexcos20:
We need to use storage classes, and not just try to download a url

At https://github.com/oceanprotocol/ocean-node/blob/issue-128-download/src/components/core/downloadHandler.ts#L134
```
const inputStream = task.url.startsWith('http')
      ? await getFileFromURL(task.url) // remote url
      : fs.createReadStream(task.url) //  local file
 ```

**Review by alexcos20 — CHANGES_REQUESTED** [review]
see my comments

File: src/components/core/downloadHandler.ts [hunk 2]
```diff
@@ -111,10 +112,24 @@ export async function handleDownload(
...
+  const encryptedUrlHex = ddo.services[task.serviceIndex].files
+  // Check if the string starts with '0x' and remove it if present
+  const hexString = encryptedUrlHex.startsWith('0x')
+    ? encryptedUrlHex.substring(2)
+    : encryptedUrlHex
+
+  // Convert the hex string to a Uint8Array
+  const encryptedUrlBytes = Uint8Array.from(Buffer.from(hexString, 'hex'))
+  // Call the decrypt function with the appropriate algorithm
+  const decryptedUrlBytes = await decrypt(encryptedUrlBytes, 'AES')
```
alexcos20: it's ECIES, not AES :)

jamiehewitt15: ok, thanks

Issue #159, opened 2023-12-12
Tests fail when instantiating OceanP2P


jamiehewitt15:
Tests fail when instantiating OceanP2P

jamiehewitt15:
This is fixed now

PR #160, opened 2023-12-12
Creating a simple test for OceanP2P

jamiehewitt15:
Related to #159 

Changes proposed in this PR:

- Creating a simple test for OceanP2P
- This PR shows the issues with trying to instantiate OceanP2P from within the tests
- The only output from the test is ` ERROR: null` which is making it hard to debug. 
- OceanP2P is needed as part of other tests (such as testing the download) so this first step of initialising it from within the tests needs to be fixed. 
- Getting the same issue when running it locally and remotely.


src/test
```diff
@@ -0,0 +1 @@
+../tests/
```
jamiehewitt15, line 1: Another option is that we could just move the test folder within the src directory but I think this is fine for now.

jamiehewitt15 (commented):
It's great that the tests now work with instantiating OceanP2P. I suggest that we merge this so that we can move forward with the other tests. I can't approve it as I made the PR...

src/test
jamiehewitt15: If we moved the test directory within src it would have the advantage of making the import paths less confusing.

paulo-ocean: yeah, for now i just created a symbolic link from `src/test` to `tests` .. so we have the tests under ./src ... so they can get compiled.. 
I updated the package.json scritps + the imports on the tests (so we don't reference 'src') ... These get easier this way as well

When i did the change i was able to figure out a lot of import problems on the tests files, that we're not caught   before, as we were not compiling them before...

This was the only way i could make all this thing work... Mocha has definitively an issue with that OceanP2P file... i was stripping down lines of code, imports, etc.. but nothing worked .. got a lot of mixed and very strange results... Overall, i do not think anything is wrong with that file...but there is some bug on mocha or on the typescript loaders that just breaks there... i don't think its on our side at all.

bogdanfazakas: wouldn't it have been better to have a tsconfig for tests and run the build using it ? rather than having this reference and adding the tests to the build image ?

paulo-ocean: the issue is that if, the tests are not under 'src' they do not compile, and if they don't compile they don't work and we're back to square one :-) . That's the only way i could make it work
I'm curious, what's the problem with having the tests on src anyway? Its also a very common practice to have the test file right in the same folder of the component it should test. I've seen that in many projects as well

bogdanfazakas: from my point of view the only downside would be that compiled test files would be in our build folder all the time. If we are ok with this we can merge the PR, looks good, Nice job @paulo-ocean and @jamiehewitt15

bogdanfazakas (approved):

paulo-ocean (approved):
lets merge this so we can move on, and if we can improve later, we'll do

PR #165, opened 2023-12-14
Issue 133 download endpoint

jamiehewitt15:
Fixes # .

Changes proposed in this PR:

- 
- 
-


src/components/core/downloadHandler.ts
```diff
@@ -21,6 +21,17 @@ export async function handleDownload(
   task: DownloadTask,
   node: OceanP2P
 ): Promise<P2PCommandResponse> {
+  P2P_CONSOLE_LOGGER.logMessage(
+    'Download Request recieved with arguments: ' +
+      task.filesIndex +
+      task.documentId +
+      task.serviceIndex +
```
mariacarmina, line 28: Can we check if you mean `serviceIndex` which should be a Number or `serviceId`?

jamiehewitt15: Ocean.js only sends `serviceId` which is a string so I've renamed `serviceIndex` to `serviceId`

https://github.com/oceanprotocol/ocean.js/blob/da262f7498f8d2a307d1815b64bd6aa2787e2974/src/services/Provider.ts#L481

PR #167, opened 2023-12-14
bump p2p libs

alexcos20:
Update all libp2p libs


paulo-ocean (approved):
LGTM, seems to be working as before

PR #171, opened 2023-12-18
Create Handler class

### LINKED ISSUE #130: Create Handlers class
Based on [this discussion on Slack](https://oceanprotocol.slack.com/archives/C06435G1AN4/p1700767298880449), we decided that handlers need to be actual classes. Follow-up from #122

mariacarmina:
Fixes #130  .

Changes proposed in this PR:

- create Handler class
- create handlers classes which extends Handler class
- deleted command DOWNLOAD_URL
- created utils folder for extra functions that came in the handlers files.


mariacarmina:
I encountered some issues defining the handlers in separate files. In CI, ReferenceError kept popping up with this message: `ReferenceError: Cannot access 'Handler' before initialization`. So I tried to place the classses in the same file with main abstract class `Handler`, which apparently solved the problem.

src/components/core/handlers/handler.ts
```diff
@@ -0,0 +1,626 @@
...
+import { DDO } from '../../../@types/DDO/DDO.js'
+
+export abstract class Handler {
+  private config: OceanNodeConfig
+  // Put database separately because of async constructor
+  // that Database class has
+  private db: Database
+  private task: any
+  private p2pNode: OceanP2P
+  public constructor(task: any, config?: OceanNodeConfig, db?: Database) {
```
paulo-ocean, line 43: Hi @mariacarmina 
why is the `task` argument of type 'any' ? isn't this always something of type '`Command`'?
If you make the abstract class receive a 'Command' instead, you can on the subclasses, use the specific Command as parameter..
For instance, lets take `QueryHandler` class, as example:
Instead of :
```
export class QueryHandler extends Handler {
  public constructor(task: any, database: Database) { 
```
It could be:
```
export class QueryHandler extends Handler {
  public constructor(task: QueryCommand, database: Database) {
```
And the same applies for all the other classes..
There is no need for an extra function to check if the `task` argument has the expected type for the class, since you can provide that right on the constructor
In any case all these arguments were already verified on the `validateCommandAPIParameters`  when the request is received, before any call to `handleDirectCommands`... 

Another thing i don't understand is why we have most of the implementation of the handlers on the same `handler.ts` file.. while they also have related functions on separate files
Its much easier to maintain if we keep each handler logic on its own separate file, as it was..
And the `DownloadHandler` for instance, is the only one on a separate file ? and for this we can call `handle()` (which is common to all the handlers), but we also call h`andleDownloadUrlCommand()` ?

IMO, its very confusing

mariacarmina: Hello @paulo-ocean, I wanted to raise this question as well regarding Commands. In my first commits, task was Command type but since it's defined like an interface, the types that inherits Command interface are not recognized as Command type (the main type), it accepta only the specific type (QueryCommand etc.). So my question was if we decide to convert Command to actual classes (it was already a big refactor, so that's why I let the types as they were and left checks for task type instead).

For DownloadHandler, I put 2 handlers there due to ReferenceError: Cannot access 'Handler' before initialization. DownloadUrlCommand is used in DownloadCommand handler which leaded to the mentioned error. I tried to combine in the same class the both logic to make sure that the error doesn't occur anymore. I am open to try different approach for this one, as long as Reference error doesn't show up.

mariacarmina: In my previous commits I had DownloadUrlHandler class which broke the CI first.

mariacarmina: Possible solution: To integrate DownloadHandler inside handler.ts and create separately as it was before, DownloadUrlHandler within the same file -> have a huge handler file.

paulo-ocean:
> I encountered some issues defining the handlers in separate files. In CI, ReferenceError kept popping up with this message: `ReferenceError: Cannot access 'Handler' before initialization`. So I tried to place the classses in the same file with main abstract class `Handler`, which apparently solved the problem.

I understand, but that means there is an issue with the dependency cycle, which means we're trying to use something before it gets initialized. There is something out of order that would need to debugged and fixed, but not necessarily move everything around. Probably just something that would need to be exported from a different place

mariacarmina:
> > I encountered some issues defining the handlers in separate files. In CI, ReferenceError kept popping up with this message: `ReferenceError: Cannot access 'Handler' before initialization`. So I tried to place the classses in the same file with main abstract class `Handler`, which apparently solved the problem.
> 
> 
> 
> I understand, but that means there is an issue with the dependency cycle, which means we're trying to use something before it gets initialized. There is something out of order that would need to debugged and fixed, but not necessarily move everything around. Probably just something that would need to be exported from a different place

All of the classes extended Handler class and some handlers/util functions were used in other handlers and that caused. It's a bit strange because I only created a class which includes the main functionality and put "extends Handler" -> error.

src/components/core/handlers/handler.ts
paulo-ocean: i honestly still don't understand what's the issue
just to be clear, as an example:

instead of this:
```
export class GetDdoHandler extends Handler {
  public constructor(task: any, database: Database) {
    super(task, null, database)
    **if (!this.isGetDdoCommand(task)) {
      throw new Error(`Task has not QueryCommand type. It has ${typeof task}`)
    }**
  }

  **isGetDdoCommand(obj: any): obj is GetDdoCommand** {
    return typeof obj === 'object' && obj !== null && 'command' in obj && 'id' in obj
  }
```
you could have only:

```
export class GetDdoHandler extends Handler {
  public constructor(task: **GetDdoCommand**, database: Database) {
    super(task, null, database)
  }
```
It seems you're saying you cannot have 'task' argument as `GetDdoCommand` on the constructor, but right after the `super()` call you´re checking that. Am i missing something?? :-)
You can, and of course, update accordingly, on the other classes and parent Handler class
and the Commands can still be interfaces, why would they need to be classes? its just something that has certain fields, not everything needs to be a class

paulo-ocean:
> > > I encountered some issues defining the handlers in separate files. In CI, ReferenceError kept popping up with this message: `ReferenceError: Cannot access 'Handler' before initialization`. So I tried to place the classses in the same file with main abstract class `Handler`, which apparently solved the problem.
> > 
> > 
> > I understand, but that means there is an issue with the dependency cycle, which means we're trying to use something before it gets initialized. There is something out of order that would need to debugged and fixed, but not necessarily move everything around. Probably just something that would need to be exported from a different place
> 
> All of the classes extended Handler class and some handlers/util functions were used in other handlers and that caused. It's a bit strange because I only created a class which includes the main functionality and put "extends Handler" -> error.

What can i say, i find it all a bit  confusing @mariacarmina 
the abstract handler class  is also instantiating a new OceanP2P instance
`this.p2pNode = new OceanP2P(this.db, this.config)`
why is that?

paulo-ocean:
> > > I encountered some issues defining the handlers in separate files. In CI, ReferenceError kept popping up with this message: `ReferenceError: Cannot access 'Handler' before initialization`. So I tried to place the classses in the same file with main abstract class `Handler`, which apparently solved the problem.
> > 
> > 
> > I understand, but that means there is an issue with the dependency cycle, which means we're trying to use something before it gets initialized. There is something out of order that would need to debugged and fixed, but not necessarily move everything around. Probably just something that would need to be exported from a different place
> 
> All of the classes extended Handler class and some handlers/util functions were used in other handlers and that caused. It's a bit strange because I only created a class which includes the main functionality and put "extends Handler" -> error.

I would say the error was probably on the place where you defined the abstract class..
If  we were able to call the handler functions from `handleProtocolCommands`  (as on develop branch)
like this:
```
      ...
       response = await handleGetDdoCommand(this, task)
       response = await handleQueryCommand(this, task)
      ...
```
We should be able to also call the corresponding classes, if they were defined/exported on those corresponding files
The issue was probably that the super abstract class `Handler` was just defined on the wrong place

mariacarmina:
> > > > I encountered some issues defining the handlers in separate files. In CI, ReferenceError kept popping up with this message: `ReferenceError: Cannot access 'Handler' before initialization`. So I tried to place the classses in the same file with main abstract class `Handler`, which apparently solved the problem.
> 
> > > 
> 
> > > 
> 
> > > I understand, but that means there is an issue with the dependency cycle, which means we're trying to use something before it gets initialized. There is something out of order that would need to debugged and fixed, but not necessarily move everything around. Probably just something that would need to be exported from a different place
> 
> > 
> 
> > All of the classes extended Handler class and some handlers/util functions were used in other handlers and that caused. It's a bit strange because I only created a class which includes the main functionality and put "extends Handler" -> error.
> 
> 
> 
> I would say the error was probably on the place where you defined the abstract class..
> 
> If  we were able to call the handler functions from `handleProtocolCommands`  (as on develop branch)
> 
> like this:
> 
> ```
> 
>       ...
> 
>        response = await handleGetDdoCommand(this, task)
> 
>        response = await handleQueryCommand(this, task)
> 
>       ...
> 
> ```
> 
> We should also be able to call the corresponding classes, if they were defined/exported on those same corresponding files
> 
> The issue was probably that the super abstract class `Handler` was just defined on the wrong place
> 
> 

I will try also removing handlers folder and keep everything that belongs to handlers inside core folder and post the update.

paulo-ocean:
> > > > > I encountered some issues defining the handlers in separate files. In CI, ReferenceError kept popping up with this message: `ReferenceError: Cannot access 'Handler' before initialization`. So I tried to place the classses in the same file with main abstract class `Handler`, which apparently solved the problem.
> > 
> > 
> > > >
> > 
> > 
> > > >
> > 
> > 
> > > > I understand, but that means there is an issue with the dependency cycle, which means we're trying to use something before it gets initialized. There is something out of order that would need to debugged and fixed, but not necessarily move everything around. Probably just something that would need to be exported from a different place
> > 
> > 
> > > 
> > 
> > 
> > > All of the classes extended Handler class and some handlers/util functions were used in other handlers and that caused. It's a bit strange because I only created a class which includes the main functionality and put "extends Handler" -> error.
> > 
> > 
> > I would say the error was probably on the place where you defined the abstract class..
> > If  we were able to call the handler functions from `handleProtocolCommands`  (as on develop branch)
> > like this:
> > ```
> > 
> >       ...
> > 
> >        response = await handleGetDdoCommand(this, task)
> > 
> >        response = await handleQueryCommand(this, task)
> > 
> >       ...
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > We should also be able to call the corresponding classes, if they were defined/exported on those same corresponding files
> > The issue was probably that the super abstract class `Handler` was just defined on the wrong place
> 
> I will try also removing handlers folder and keep everything that belongs to handlers inside core folder and post the update.

cool. I would suggest that maybe is better that you pass the OceanP2P node on the constructors, with the 'this' (from the node, you can access the 'config' and the 'db'  on the handler, when you need those.
And probably make the `handle()` function receive the Command task as argument (similar what we have currently), instead of being on the class constructor... and for the `DownloadHandler`.. it can be either one or the other... `DownloadURL` or `DownloadTask`, 

Although the last would need refactor, as it does not follow the Command interface unlike all the others (maybe not for this task)

paulo-ocean:
well done @mariacarmina :+1:  , definitely much better.
there might be some refactor to do but i guess we could tackle that later.
couple of examples:
on constants.ts

```
export const HANDLERS_COMMANDS = [
  PROTOCOL_COMMANDS.DOWNLOAD,
  PROTOCOL_COMMANDS.ENCRYPT,
  PROTOCOL_COMMANDS.GET_DDO,
  PROTOCOL_COMMANDS.QUERY,
  PROTOCOL_COMMANDS.NONCE,
  PROTOCOL_COMMANDS.STATUS,
  PROTOCOL_COMMANDS.FIND_DDO,
  PROTOCOL_COMMANDS.GET_FEES
]
```
this is basically a duplicate of the existing `SUPPORTED_PROTOCOL_COMMANDS` array but with 'echo' removed
there is no need for another duplicate array and, besides that.. `PROTOCOL_COMMANDS.ECHO` is still supported so there no need for this.

Regarding the abstract class and the other handlers, there is also no need for the extra function to check the input type of the `handle()`. handle function could simply receive something of type 'Command', instead of 'any'.. All 'Command' extend that interface (or should :-) )

But, anyway.. we could do this later

thanks

paulo-ocean (approved):

alexcos20 (approved):

paulo-ocean:
btw, don't forget to check the conflicts  :-)

mariacarmina:
> Regarding the abstract class and the other handlers, there is also no need for the extra function to check the input type of the `handle()`. handle function could simply receive something of type 'Command', instead of 'any'.. All 'Command' extend that interface (or should :-) )

My first approach was that task should have Command type, but then I encounter some issues with types because some tasks did not have the fields from the Command interface and linter was complaining :(. But maybe this can be solved if we can convert them to classes in the future (just an idea). Thanks

PR #175, opened 2023-12-18
Improve ocean node class

### LINKED ISSUE #136: Enhance OceanNode class
Enhance OceanNode class to remove `setOceanNode` function and replace with a function that does the instanting all in one function.

Tried approaches:

- builder pattern
- put inside constructor code

Both of them failed due to mocha, not running the tests and receiving `ERROR: null`.

Follow up from https://github.com/oceanprotocol/ocean-node/pull/122#discussion_r1409278955

Maybe change the Database constructor since it is the one that throws the error in mocha testing.

paulo-ocean:
Fixes #136 

Changes proposed in this PR:

- refactor OceanNode class a bit
- add handlers registry support

(eventually will need some refactor once we finish #171 to connect the dots)


bogdanfazakas (approved):
lgtm

alexcos20 (approved):

PR #176, opened 2023-12-19
fix upnp nat traversal

alexcos20:
While debuging upnp nat traversal, I realized that mappingIP code is triggered too soon, before transportManager has all bind interfaces.

Just triggering it manually after node, and works like a charm:
Screenshot 2023-12-19 112220


paulo-ocean (approved):
LGTM

bogdanfazakas (approved):

PR #182, opened 2023-12-19
Making the database optional

### LINKED ISSUE #173: Make DB config optional
Now, is mandatory to have a database connection
But we might have use cases which does not require indexer&provider, so let's make it optional.
If it's not set, we should set hasIndexer=false, to disable it and disable the provider part as well

Use cases:
 - circuit relay for p2p
 - node  for car UI   (will not store anything, just forward commands to another nodes)

jamiehewitt15:
Fixes #173 

Changes proposed in this PR:

- Making the database optional


paulo-ocean (changes requested):
hi @jamiehewitt15 
'm not 100% sure this is quite ready yet.
For instance, we still have default .env values for dbConfig (the DB_URL) , so it still tries to connect nevertheless
If i shutdown the DB (barge) and also unset DB_URL , i still have indexer working (even we cannot connect to db there is still a default config for it)
Also, we have some handlers that are doing DB operations (like getDDO, query, download, etc) , IMO, might worth a double check if it does not break
thanks

paulo-ocean (approved):
LGTM.

Btw, we probably need to create some additional issues; 
For instance, the 'status' command always returns info about provider and indexer (even if not enabled cause there is no DB configured). IMO, this can be misleading, but that is not part of this issue

paulo-ocean:
fyi, created an issue for the status: [https://github.com/oceanprotocol/ocean-node/issues/203](https://github.com/oceanprotocol/ocean-node/issues/203)

PR #184, opened 2023-12-20
downgrade to node 18.19.0

alexcos20:
With Node 18, all my nodes are up for more than 10 hours
Also, Ilia's node was up yesterday for a couple of hours, no restarts

Also, added nodejs version to status command, so we can track it


paulo-ocean (approved):
lgtm

bogdanfazakas (approved):

PR #185, opened 2023-12-20
Feature/ Make RPCS env var optional

### LINKED ISSUE #172: Make env RPCS optional
Now, env RPCS is mandatory.
But we might have use cases which does not require indexer&provider, so let's make it optional.
If it's not set, we should set hasIndexer=false, to disable it.

bogdanfazakas:
Fixes #172  .

Changes proposed in this PR:

- 
- 
-


paulo-ocean (approved):
LGTM. :+1: 
btw, i think we will need to refactor something related with download (the error response on `P2PCommandResponse`), but will create a separate issue for that.. its not related with this task

alexcos20 (approved):

Issue #138, opened 2023-12-04
Setup / tear down specific environment for test suite


paulo-ocean:
**Is your feature request related to a problem? Please describe.**
Right now, some of our tests require that we setup specific a `process.env.VAR_NAME` with a given value. This affects (overrides) any other values that we have currently set locally). Running the local setup and running the tests should have a separate .env config so one does not break the other

**Describe the solution you'd like**
An utility class that saves current .env values (we should have a list of all the variables required by the nodes) with methods to setup the test environment (save current running config and setup new values) before running each test suite, and another one to tear down the test .env values and restore the original ones... There will be a dedicated .env file as well  with default values... Any test suite could just use the default values or explicitly override any

**Describe alternatives you've considered**
Just setting a specific .env for test environment, however, there are occasions were we might need specific values and a generic approach for all tests might not be the best alternative, instead we can (if we want) to override sepecific values, and restore them to the previous values after the tests

**Additional context**
This was detected while running some tests that were setting up specific values and overriding local setup, what was impacting not only the local run of the node but other tests as well, that relied on other values

PS: there is already some work done here [https://github.com/oceanprotocol/ocean-node/blob/develop/src/test/utils/utils.ts](https://github.com/oceanprotocol/ocean-node/blob/develop/src/test/utils/utils.ts)

alexcos20:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/S1---DApps-63f8b8c26e838c0017f34d23/issues/oceanprotocol/ocean-node/138?planning-poker) with Zenhub @an7e @bogdanfazakas @jamiehewitt15 @mariacarmina @paulo-ocean

paulo-ocean:
closed on `[https://github.com/oceanprotocol/ocean-node/pull/211](https://github.com/oceanprotocol/ocean-node/pull/211)`

PR #190, opened 2023-12-22
Feature/more_p2p_tests

alexcos20:
- Move bootstrap nodes to config
- add more p2p unit tests


paulo-ocean (approved):
LGTM. :+1: 
but i think we should consider this at some point: [https://github.com/oceanprotocol/ocean-node/issues/138](https://github.com/oceanprotocol/ocean-node/issues/138)
When we run our tests locally we're in many cases overriding existing process.env properties

mariacarmina:
There are some conflicts here

PR #195, opened 2024-01-02
Fix tests folder

### LINKED ISSUE #189: Get rid of symbolic link on tests - just have all under src/test folder
Right now it can get a bit confusing with the symbolic link from src/test to tests/
Since we agreed to have them on src/test so they compile and run, better just have that one folder

paulo-ocean:
Fixes #189 

Changes proposed in this PR:

- rm tests folder, move everything to src/test (no symlink)
- 
-


alexcos20 (approved):
lgtm

PR #197, opened 2024-01-03
refactor ocean node, clean commands + handlers

### LINKED ISSUE #181: Complete refactor of Ocean Node class with handlers support
Once #171 is merged
we need to do some refactor on OceanNode class and other parts (commands related), to incorporate the handler logic in a more structured way
related with #136

paulo-ocean:
Fixes #181 

Changes proposed in this PR:

- Clean handlers classes + protocol commands usage, refactor OceanNode class and a little bit on the handlers

- Main changes:
- new class `CoreHandlersRegistry`. its used to register all core handlers (easy to add more on demand). Some helper methods on the same file. Singleton instance.
- Ocean Node has a property of this type (`coreHandlers`), which is used to get the correct handler for a given `Command`. Basically uses the Singleton above. 

No more switch / case.. just get the handler from the OceanNode or Singleton (see `handleDirectProtocolCommands` and `handleProtocolCommands`)

- Moved all the logger instances that were scattered within multiple files, to a single file under: `/utils/logging/common.js `. Its easier to keep track of them and avoids issues with dependency cycling

(Note: There are many files changed but in most cases it was just these  "Logger" related imports)


src/components/core/coreHandlersRegistry.ts
```diff
@@ -0,0 +1,119 @@
...
+    this.registerCoreHandler(PROTOCOL_COMMANDS.DOWNLOAD, new DownloadHandler(node))
+    this.registerCoreHandler(PROTOCOL_COMMANDS.ENCRYPT, new EncryptHandler(node))
+    this.registerCoreHandler(PROTOCOL_COMMANDS.NONCE, new NonceHandler(node))
+    this.registerCoreHandler(PROTOCOL_COMMANDS.GET_DDO, new GetDdoHandler(node))
+    this.registerCoreHandler(PROTOCOL_COMMANDS.QUERY, new QueryHandler(node))
+    this.registerCoreHandler(PROTOCOL_COMMANDS.STATUS, new StatusHandler(node))
+    this.registerCoreHandler(PROTOCOL_COMMANDS.FIND_DDO, new FindDdoHandler(node))
+    this.registerCoreHandler(PROTOCOL_COMMANDS.GET_FEES, new FeesHandler(node))
+    this.registerCoreHandler(PROTOCOL_COMMANDS.ECHO, new EchoHandler(node))
+  }
```
an7e, line 65: please consider my previous variant where I suggested to add the `command ` parameter in the base class constructor to make for example `PROTOCOL_COMMANDS.ECHO` a part of the class, we will get several advantages from this, we will have the context of the command, for example in the future we will add validate to the classes in addition to the command handle, and also we can use an array for `registerCoreHandler` and it will iterate through the classes and register the necessary handlers itself

paulo-ocean: not sure if i follow you.. so for each received command you would create a new Handler class?

an7e: i meant
```
export abstract class Handler {
  private p2pNode: OceanP2P
  public command
  public constructor(p2pNode: OceanP2P, command) {
    this.p2pNode = p2pNode
  }
```

paulo-ocean: Ok, but the command parameters changes on every request, so what would be the advantage there?

src/components/P2P/index.ts
```diff
@@ -379,20 +367,20 @@ export class OceanP2P extends EventEmitter {
...
-    // direct message to self
-    // create a writable stream
-    // const outputStream = new Stream.Writable()
-    response.stream = new Stream.Writable()
-    // read from input stream to output one and move on
-    await handleDirectProtocolCommand.call(this, message, sink)
-
-    return response
-  }
+  // async sendToSelf(message: string, sink: any): Promise<P2PCommandResponse> {
```
bogdanfazakas, line 370: i think we can delete now

src/utils/constants.ts
```diff
@@ -29,30 +30,17 @@ export const SUPPORTED_PROTOCOL_COMMANDS: string[] = [
...
-  PROTOCOL_COMMANDS.QUERY,
-  PROTOCOL_COMMANDS.NONCE,
-  PROTOCOL_COMMANDS.STATUS,
-  PROTOCOL_COMMANDS.FIND_DDO,
-  PROTOCOL_COMMANDS.GET_FEES
-]
-
 export interface Command {
-  command: string
+  command: string // command name
```
bogdanfazakas, line 34: i was wondering if we should still keep this interfaces (won't add a comment for each of them 😄 ) here in the constants file or the types folder in some new files would be more suitable for them

bogdanfazakas (approved):
lgtm, some minor questions but not critical or urgent

src/utils/constants.ts
paulo-ocean: tbh, that idea crossed my mind too :-) i agree that these should be moved outside of constants

alexcos20 (approved):

jamiehewitt15 (approved):
Looks good to me

PR #198, opened 2024-01-03
Improve Ocean class tests

### LINKED ISSUE #158: Improve tests/unit/ocean.test.ts
Currently the tests in `tests/unit/ocean.test.ts` check that the values of `oceanNode.getP2PNode()` and `oceanNode.getIndexer()` and `oceanNode.getProvider()` are not null. However, the values of each of these is `undefined`. We need more comprehensive tests to check that each part of the node is initialised correctly. 
 
 For example, if you modify the tests to include logs: 
 
 ```
    if (config.hasP2P) {
      console.log('oceanNode.getP2PNode()', oceanNode.getP2PNode())
      // assert(oceanNode.getP2PNode()) // Currently the value is undefined so this assert will fail.
      expect(oceanNode.getP2PNode()).to.not.eql(null)
    }
    if (config.hasIndexer) {
      console.log('oceanNode.getIndexer()', oceanNode.getIndexer())
      expect(oceanNode.getIndexer()).to.not.eql(null)
    }
    if (config.hasProvider) {
      console.log('oceanNode.getProvider()', oceanNode.getProvider())
      expect(oceanNode.getProvider()).to.not.eql(null)
    }
```

The output is: 

```
 Status command tests
oceanNode.getP2PNode() undefined
oceanNode.getIndexer() undefined
oceanNode.getProvider() undefined

```

### PR DESCRIPTION
**Pull request by mariacarmina**
Fixes #158 .

Changes proposed in this PR:

- added getter methods where needed
- fix unit tests for Ocean class

### CODE REVIEW

File: src/components/Indexer/index.ts [hunk 1]
```diff
@@ -31,6 +31,14 @@ export class OceanIndexer {
     this.startThreads()
   }
 
+  public getSupportedNetworks(): RPCS {
```
bogdanfazakas, line 34: do we really need this 2 methods ?

mariacarmina: I used them for asserting in tests, so the answer is yes, I need them. They are getters so they can be useful in the future.

package.json
```diff
@@ -21,6 +21,7 @@
     "test": "npm run lint && npm run test:unit:cover && npm run test:integration:cover",
     "test:unit": "npm run build-tests && npm run mocha -- './dist/test/unit/**/*.test.js'",
     "test:unit:storage": "npm run mocha -- './dist/test/unit/storage.test.js'",
+    "test:unit:ocean": "npm run mocha -- './dist/test/unit/ocean.test.js'",
```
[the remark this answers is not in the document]
mariacarmina: Ok I can delete it, but not sure why is recommended to delete tests scripts from `package.json`. Instead of running the whole suite, maybe we can have some scripts that test certain files.

paulo-ocean (approved):
overall , LGTM. i don´t have any issues with having extra getters here and there :-) but i think that we should think about removing these individual tests (including `unit:storage`), its getting too big and we can always run the command manually,
That said, i think these tests are pretty much useless to be honest :-)
If we want to keep them its ok, but they are not really testing anything;
We create one `config`  object, pass it in the constructor to a bunch of different objects and then check if the properties match?? But of course they match, its the same object...
It like adding 1 + 1 and checking if its 2 :-)
I'm not saying they are wrong, but i don't see any added value

mariacarmina:
> overall , LGTM. i don´t have any issues with having extra getters here and there :-) but i think that we should think about removing these individual tests (including `unit:storage`), its getting too big and we can always run the command manually, That said, i think these tests are pretty much useless to be honest :-) If we want to keep them its ok, but they are not really testing anything; We create one `config` object, pass it in the constructor to a bunch of different objects and then check if the properties match?? But of course they match, its the same object... It like adding 1 + 1 and checking if its 2 :-) I'm not saying they are wrong, but i don't see any added value

I can remove the scripts for package.json, because I understood the reason to not populate useless with scripts and I'll remove them. For testing I wanted to test that the components were initialized correctly (and wanted to access some fields). The issue mentioned that we should have individual/granular test for OceanIndexer, OceanP2P and OceanProvider. If you have any suggestions regarding the assert statements, I am happy to add those.
Thank you @paulo-ocean for review!

src/components/Indexer/index.ts
bogdanfazakas: got it, ok having them if they actually bring value

bogdanfazakas (approved):

PR #200, opened 2024-01-04
Return the error for db operations

### LINKED ISSUE #150: Indexer should return a response when a did is not found
Currently, we have: 
```
  async retrieve(id: string) {
    try {
      return await this.provider
        .collections(this.schemas[0].name)
        .documents()
        .retrieve(id)
    } catch (error) {
      return null
    }
  }

``` 

I suggest we return the error. If the market calls the indexer and nothing is found it should be able to display an error message to the user. So we can change it to:

```
  async retrieve(id: string) {
    try {
      return await this.provider
        .collections(this.schemas[0].name)
        .documents()
        .retrieve(id)
    } catch (error) {
      return error
    }
  }

```
There will also need to be some additional changes needed to make sure that this doesn't break the existing functionality. 

Then we can test it with something like this:

```
const response = await database.ddo.retrieve(fakeDid)
assert(response.status == 404)
assert(response.message == "Not found")
```

mariacarmina:
Fixes #150  .

Changes proposed in this PR:

- 
- 
-

### DIFF
--- a/src/components/Indexer/eventProcessor.ts
+++ b/src/components/Indexer/eventProcessor.ts
@@ -96,7 +96,12 @@ export const processMetadataStateEvent = async (
       .update(getAddress(event.address) + chainId.toString(10))
       .digest('hex')
   try {
-    let ddo = await dbconn.ddo.retrieve(did)
+    let ddo = null
+    try {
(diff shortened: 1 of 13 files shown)


mariacarmina:
Closing this and creating a fresh one :)

Issue #52, opened 2023-10-18
Retry queue


AnaLoznianu:
- [ ]  start a thread for queue processing
 - [ ]  process retry queue when possible

PR #201, opened 2024-01-04
22 Command reindex and queue

### LINKED ISSUE #22: Command:   reIndex
Sometimes, it's needed to force indexer to reprocess an event or an entire transaction

TO DO:  

- [ ]  implement indexer core function reIndex , which takes an event (or all events ) from a transaction and inserts it in retry queue. From there, it will be picked up by indexer retry queue mechanism
- [ ]   expose it on http interface
- [ ]   expose it on p2p interface

Command description:

{
command: "reIndex",
txId: "0x123",
chainId: "100",
eventIndex: 1  // optional, if present, reprocess only this index
}

an7e:
Fixes #22 #52

Changes proposed in this PR:

- added command: "reIndex" on http and p2p interface
- added reindex thread
- added process queue reindex


src/components/Indexer/index.ts
```diff
@@ -74,6 +82,45 @@ export class OceanIndexer {
     }
   }
 
+  public async reindexThread(): Promise<void> {
```
bogdanfazakas, line 85: do we really need to start a new thread for each network for reindexing the events?
can't we reuse the existing network working thread ?

an7e: @bogdanfazakas we only use one worker for reindexing for all networks

src/components/Indexer/reindexThread.ts
```diff
@@ -0,0 +1,68 @@
+import { parentPort, workerData } from 'worker_threads'
+import { sleep } from '../../utils/util.js'
+import { RPCS } from '../../@types/blockchain.js'
+import { Blockchain } from '../../utils/blockchain.js'
+import { REINDEXER_LOGGER } from './index.js'
+import { LOG_LEVELS_STR } from '../../utils/logging/Logger.js'
+import { processChunkLogs } from './utils.js'
+
+export interface ReindexTask {
```
paulo-ocean, line 9: do we really need this? Its a duplicate of  utils/constants  `ReindexCommand`

src/components/core/reindexHandler.ts
```diff
@@ -0,0 +1,54 @@
...
+        }
+      }
+      const chainId: string = String(task.chainId)
+      if (!chainId) {
+        return {
+          stream: null,
+          status: { httpStatus: 400, error: 'Invalid parameter chainId' }
+        }
+      }
+      const eventIndex: number = Number(task.eventIndex)
```
paulo-ocean, line 37: If i understood correctly the issue, the `eventIndex` is optional, if present it only processes this specific event, and if not present, it will process all events of that transaction..
Here, we have an optional field, but then we try to convert to a Number, so if not present, instead of processing all events, it returns an error and processes none

src/components/Indexer/index.ts
```diff
@@ -10,7 +10,13 @@ import {
 } from '../../utils/logging/Logger.js'
 import { EVENTS } from '../../utils/index.js'
 import EventEmitter from 'node:events'
+import { ReindexTask } from './reindexThread.js'
 
+export const REINDEXER_LOGGER: CustomNodeLogger = getCustomLoggerForModule(
```
paulo-ocean, line 15: Better reuse the indexer logger, no need to create an additional logger, its the same component anyway, so probably is enough to just change the message (we already refactored to reduce loggers and here we're adding more again :-) )

src/components/Indexer/reindexThread.ts
```diff
@@ -0,0 +1,68 @@
...
+  chainId: string
+  eventIndex?: number
+}
+
+const { supportedNetworks } = workerData as { supportedNetworks: RPCS }
+
+const REINDEX_QUEUE: ReindexTask[] = []
+
+async function processReindex(): Promise<void> {
+  while (true) {
```
paulo-ocean, line 20: Do we really need an infinite loop? these are bad boys :-) 
IMO, this would be perfect for a regular setInterval() or recursive function with a timeout ... 
when we receive the command we just add to the queue (without processing), in my opinion would be better to "wake up" at given intervals and process whatever is left on the queue, instead of locking processing there 
my 2 cents

src/components/Indexer/reindexThread.ts
an7e: Yes, even though they look the same now, they have different meanings and may change, for example tomorrow we go to add one more parameter to the command
`interface ReindexCommand extends Command`

src/components/core/reindexHandler.ts
an7e: No, in any bad case the Number constructor returns NaN, further the NaN used as array index will just return undefined

src/components/Indexer/reindexThread.ts
paulo-ocean: and?

src/components/Indexer/reindexThread.ts
an7e: Unfortunately everything about threads in Node js is a very simplified high-level abstraction, we have almost no control over anything, on the other hand we don't need to bother with mutexes, atomic ops and joins

src/components/core/reindexHandler.ts
paulo-ocean: ok, although not very "readable"

src/components/core/reindexHandler.ts
```diff
@@ -0,0 +1,54 @@
...
+      'chainId' in obj
+    )
+  }
+
+  async handle(task: any): Promise<P2PCommandResponse> {
+    try {
+      if (!this.isReindexCommand(task)) {
+        throw new Error(`Task has not ReindexCommand type. It has ${typeof task}`)
+      }
+      const txId: string = String(task.txId).toLowerCase()
```
paulo-ocean, line 23: Shouldn't we validate the params on `validateCommands.ts` , before passing stuff to the task? as we do for other commands..

an7e: it's weird, but it wasn't in my version
I'll add it now

an7e: I may have accidentally erased it when I was merging

an7e: done

paulo-ocean (approved):

jamiehewitt15 (approved):
Looks good to me

PR #202, opened 2024-01-04
Feature/ Refactor indexer

### LINKED ISSUE #199: Indexer Refactor
- [ ] research & find a better solution for the db connection inside worker threads
- [ ] make eventsHandler a class and reuse code
- [ ] make block query interval configurable and get rid of the hardcoded 
value

bogdanfazakas:
Fixes #199  .

Changes proposed in this PR:

- 
- 
-


src/components/Indexer/processor.ts
```diff
@@ -0,0 +1,314 @@
...
+  ): Promise<any> {
+    INDEXER_LOGGER.logMessage(`Processing metadata state event...`, true)
+    const decodedEventData = await this.getEventData(
+      provider,
+      event.transactionHash,
+      ERC20Template.abi
+    )
+    const metadataState = parseInt(decodedEventData.args[1].toString())
+    INDEXER_LOGGER.logMessage(`Processed new metadata state ${metadataState} `, true)
+    const dbconn = await new Database(await (await this.getConfiguration()).dbConfig)
```
an7e, line 107: Let's make a dbconn parameter in the constructor that requires a ready-made database instance so that we can write this kind of code
import Database
```
const DB = await new Database(await (await this.getConfiguration()).dbConfig)

somewhere
const eventProcess = BaseEventProcessor(DB)

class BaseEventProcessor {
  protected config: OceanNodeConfig
  protected dbconn: Database
  constructor(dbconn) {
    this.config = null // config would even be redundant in a case like this
    this.dbconn = dbconn
  }
```

src/components/Indexer/processor.ts
```diff
@@ -0,0 +1,353 @@
...
+      )
+      const byteArray = getBytes(decodedEventData.args[4])
+      const utf8String = toUtf8String(byteArray)
+      const ddo = JSON.parse(utf8String)
+      ddo.datatokens = this.getTokenInfo(ddo.services)
+      INDEXER_LOGGER.logMessage(
+        `Processed new DDO data ${ddo.id} with txHash ${event.transactionHash} from block ${event.blockNumber}`,
+        true
+      )
+      this.createOrUpdateDDO(ddo, EVENTS.ORDER_REUSED)
```
mariacarmina, line 122: For me it's unclear why we pass `EVENTS.ORDER_REUSED` as method for createOrUpdateDDO. Should it be changed or do you have another reason behind for this?

src/components/Indexer/processor.ts
```diff
@@ -0,0 +1,353 @@
...
+
+  public async createOrUpdateDDO(ddo: any, method: string): Promise<void> {
+    try {
+      const { ddo: ddoDatabase } = await this.getDatabase()
+      const saveDDO = await ddoDatabase.update({ ...ddo })
+      INDEXER_LOGGER.logMessage(
+        `Saved or updated DDO  : ${saveDDO.id} from network: ${this.networkId} `
+      )
+      // emit event
+      if (method === EVENTS.METADATA_CREATED) {
```
mariacarmina, line 89: If we include update metadata event in `createOrUpdateDDO`, we need to also check for `EVENTS.METADATA_UPDATED`, `EVENTS.METADATA_STATE` and order events for logging the right event. Is it correct?

mariacarmina (commented):
Couple of comments for better understanding, thanks!
Can we also update the description of the PR to see what we include in the refactor?

src/components/Indexer/processor.ts
paulo-ocean: i think the main reason why we are passing the event is mainly, because on a case of adding a new DDO to DB, we need to advertise to the network that we are serving as providers of that ID (emit an event).
And that can be only done after a successful DB operation.   So, its either passing the event to that function, or do it outside, but for that the `createOrUpdateDDO` would have to return something, like a boolean for instance.
In any case, i cannot see any place anymore where we are passing `METADATA_CREATED` to that function, so it seems to me that this part bellow is already broken:
```
if (method === EVENTS.METADATA_CREATED) {
        INDEXER_DDO_EVENT_EMITTER.emit(EVENTS.METADATA_CREATED, saveDDO.id)
      }
```

paulo-ocean, line 89: @bogdanfazakas is this ever called? I don´t think we are passing this event anymore, so no event will ever be emmited

bogdanfazakas: good catch thx, there is a mistake here https://github.com/oceanprotocol/ocean-node/pull/202/files#diff-18c89330be3d05a89e8fa07b7e16560745d6ad5af06e8e7e4e81007f7b1f6961R122 will update it to be ` EVENTS.METADATA_CREATED` instead of `EVENTS.ORDER_REUSED` but should be called

bogdanfazakas: yep of this check is to advertise new DDO's by using the emit method, and that is the reason we have the check here after the ddo is saved and we advertise only new ddo's not updated ones

src/components/Indexer/processor.ts
bogdanfazakas: good point, the method param can also be optional, 
but think it would be good to know also when updating the the ddo the event that triggered it within the logs, so will update this to reflect it `INDEXER_LOGGER.logMessage(
        `Saved or updated DDO  : ${saveDDO.id} from network: ${this.networkId} triggered by ${method} event !`
      )`

paulo-ocean:
btw @bogdanfazakas i see that the event is being emited now, but its not "listened" anymore on the P2P node... so something else must have changed on the refactor

bogdanfazakas:
> btw @bogdanfazakas i see that the event is being emited now, but its not "listened" anymore on the P2P node... so something else must have changed on the refactor

hmm should be there just updated the import path in the P2P index file
https://github.com/oceanprotocol/ocean-node/pull/202/files#diff-f1ac505a4ab708dae670df1269cd81836276455e82e45e296447566142268a7aL126

src/components/Indexer/index.ts
```diff
@@ -1,14 +1,9 @@
...
 import { RPCS, SupportedNetwork } from '../../@types/blockchain.js'
+import { ReindexTask } from './crawlerThread.js'
 import { LOG_LEVELS_STR } from '../../utils/logging/Logger.js'
-import { EVENTS } from '../../utils/index.js'
-import EventEmitter from 'node:events'
 import { INDEXER_LOGGER } from '../../utils/logging/common.js'
-import { ReindexTask } from './crawlerThread.js'
-
-// emmit events for node
-export const INDEXER_DDO_EVENT_EMITTER = new EventEmitter()
```
paulo-ocean, line 11: The event emiter should stay here, otherwise it does not work

paulo-ocean (changes requested):

src/components/Indexer/index.ts
bogdanfazakas: good catch, might be something related to working threads, let me dig more into this

mariacarmina (approved):
lgtm

jamiehewitt15 (commented):
Looks good in general, but it would be good to have unit tests that instantiate the new classes and then call the public methods.

paulo-ocean:
its still not working @bogdanfazakas 
Are you posting the mesaage? This part is not getting called
```
worker.on('message', (event: any) => {
        if (event.method === EVENTS.METADATA_CREATED) {
          this.advertiseDDO(event.network, event.data, event.method)
        } 
})
```
Also you are already checking if `event.method === EVENTS.METADATA_CREATED)` for calling `advertiseDDO` but you're doing again the same check there, is that really necessary? Or the fact that it reruns a Promise?
public async advertiseDDO(network: number, ddo: any, method: string): Promise<void> {
```
    if (method === EVENTS.METADATA_CREATED) {
      INDEXER_LOGGER.logMessage(
        `Advertising new DDO : ${ddo.id} from network: ${network} `
      )
      INDEXER_DDO_EVENT_EMITTER.emit(EVENTS.METADATA_CREATED, ddo.id)
    }
  }
```
thanks

paulo-ocean (commented):
still not working, please see my previous comment

bogdanfazakas:
> still not working, please see my previous comment



> its still not working @bogdanfazakas Are you posting the message? This part is not getting called:
> 
> ```
> worker.on('message', (event: any) => {
>         if (event.method === EVENTS.METADATA_CREATED) {
>           this.advertiseDDO(event.network, event.data, event.method)
>         } 
> })
> ```
> 
> Also you are already checking if `event.method === EVENTS.METADATA_CREATED)` for calling `advertiseDDO` but you're doing again the same check there, is that really necessary? Or the fact that it returns a Promise?
> 
> ```
> public async advertiseDDO(network: number, ddo: any, method: string): Promise<void> {
>     if (method === EVENTS.METADATA_CREATED) {
>       INDEXER_LOGGER.logMessage(
>         `Advertising new DDO : ${ddo.id} from network: ${network} `
>       )
>       INDEXER_DDO_EVENT_EMITTER.emit(EVENTS.METADATA_CREATED, ddo.id)
>     }
>   }
> ```
> 
> PS: you can try to use the ocean-cli to publish with the node running and double check it thanks

Sorry my mistake, there was a typo `processBlocks` instead of `processedBlocks` 😮‍💨

jamiehewitt15:
> adding this unit tests is a bit complex since it requires a lot of mocked data, we need to mock all database interactions and the contract calls as well, i did the database mocks but the contact mocks require more work, the other way around with adding barge and having actual contract calls like we have in ocean.js for eg is somehow breaking the concepts of unit tests. let's have a talk at todays meetings and see if we need to spend more time on it or we should focus on something else,

Yeah, that's fair enough.

paulo-ocean:
alright, working now :+1:

paulo-ocean (approved):

PR #208, opened 2024-01-08
Error treatment for db operations

### LINKED ISSUE #150: Indexer should return a response when a did is not found

mariacarmina:
Fixes #150  .

Changes proposed in this PR:

- 
- 
-


mariacarmina:
UPDATE: kept only error logs at the database level for each operation. My previous commits focused to return an object (DatabaseError) in case of error on the NonceDatabase. Each integration run was failing with `stream not present` error [from download tests](https://github.com/oceanprotocol/ocean-node/actions/runs/7451806390/job/20273696185) due to the fact that previous nonce from a consumer address (that means whenever `checkNonce` was called in donwload handler) could not be found.

jamiehewitt15 (approved):
Looks good to me!

alexcos20 (approved):

PR #210, opened 2024-01-09
Integrate shacl schemas.

### LINKED ISSUE #206: Integrate Schacl schema validation
- have a folder that contains multiple schema definitions
- load list of schemas

mariacarmina:
Fixes #206  .

Changes proposed in this PR:

- added shacl schemas
- create validation function which will be used in `validateDDO` command
- created tests


mariacarmina:
In this PR, @alexcos20, there is the validation function (+ tests for it) against a certain shacl schema. Let me know if I should also integrate the validateDDO command or leave the command in a separate task.  And also if you have any suggestions for this current implementation, let me know.
Thanks!

src/test/unit/indexer/validation.test.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  after(() => {
+    // Restore original local setup / env variables after test
+    process.env.PRIVATE_KEY = ORIGINAL_PRIVATE_KEY
+    process.env.IPFS_GATEWAY = ORIGINAL_IPFS_GATEWAY
+    process.env.ARWEAVE_GATEWAY = ORIGINAL_ARWEAVE_GATEWAY
+    process.env.RPCS = ORIGINAL_RPCS
+  })
+
+  it('should pass the validation', async () => {
```
jamiehewitt15, line 35: It would be good to add some additional tests here so that we test both a valid and invalid DDO for each of the schema types.

mariacarmina: At the moment for DDO examples are using only 4.1.0 schema. I'll add other examples then.

jamiehewitt15: Cool thanks

mariacarmina: I have an update regarding you comment @jamiehewitt15. I'll note here the changes for each version of the SHACL schemas.
For version 4.1.0 - already created tests and the test for `should not fail` are applicable for all schemas.
Version 4.3.0 - had only constraints updates, `maxLength` was added to each field to have a limit for string field (I personally don't think it will bring more value to the test)
Version 4.5.0 - allow datasets with 0 services. [(added in the last commit)](https://github.com/oceanprotocol/ocean-node/pull/210/commits/711049413e620cfb562ae1a38d9e07c995240d22)

What I suggest is maybe to check the error messages, since the changes between schemas are not huge. What do you think?

jamiehewitt15: Sounds good, yeah it makes sense to check the error messages

src/test/unit/indexer/validation.test.ts
```diff
@@ -0,0 +1,57 @@
+import { incorrectDDO, DDOExample, ddov5 } from '../../data/ddo.js'
+import { validateObject } from '../../../components/core/utils/validateDdoHandler.js'
+
+import { expect } from 'chai'
+
+// avoid override local setup / env variables
+const ORIGINAL_PRIVATE_KEY = process.env.PRIVATE_KEY
```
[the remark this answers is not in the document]
mariacarmina: Thanks for reminder, I updated based on Ocean class tests (`ocean.test.ts` file).

src/test/unit/indexer/validation.test.ts
mariacarmina: I completed tests for checking the error messages.

jamiehewitt15 (approved):
Looks good to me

src/components/core/utils/validateDdoHandler.ts
```diff
@@ -0,0 +1,167 @@
...
+  LOG_LEVELS_STR,
+  defaultConsoleTransport,
+  getCustomLoggerForModule
+} from '../../../utils/logging/Logger.js'
+
+const CURRENT_VERSION = '4.5.0'
+const ALLOWED_VERSIONS = ['4.1.0', '4.3.0', '4.5.0']
+
+// Project cannot be buildt if the logger is imported, so created one locally
+export const SCHEMA_CONSOLE_LOGGER: CustomNodeLogger = getCustomLoggerForModule(
```
paulo-ocean, line 24: Why another Logger instance? We should just use one of the existing ones. Should be enough 1 per component/module (core, p2p, indexer, db, provider, etc). Not per command. On one hand we're trying to reduce them, on the other we keep adding more ?
Btw, on the other PR related with Logger, i already renamed the `STATUS_CONSOLE_LOGGER` to be `CORE_LOGGER`,  and we can use that on everything 'components/core'
thanks

src/components/core/utils/validateDdoHandler.ts
```diff
@@ -0,0 +1,167 @@
...
+  })
+
+  fileStream.on('error', (error: Error) => {
+    SCHEMA_CONSOLE_LOGGER.logMessage(`Error reading RDF file: ${error}`, true)
+  })
+
+  // create a validator instance for the shapes in the given dataset
+  const validator = new shaclEngine.Validator(dataset, {
+    factory: rdfDataModel
+  })
```
an7e, line 142: it could be dangerous
you create a stream, attach listeners on data and then immediately use dataset, where is the guarantee that everything is read?
please consider an example
**readStream** function in our **util.ts** file
or you can make it even shorter
```
const readStream = fs.createReadStream(file);

const chunks = [];
for await (const chunk of readStream) {
    chunks.push(chunk);
}

console.log(Buffer.concat(chunks));
```

src/components/core/utils/validateDdoHandler.ts
mariacarmina: I created another logger instance, because I received errors at building step if I was trying to export from different files. I'll try again with CORE_LOGGER as you did in your PR and check if building step can be done successfully.

src/components/core/utils/validateDdoHandler.ts
mariacarmina: Thank you so much @an7e for suggestion. Will do!

src/components/core/utils/validateDdoHandler.ts
paulo-ocean: What do you mean @mariacarmina ? The logger imports on this file:
```
import {
  CustomNodeLogger,
  LOGGER_MODULE_NAMES,
  LOG_LEVELS_STR,
  defaultConsoleTransport,
  getCustomLoggerForModule
} from '../../../utils/logging/Logger.js'
```
Are exactly the same imports used on `common.js` , so if it works on one side it works on the other. Don't think  that was not the root cause. In any case, i already tried and it works just fine. There is absolutely no need for this extra logger here

mariacarmina: I have a [certain run](https://github.com/oceanprotocol/ocean-node/actions/runs/7489790902/job/20387166230) that failed on it and the logger was exported :), that's why I wrote that, I'll put it again and remove this extra logger.

paulo-ocean: that error was pretty clear :-) . 
check [https://github.com/oceanprotocol/ocean-node/pull/210/commits/1a435a420104178a20672253f7438d79b00fa972](https://github.com/oceanprotocol/ocean-node/pull/210/commits/1a435a420104178a20672253f7438d79b00fa972)
You were trying import the INDEXER_LOGGER from `Indexer/index.js `instead of` logging/common.ts` where it is exported. Nothing to do with dependency cycles here

src/components/core/utils/validateDdoHandler.ts
mariacarmina: I think that implementation is also in `streamToString` function also from utils.ts.

src/components/core/utils/validateDdoHandler.ts
```diff
@@ -0,0 +1,155 @@
...
+    if ('metadata' in obj && attr in obj.metadata && !isIsoFormat(obj.metadata[attr])) {
+      extraErrors.metadata = `${attr} is not in ISO format.`
+    }
+  })
+
+  if (!chainId) {
+    extraErrors.chainId = 'chainId is missing or invalid.'
+  }
+
+  if (!nftAddress || nftAddress === '' || !isAddress(nftAddress.toLowerCase())) {
```
an7e, line 101: there are two useful functions isAddress and getAddress
it is better to use getAddress because it does both checks and transformation to the standard form of address in general
keep in mind that if the address is bad, it will throw an error rather than return false or null
https://docs.ethers.org/v6/api/address/#getAddress

mariacarmina: Sure, thank you!

paulo-ocean (approved):
LGTM

PR #211, opened 2024-01-09
Feature improve test dotenv

### LINKED ISSUE #138: Setup / tear down specific environment for test suite

paulo-ocean:
Fixes #138 

Changes proposed in this PR:

- Helper functions on `test/utils/utils.ts` to setup/teardown environment for tests (save  / override existing process.env variables and restore afterwards)
- use mocha root hooks (`test/utils/hooks.ts` => beforeAll() / affterAll() ), to setup enviroment before running any test and teardown after tests have runned 
- make usage of local hooks (before() / after() ) on test files overriding process.env vars

The goal is to guarantee that no required variables are forgotten when running the test suite, and when we explicitly override something,  the environment is always restored to the previous state, so our test do not mess up the env , 
even if we forget to explicitly handle it ( although ideally we should try to use the hooks when our test directly changes existing config vars)

So, now, it is possible to set your env/config, and alternate between running the node and/or the test suite without any configuration interferences between each other


mariacarmina (approved):
lgtm

jamiehewitt15 (approved):
Looks good to me

PR #212, opened 2024-01-10
decrypt ddo command

### LINKED ISSUE #53: Command: decryptDDO
Implement deencryptDDO command  (used by other nodes when indexing)

TO DO:

- [ ]  implement core function decryptDDO (with access control)
- [ ]  expose it on http interface
- [ ]  expose it on p2p interface

Command description:

{
command: "decryptDDO",
decrypterAddress,
chainId,
transactionId,
dataNftAddress,
encryptedDocument,
flags.
documentHash,
nonce,
signature
}

an7e:
Fixes #53

Changes proposed in this PR:

- decrypt ddo with transactionId
- decrypt ddo with encryptedDocument, flags, documentHash
- parameter checks
- integration tests


mariacarmina:
Hello @an7e, is this draft PR still work in progress? I've seen that you have some conflicts on this branch. Thanks!

bogdanfazakas:
Hi Ilia, i think the following is missing from the task definition

- [ ]  expose it on http interface

paulo-ocean:
> Hi Ilia, i think the following is missing from the task definition
> 
> * [ ]   expose it on http interface

It is via directCommand, 
It looks good to me unless we are talking about the legacy API/URL ?

src/components/core/ddoHandler.ts
```diff
@@ -18,13 +20,257 @@ import { sleep, readStream } from '../../utils/util.js'
...
+import { decrypt } from '../../utils/crypt.js'
+import { createHash } from 'crypto'
 
 const MAX_NUM_PROVIDERS = 5
 // after 60 seconds it returns whatever info we have available
 const MAX_RESPONSE_WAIT_TIME_SECONDS = 60
 // wait time for reading the next getDDO command
 const MAX_WAIT_TIME_SECONDS_GET_DDO = 5
 
+export class DecryptDdoHandler extends Handler {
```
paulo-ocean, line 37: Please add the handler to `CoreHandlersRegistry` constructor
i noticed the NonceHandler(node) appears 2 time in that constructor (a bad merge of mine last week) so please fix that while you are at it.
thanks

paulo-ocean:
> > > Hi Ilia, i think the following is missing from the task definition
> > > 
> > > * [ ]     expose it on http interface
> > 
> > 
> > It is via directCommand, It looks good to me unless we are talking about the legacy API/URL ?
> 
> i was referring to the legacy API 😄 , though that was the meaning in the task definition

Well, honestly i don't know for sure, i would probably just do the directCommand as well, but you might be right, we do have "encrypt" on the provider routes so i guess the "decrypt" makes sense after all

src/components/core/ddoHandler.ts
```diff
@@ -18,13 +20,257 @@ import { sleep, readStream } from '../../utils/util.js'
...
+            error: `Decrypt DDO: nonce value is not a number`
+          }
+        }
+      }
+
+      const node = this.getP2PNode()
+      const dbNonce = node.getDatabase().nonce
+      const existingNonce = await dbNonce.retrieve(decrypterAddress)
+
+      if (existingNonce && existingNonce.nonce === nonce) {
```
paulo-ocean, line 73: hi @an7e
I'm not 100% sure of this logic implemented here.
What if the nonce on the request is lower than the one on DB?
You're just checking if is different, and if so, proceed, otherwise return error 

I'm not sure if this decrypt logic here is supposed to have something in common with the nonce handler functions or not, honestly.. .
But i remember than on the nonce, if the nonce did not existed on DB we created it with value 0 ... Then on the following requests, we would check if it was bigger than the existing one. Shouldn't we be using the core functions related with nonce here?? the ones on` core/utils/nonceHandler.ts` for instance?  Seems to me that we are just creating a new logic here without using the nonce logic that we already implemented..

src/utils/config.ts
```diff
@@ -62,6 +62,42 @@ function getSupportedChains(): RPCS {
   return supportedNetworks
 }
 
+function getAuthorizedDecrypters(): string[] {
+  if (
+    !process.env.AUTHORIZED_DECRYPTERS ||
```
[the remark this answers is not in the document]
an7e: @paulo-ocean sure, added

src/components/core/ddoHandler.ts
an7e: @paulo-ocean I missed it while merging branches and solving conflits

src/components/core/ddoHandler.ts
an7e: I was relying on the provider's logic
https://github.com/oceanprotocol/provider/blob/4cd244bc18565c0d7bf631f97b584cbd199f0d03/ocean_provider/routes/decrypt.py
by logic, there should be no requests with the same nonce

paulo-ocean: > I was relying on the provider's logic https://github.com/oceanprotocol/provider/blob/4cd244bc18565c0d7bf631f97b584cbd199f0d03/ocean_provider/routes/decrypt.py by logic, there should be no requests with the same nonce

fair enough, you're right. Honestly i don't know how much of that provider code is buggy or not... looking at that code there is also the required "signature", that is never validated , strange

src/components/core/coreHandlersRegistry.ts
```diff
@@ -48,12 +48,12 @@ export class CoreHandlersRegistry {
   private static instance: CoreHandlersRegistry
   // map of handlers registered
   private coreHandlers: Map<string, Handler> = new Map<string, Handler>()
+
   // private readonly node: OceanP2P
   private constructor(node: OceanP2P) {
     // implement core handlers
-    this.registerCoreHandler(PROTOCOL_COMMANDS.DOWNLOAD, new DownloadHandler(node))
```
paulo-ocean, line 54: Where is the download handler now?

paulo-ocean (changes requested):

src/components/core/coreHandlersRegistry.ts
an7e: @paulo-ocean it's very funny, but it's also sad that the tests didn't catch it

src/components/core/ddoHandler.ts
```diff
@@ -18,13 +20,257 @@ import { sleep, readStream } from '../../utils/util.js'
...
+          }
+        }
+      }
+
+      if (!config.authorizedDecrypters.includes(decrypterAddress)) {
+        P2P_CONSOLE_LOGGER.logMessage('Decrypt DDO: Decrypter not authorized', true)
+        return {
+          stream: null,
+          status: {
+            httpStatus: 400,
```
paulo-ocean, line 112: on the provider code this is a 403

src/components/core/ddoHandler.ts
```diff
@@ -18,13 +20,257 @@ import { sleep, readStream } from '../../utils/util.js'
...
+      }
+
+      const templateContract = new ethers.Contract(
+        dataNftAddress,
+        ERC721Template.abi,
+        signer
+      )
+      const metaData = await templateContract.getMetaData()
+      const metaDataState = metaData[2]
+      if (
```
paulo-ocean, line 206: same thing here, on the provider, any of these 3 states return a 403 error code.
besides only  
```
MetadataState.ACTIVE,
MetadataState.TEMPORARILY_DISABLED,
MetadataState.UNLISTED,
```
seem valid, and the others (excluding the ones that trigger the 403)  should return 400.
This logic here does not match provider one

src/components/core/ddoHandler.ts
an7e: I wasn't sure if I should do that, but I fixed it

src/components/core/ddoHandler.ts
an7e: I wasn't sure if I should do that, but I fixed it

paulo-ocean: ok, but then i don't understand the task.. 
are we doing the same logic or just half of it? :-)

paulo-ocean (approved):
lgtm

src/components/httpRoutes/validateCommands.ts
```diff
@@ -100,6 +100,20 @@ export function validateCommandAPIParameters(requestBody: any): ValidateParams {
           reason: 'Missing required parameter(s): "txId","chainId"'
         }
       }
+    } else if (command === PROTOCOL_COMMANDS.DECRYPT_DDO) {
+      if (
+        !requestBody.decrypterAddress ||
+        !requestBody.chainId ||
+        !requestBody.nonce ||
+        !requestBody.signature
```
bogdanfazakas, line 108: i see signature beeing required but we never seem to use in the decrypt method, do we really need it ? i know we have it in old provider but not sure if used there as well, also the tests seem to just pass a random string value

src/components/httpRoutes/provider.ts
```diff
@@ -8,9 +8,28 @@ import { LOG_LEVELS_STR } from '../../utils/logging/Logger.js'
 import { PROTOCOL_COMMANDS } from '../../utils/constants.js'
 import { EncryptHandler } from '../core/encryptHandler.js'
 import { HTTP_LOGGER } from '../../utils/logging/common.js'
+import { DecryptDdoHandler } from '../core/ddoHandler.js'
 
 export const providerRoutes = express.Router()
 
+providerRoutes.post('/decrypt', async (req, res) => {
```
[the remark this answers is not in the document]
an7e: @jamiehewitt15 `httpRoutes.use('/api/services/', providerRoutes)`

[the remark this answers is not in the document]
jamiehewitt15: Ah yeah, sorry I missed that

src/components/core/ddoHandler.ts
```diff
@@ -27,6 +36,263 @@ const MAX_RESPONSE_WAIT_TIME_SECONDS = 60
...
+      )
+      const factoryContract = new ethers.Contract(
+        factoryAddress,
+        ERC721Factory.abi,
+        signer
+      )
+      const dataNftAddress = ethers.getAddress(task.dataNftAddress)
+      const factoryListAddress = await factoryContract.erc721List(dataNftAddress)
+      // console.log('factoryContract', isDatatokenDeployed)
+      // console.log('artifactsAdresses', artifactsAdresses)
```
jamiehewitt15, line 135:
```suggestion
```

src/components/core/ddoHandler.ts
```diff
@@ -27,6 +36,263 @@ const MAX_RESPONSE_WAIT_TIME_SECONDS = 60
...
+
+      return {
+        stream: Readable.from(decryptedDocument.toString()),
+        status: { httpStatus: 201 }
+      }
+    } catch (error) {
+      P2P_CONSOLE_LOGGER.logMessage(`Decrypt DDO: error ${error}`, true)
+      return {
+        stream: null,
+        status: { httpStatus: 500, error: 'Decrypt DDO: Unknown error: ' }
```
jamiehewitt15, line 290: Why not return `error.message`?

src/components/httpRoutes/provider.ts
```diff
@@ -8,9 +8,28 @@ import { LOG_LEVELS_STR } from '../../utils/logging/Logger.js'
...
+    })
+    if (result.stream) {
+      const decryptedData = await streamToString(result.stream as Readable)
+      res.status(201).send(decryptedData)
+    } else {
+      res.status(result.status.httpStatus).send(result.status.error)
+    }
+  } catch (error) {
+    HTTP_LOGGER.log(LOG_LEVELS_STR.LEVEL_ERROR, `Error: ${error}`)
+    res.status(500).send('Internal Server Error')
```
jamiehewitt15, line 29: Might also be useful to return `error.message` here as well: 

```suggestion
    res.status(500).send(`Internal Server error: ${error.message}`)
```

src/components/core/ddoHandler.ts
an7e: I was afraid that if any sensitive information got into output

src/components/httpRoutes/provider.ts
an7e: I was afraid that if any sensitive information got into output

src/components/httpRoutes/validateCommands.ts
an7e: we're going to implement it

jamiehewitt15 (approved):
Looks good to me

PR #217, opened 2024-01-11
Issue 105 fileinfo

### LINKED ISSUE #105: Storage: fileinfo
fileInfo command will return info about a certain file.
It has the following purposes:
- allow users to observe details about a file (type, length, name, type of storage)
- allow platforms to prevent orders if file is not accessible
- allow multiple parties to compute checksums (used in c2d flows)

Command description:
- for published assets
```json
{
command: "fileInfo",
node: "NodeB ID", // optional, use only that node
did: "did:op:123",
serviceId: "123",
fileIndex: "0"  // optional, narrow results to a single file,
checksum: true // optional, compute sha256 checksum of file if required
}
```
- for new assets
```json
{
command: "fileInfo",
node: "NodeB ID", // optional, use only that node
file:  { ... fileObject ... }
checksum: true // optional, compute sha256 checksum of file if required
}
```

Returns an array of objects (for each file):
```json
[
    {
        "valid": true,   (if url is accesible, ipfs cid exists, etc.  If yes, next fields are populated)
        "contentLength":"1161",
        "contentType":"application/json",
        "name": if available & secure
        "type":  [url, ipfs, arweave, etc]
        "checksumType":  "sha256",
        "checksum": ".."
    },...
]
```

Implementation:
 - extend storage class with fileInfo function (each storage type might have it's own implementation)
  - implement core command
  - expose on http
  
  References:
   - file object:  https://docs.oceanprotocol.com/developers/storage
   - existing provider implementation:  https://github.com/oceanprotocol/provider/blob/main/ocean_provider/routes/consume.py#L95

jamiehewitt15:
Fixes #105 

Changes proposed in this PR:

- Updating storage classes to include the `getFileInfo` method
- Createing core `FileInfoHandler` handler for 
- Creating http api endpoint for requesting file info  `/api/fileinfo`
- Updating storage and complete flow tests to include tests for the file info.


bogdanfazakas (approved):
lgtm

src/components/core/fileInfoHandler.ts
```diff
@@ -0,0 +1,144 @@
+import { Readable } from 'stream'
+import urlJoin from 'url-join'
+import { P2PCommandResponse } from '../../@types'
+import {
+  ArweaveFileObject,
+  IpfsFileObject,
+  UrlFileObject
+} from '../../@types/fileObject.js'
+import { FileInfoCommand } from '../../utils'
+import { P2P_CONSOLE_LOGGER } from '../../utils/logging/common.js'
```
paulo-ocean, line 10: can we change the Logger and use the CORE_LOGGER instead?,  all core components should use the core logger, not the P2P
thkx

src/utils/constants.ts
```diff
@@ -55,6 +58,14 @@ export interface DownloadCommand extends Command {
   aes_encrypted_key?: string // if not present it means download without encryption
 }
 
+export interface FileInfoCommand extends Command {
```
paulo-ocean, line 61: All fields are optional? a bit strange IMO, but don't know the details on this

src/utils/util.ts
```diff
@@ -7,6 +7,20 @@ export function sleep(ms: number) {
   return new Promise((resolve) => setTimeout(resolve, ms))
 }
 
+export async function streamToObject(stream: Readable): Promise<any> {
+  const chunks: Buffer[] = []
```
paulo-ocean, line 11: I think we could reuse `streamToString` here.
this entire piece:
```
const chunks: Buffer[] = []
  for await (const chunk of stream) {
    chunks.push(Buffer.from(chunk))
  }
  const buffer = Buffer.concat(chunks)
  const jsonString = buffer.toString()
```
is basically what `streamToString` does.
We could just call that function and then do the `JSON.parse(string)`
Not a big deal but i think we should avoid duplicated code when possible
tkx

src/test/unit/storage.test.ts
```diff
@@ -222,3 +223,116 @@ describe('Arweave Storage tests', () => {
...
+  })
+
+  beforeEach(() => {
+    storage = new IpfsStorage({
+      type: 'ipfs',
+      hash: 'QmRhsp7eghZtW4PktPC2wAHdKoy2LiF1n6UXMKmAhqQJUA'
+    })
+  })
+
+  it('Successfully retrieves file info for an IPFS hash', async () => {
```
paulo-ocean, line 315: hi @jamiehewitt15 
this tests always fails for me locally. The only way it works is by moving the content of `beforeEach()`
to inside `before() `function, i mean this part:
```
storage = new IpfsStorage({
      type: 'ipfs',
      hash: 'QmRhsp7eghZtW4PktPC2wAHdKoy2LiF1n6UXMKmAhqQJUA'
 })
```
By the way, is there any specific need for using the` beforeEach()` on these tests? Its  basically doing the same thing before every single test (and there is no corresponding `afterEach()` anyway) , imo, this could easily be done just once on the corresponding `before()`

src/utils/constants.ts
jamiehewitt15: Yeah, nothing is compulsary across all the scenarios depending if they are making the request with a did, url, ipfs has or arweave trnasactionId.  This is the validation we had in the provider: 

```python
class FileInfoRequest(CustomJsonRequest):
    def rules(self):
        return {
            "type": [
                "required_without:did",
                f"in:{','.join(FilesTypeFactory.ALLOWED_FILE_TYPES)}",
            ],
            "did": ["required_without:type", "regex:^did:op"],
            "hash": ["required_if:type,ipfs"],
            "url": ["required_if:type,url"],
            "transactionId": ["required_if:type,arweave"],
            "serviceId": ["required_without:type"],
        }
```

src/utils/util.ts
jamiehewitt15: Yeah good point, thanks for the feedback

src/test/unit/storage.test.ts
jamiehewitt15: Weird that this caused the tests to fail for you, but you make a good point - the `beforeEach` is unnecessary so I've moved it to `before`

paulo-ocean (approved):
lgtm

PR #218, opened 2024-01-15
refactor logger, transports based on environment

### LINKED ISSUE #196: All Loggers should log to the same transports
Right now when we create a new Logger instance, we can decide which transport to use.
As a result, some loggers are just logging to the console, others to the console and DB, etc..
It should not be up to the developer to choose where to log..
Logger transports should be configurable, as they are already, but it needs some extra refactoring so we don't have to choose the transport every time we create a new instance (so a little less freedom than current setup).
For instance;
We can have different log levels logging to different transports (ex: Info goes only to console, Error to db and console, etc)
Or we can also have some transports for development, others for production

paulo-ocean:
Fixes #196 

Changes proposed in this PR:

- Make use of `NODE_ENV` property to determine where (which transports) we are logging to 

- Remove custom transports param from `getCustomLoggerForModule`. Refactored custom loggers.

- by default, all loggers log to a file with the module/component name. If NODE_ENV is 'development' we also log to console. And if 'production' (or different from 'development' ) and having a Database, the DB transport is also added. First to DB logger, than in a "lazy" way to all other registered loggers (mostly due to dependency cycle problems).

- refactored Logger instances. Strictly deny creating more than 10 different instances. There are 8 different loggers instances on common.js and should be sufficient to cover all the different components of OceanNode. (ocean node, http, core, db, config, indexer, p2p, provider)

(btw, the only reason why there is not just a single one, is because its helpful and easy to prepend the component name to the logger message, and we can easily separate messages like this... otherwise we would have to pass it always to the message string and logging would be more confusing)


an7e (approved):

src/index.ts
```diff
@@ -11,7 +11,8 @@ import { getConfig } from './utils/index.js'
 import {
   GENERIC_EMOJIS,
   LOG_LEVELS_STR,
-  newCustomDBTransport
+  configureCustomDBTransport,
```
jamiehewitt15, line 14: Can you remove this import of `configureCustomDBTransport`

paulo-ocean: done

jamiehewitt15 (approved):
Looks good to me

bogdanfazakas (approved):

PR #221, opened 2024-01-17
Feature/ Add handle decrypt method

### LINKED ISSUE #108: Handle Encrypted DDOs
- if encrypted, reach out to provider/node that was used to encrypt and fetch decrypted ddo
 
Reference:  https://github.com/oceanprotocol/aquarius/blob/main/aquarius/events/decryptor.py

bogdanfazakas:
Fixes #108   .

Changes proposed in this PR:

- handle all possible encryption cases when processing a new DDO  
- added double checs for signature in ddo & download handler
- fixes teardown issues when tearing down the process env vars set in test files


src/test/utils/utils.ts
```diff
@@ -113,8 +113,12 @@ export async function tearDownEnvironment(overrideVars?: OverrideEnvConfig[]) {
     overrideVars.forEach((element: OverrideEnvConfig) => {
       if (element.override && element.newValue !== element.originalValue) {
         // only restore what we have explicilty touched
-        CONFIG_LOGGER.debug('Restoring environment variable: ' + element.name)
-        process.env[element.name] = element.originalValue
+        CONFIG_LOGGER.debug('Restoring environment variable: ' + element.originalValue)
+        if (element.originalValue) {
```
paulo-ocean, line 117: can you explain me why this change?
on line 114 we have:
`if (element.override && element.newValue !== element.originalValue) {`
So, if we explicitly overrided this variable, and the new value (most recent one) is different from the original one, it is restored, exactly as it was..
You're deleting a process.env key, if it was not set to anything OR...
what happens now if the element.originalValue was "false" ? :-)
if this:
```
if (element.originalValue) {
...
``` evaluates to false?

you're deleting it, instead of restoring to the original value as it was

Comment (bogdanfazakas) [reply src/test/utils/utils.ts:117]: so the use case was the following for eg:
the env AUTHORIZED_DECRYPTERS is undefined 
we use **setupEnvironment** to make it the following` ['0xe2DD09d719Da89e5a3D0F2549c7E24566e947260']`
at the end we want to make it undefined again
with the code above we will end up with `process.env.AUTHORIZED_DECRYPTERS = 'undefined'`
and this check https://github.com/oceanprotocol/ocean-node/blob/develop/src/utils/config.ts#L72 will fail later on when getting the new config with force reload because of unexpected values at index .. and `!process.env.AUTHORIZED_DECRYPTERS`  is also `false`

so we need to properly delete an env var that we set with **setupEnvironment** method when we are doing the **tearDownEnvironment** after we ryb all the tests

Comment (paulo-ocean) [reply src/test/utils/utils.ts:117]: ok, i see your point, but to me, that check [https://github.com/oceanprotocol/ocean-node/blob/develop/src/utils/config.ts#L72](https://github.com/oceanprotocol/ocean-node/blob/develop/src/utils/config.ts#L72) is worng. It should parse an array of **valid** eth addresses, anything except that should be consider empty array [] :-)

Comment (paulo-ocean) [reply src/test/utils/utils.ts:117]: Oops, @bogdanfazakas , i accidentally did a commit on this branch, instead of mine (fix a storage test)  sorry :-)

Comment (bogdanfazakas) [reply src/test/utils/utils.ts:117]: no worries :)
i was wondering what are those changes 😄

Comment (paulo-ocean) [reply src/test/utils/utils.ts:117]: just reverted it :-)

Comment (paulo-ocean) [reply src/test/utils/utils.ts:117]: by the way, just fixed `getAuthorizedDecrypters()` on my branch, it was just wrong, 
There was too much code and still not resilient to a wrong value set
here's the change: 
[https://github.com/oceanprotocol/ocean-node/pull/248/commits/e18ead8da409a8801bcfd0d55c552c0500695cb5](https://github.com/oceanprotocol/ocean-node/pull/248/commits/e18ead8da409a8801bcfd0d55c552c0500695cb5)

File: src/components/Indexer/processor.ts [hunk 2]
```diff
@@ -65,6 +71,105 @@ class BaseEventProcessor {
...
+    txId: string,
+    metadataHash: string,
+    metadata: any
+  ): Promise<any> {
+    let ddo
+    if (flag === '0x02') {
+      INDEXER_LOGGER.logMessage(
+        `Decrypting DDO  from network: ${this.networkId} created by: ${eventCreator} ecnrypted by: ${decryptorURL}`
+      )
+      const nonce = Number(Date.now().toString())
```
paulo-ocean, line 90: If we only need the nonce as string, why are we converting to Number() if all the usages of it above require string? :-)

src/test/integration/encryptDecryptDDO.test.ts
```diff
@@ -64,6 +64,7 @@ describe('Should encrypt and decrypt DDO', () => {
     provider = new JsonRpcProvider('http://127.0.0.1:8545')
     publisherAccount = (await provider.getSigner(0)) as Signer
     publisherAddress = await publisherAccount.getAddress()
+    console.log('publisherAddress ', publisherAddress)
```
paulo-ocean, line 67: IMO, better clean these console.log :-)

src/test/utils/utils.ts
bogdanfazakas: yep think deleting an env var that was set by a unit/integration test for real is better than having it hanging with 'undefined' or smth else

src/test/integration/encryptDecryptDDO.test.ts
bogdanfazakas: busted :), will clean those

src/components/Indexer/processor.ts
bogdanfazakas: tbh was one of the many thins i tried to just match the signature we had in ddo handler, but will revert it, looked redundant for me as well but, just tried all i saw different 😄

paulo-ocean: don't forget the `nonce: nonce.toString()` calls... its already a string now :-)

bogdanfazakas: thanks for reminding of those, done :)

paulo-ocean (approved):
lgtm

jamiehewitt15 (approved):
Looks good to me

PR #223, opened 2024-01-18
unit test to double check if all supported commands have handlers

### LINKED ISSUE #222: Simple unit test to make sure we have 1 handler for any supported command
Just to make sure we don't forget to register the specific handler for a command on the registry, or if we
accidentally delete an existing one (happened already)..
Just a simple test that loops through all supported commands and try to get the corresponding handler

paulo-ocean:
Fixes #222 .

Changes proposed in this PR:

- 1 Unit Test to double check that all supported commands have 1 registered handler (to make sure we don't forget any) 
- If this test fails we know that either we do not registered the new handler or we did not updated `SUPPORTED_PROTOCOL_COMMANDS` correctly

### DIFF
--- a/src/test/unit/commands.test.ts
+++ b/src/test/unit/commands.test.ts
@@ -1,0 +1,40 @@
+import { expect } from 'chai'
+import {
+  ENVIRONMENT_VARIABLES,
+  SUPPORTED_PROTOCOL_COMMANDS,
+  getConfig
+} from '../../utils/index.js'
+import {
+  buildEnvOverrideConfig,
+  setupEnvironment,
+  tearDownEnvironment
+} from '../utils/utils.js'
+import { OceanNodeConfig } from '../../@types/OceanNode.js'
+import { OceanP2P } from '../../components/P2P/index.js'
+import { CoreHandlersRegistry } from '../../components/core/coreHandlersRegistry.js'
+import { Handler } from '../../components/core/handler.js'
+
+describe('Commands and handlers', async () => {
+  const envOverrides = buildEnvOverrideConfig(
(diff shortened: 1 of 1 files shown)


jamiehewitt15 (approved):
Good idea, a useful extra test. Looks good to me.

src/test/unit/commands.test.ts
```diff
@@ -0,0 +1,40 @@
...
+} from '../utils/utils.js'
+import { OceanNodeConfig } from '../../@types/OceanNode.js'
+import { OceanP2P } from '../../components/P2P/index.js'
+import { CoreHandlersRegistry } from '../../components/core/coreHandlersRegistry.js'
+import { Handler } from '../../components/core/handler.js'
+
+describe('Commands and handlers', async () => {
+  const envOverrides = buildEnvOverrideConfig(
+    [ENVIRONMENT_VARIABLES.PRIVATE_KEY],
+    ['0xc594c6e5def4bab63ac29eed19a134c130388f74f019bc74b8f4389df2837a58']
```
bogdanfazakas, line 20: not related to the PR itself, but think we should stop using this key in tests and use just the one coming from the CI and locally just relay on the env vars that everyoone has configured :)

bogdanfazakas (approved):

src/test/unit/commands.test.ts
paulo-ocean: when running locally there is no CI, but the idea is that you don't even need to configure env.vars (i could had just pass the .env.test file there), i think i can take it back easily anyway

paulo-ocean: just cleaned the test

PR #224, opened 2024-01-18
Create OceanNode singleton class - Simplify access to properties

### LINKED ISSUE #209: Think about having a Singleton for OceanNode class (that will include also all handlers)
Right now most of our handlers need access to DB or OceanP2P.
The way we have the handling of commands implemented we need to pass the OceanP2P to the constructor..
But in most cases we do not need any P2P networking at all (only for FindDOO which queries alternative providers for did)

And often we have dependency cycle issues when "cross" importing stuff, especially when running tests.
Because some of these functions for instance, reference OceanP2P class before it gets initialized.. Basically we are being forced to code in a certain way, because of the test suite... When running a node instance, we have a single entry point (index.js) and we do not have any issues... So, its not the node code, in most cases, but the references to certain classes when import files that contain functions that reference classes (uninitialized classes).

Some refactor have been done, but it is still somewhow conditioned by this..

A proposed approach could be having just a Singleton OceanNode class instance...
And whenever we need P2P, DB, Indexer, Provider, just get those properties from the OceanNode instead

Right now, if we do not have a P2P node, our handlers do nothing, cause most need to access DB and they are doing it through the P2P node instead.. On the other hand, we might not have a local DB but still should be able to do P2P networking to forward the requests.. AS IS, everything seems to work but in reality its all a bit "hacked".. 

Something like:

```
export class OceanNode {

private OceanNode instance

private constructor(
    private config: OceanNodeConfig,
    private db: Database,
    private node?: OceanP2P,
    private indexer?: OceanIndexer
  )
  
  public static getInstance(config: OceanNodeConfig,
    db: Database,
    node?: OceanP2P,
    provider?: OceanProvider,
    indexer?: OceanIndexer): OceanNode {
    // get instance or build new new OCeanNode()
     return instance
  }
}
```

so, we could always refer to OceanNode.getInstance() ... to get access to other elements....
for handling commands, etc...
we have a lot of code like: `this.getP2PNode().getDatabase()` on the handlers (that don't even need P2P), and also some boilerplate code on Indexer and Provider to access Database... All this could come from the OceanNode instance

PS: Also Commands/interfaces should move out of utils/constants.ts to the types folder

paulo-ocean:
Fixes #209  #235

Changes proposed in this PR: 
PS: there are many changed files but not so many changes after all , don't be scared :-) 

- Singleton class for OceanNode

- Refactor handlers. Avoid usage of P2P node (that might not even exist... although this option will be addressed on a separate issue), to access DB for instance or node configuration

- Clean some boilerplate code related with difficulty to access DB or config ( provider, mostly indexer, etc ). Utility functions (lazy loading) for config and database... This avoids  dependency cycles issues

- Unified way to access config (dropped config as part of components, just use exported core function `getConfiguration()` (it uses `getEnvConfig()` previously named `getConfig()` and not exported anymore ) for global node config). As multiple "configs"  objects spread around, might have different settings and break logic. So, there is only 1 way to get config now.  Use boolean param to specify if we need a full reload (useful on tests for instance, where we change .env properties "on the fly")

- Log/Warn every time a .env config change is detected (when we change environment variables)

- Inside Ocean Node stack we can just use `utils/config/getConfiguration()` . It also accepts a bool param if we want to force a reload from .env vars.... For DB access use Ocean Node Singleton property of lazy function `utils/database.ts/getDatabase()`. Its accessing same DB anyway...

- For tests, we could use  setup/teardown functions if we need to change .env properties (AVOID at all cost)... in any case these functions preserve environment between tests, but also handle any config changes and eventually any needed "config reloads" automatically


src/components/core/utils/nonceHandler.ts
```diff
@@ -35,11 +35,10 @@ export type NonceResponse = {
 
 // get stored nonce for an address ( 0 if not found)
 export async function getNonce(
-  node: OceanP2P,
```
jamiehewitt15, line 38: Let's remove the `OceanP2P` import above

src/test/integration/encryptDecryptDDO.test.ts
```diff
@@ -65,16 +73,33 @@ describe('Should encrypt and decrypt DDO', () => {
...
+          JSON.stringify(mockSupportedNetworks),
+          JSON.stringify([publisherAddress])
+        ]
+      )
+    )
+
+    // WE SHOULD STOP DOING THIS!!! It breaks local env and other tests!
+    // process.env.PRIVATE_KEY =
+    // '0xc594c6e5def4bab63ac29eed19a134c130388f74f019bc74b8f4389df2837a58'
+    // use setupEnvironment / tearDownEnvironment instead (But only if its really needed to override some stuff)
```
jamiehewitt15, line 96: Might be worth putting this as a general note in the readme, as it applies to all the tests and most people won't see it here.

src/components/P2P/index.ts
```diff
@@ -551,7 +551,7 @@ export class OceanP2P extends EventEmitter {
         `Trying to store and advertise ${list.length} initial DDOS`,
         true
       )
-      const db = this.getDatabase().ddo
+      const db = this.db.ddo
```
bogdanfazakas, line 554: what if the instance of p2p was created without specifying the db param since it is optional ? this.db will be undefined and this will lead to the clasic cannout read property of undefined 😃

src/test/integration/encryptDecryptDDO.test.ts
```diff
@@ -65,16 +73,33 @@ describe('Should encrypt and decrypt DDO', () => {
...
-    process.env.PRIVATE_KEY =
-      '0xc594c6e5def4bab63ac29eed19a134c130388f74f019bc74b8f4389df2837a58'
-    process.env.RPCS = JSON.stringify(mockSupportedNetworks)
-    process.env.AUTHORIZED_DECRYPTERS = JSON.stringify([publisherAddress])
+
+    previousConfiguration = await setupEnvironment(
+      null,
+      buildEnvOverrideConfig(
+        [
+          ENVIRONMENT_VARIABLES.PRIVATE_KEY,
```
bogdanfazakas, line 81: i think we can optimize more this, for eg:

-  we have a private key env var inside the cli, and locally everyone should set the env vars using own wallet addresses
- we should also update RPCS env var in the cli to work with barge and maybe also in the readme we should have barge and maybe the optimism we used for our demos
- also AUTHORIZED_DECRYPTERS should be added to cli and README and if certain user running tests with different config should be able to just set the propper env var locally

src/test/integration/encryptDecryptDDO.test.ts
paulo-ocean: done, added some notes on readme

src/components/P2P/index.ts
paulo-ocean: That is only called on `index.ts` :

```
if (config.hasIndexer && dbconn) {
  indexer = new OceanIndexer(dbconn, config.supportedNetworks)
  // if we set this var
  // it also loads initial data (useful for testing, or we might actually want to have a bootstrap list)
  // store and advertise DDOs
  if (process.env.LOAD_INITIAL_DDOS) {
    const list = loadInitialDDOS()
    if (list.length > 0) {
      // we need a timeout here, otherwise we have no peers available
      setTimeout(() => {
        node.storeAndAdvertiseDDOS(list)
      }, 3000)
    }
  }
}
```
there are a couple of checks there as you can see :-) , and this is something just for testing & loading initial DDOS.. so that is not a problem at all
but i can make an extra "if" for sure... just to avoid misuse

jamiehewitt15: Maybe just add `?` as an extra check to avoid anything going wrong 

```suggestion
      const db = this.db?.ddo
```

src/test/integration/encryptDecryptDDO.test.ts
paulo-ocean: I agree with you that we should not override .env vars.. I'm saying that all the time as well, and that's why i'm cleaning some errors done with the tests :-)

but we don't need to setup anything on our own.. that is already arranged with the root hooks and everything is already optimized.. 

but everyone should stop doing this on the tests... i already updated the README with some more info.
Apart the CLI there is also a .env.test, but we should stop creating tests with specific .env vars , that is the only reason for this code above, to fix that
tkx

src/components/P2P/index.ts
paulo-ocean: just added the extra if, no big deal..

bogdanfazakas (approved):

src/test/integration/encryptDecryptDDO.test.ts
paulo-ocean: By the way, just for context, this is what we are doing (and why i made the change):

```
 process.env.PRIVATE_KEY =
      '0xc594c6e5def4bab63ac29eed19a134c130388f74f019bc74b8f4389df2837a58'
 process.env.RPCS = JSON.stringify(mockSupportedNetworks)
 process.env.AUTHORIZED_DECRYPTERS = JSON.stringify([publisherAddress])
```
This is definetly wrong

jamiehewitt15 (approved):
Looks good to me

Issue #205, opened 2024-01-08
DDO Schema handling


alexcos20:
Once we have validateDDO command:
- when adding new ddo (Metadata created), identify the right collection and store the ddo  (indexer)
- when query, make sure all collections are used

jamiehewitt15:
We also need to ensure that a did does not exist in more than one collection, so if a ddo is updated to use a new schema it should be removed from all other schemas.

alexcos20:
so the flow is the following:
- metaDataCreated/Update 
    - check version. If we don't have a version in the DDO, discard it (we cannot validate it)
    - if valid, put it in corresponding collection, otherwise log error and discard it
    - when updated, and new_version != current_version, delete it from old collection and put updated version in new collection
- metaDataState:
    - if changes from active to disabled:  delete it from old collection and put short version in "short_collection" collection
    - if changes from disabled to active:  delete it from "short_collection" and put it in corresponding collection according to version


When doing query:
use all "version" collections + "short_collection"

jamiehewitt15:
Merged in: https://github.com/oceanprotocol/ocean-node/pull/239

PR #225, opened 2024-01-19
Create collections specific for ddo versions

### LINKED ISSUE #207: Create collections at startup for all known schema versions
on startup, grab the list of all known schemas and create collections in the db for each schema

mariacarmina:
Fixes #207  .

Changes proposed in this PR:

- ddoSchemas are created dinamically


mariacarmina:
Should I also modify `database/index.ts` for ddo database, where we basically retrieve the first schema, into choosing the schema by DDO version? Or this should be done in #205 ?
For the moment, we load the schemas mappings dinamically into typesense db. Let me know if it's needed in this PR to modify also the CRUD methods from `database/index.ts`. Thank you!

cc @alexcos20 @jamiehewitt15

jamiehewitt15:
> Or this should be done in https://github.com/oceanprotocol/ocean-node/issues/205

I see what you mean there is some overlap between the two issues. I think it's fine to do it in the next PR, which keeps this one short and easy to review

jamiehewitt15 (approved):
Looks good to me

src/components/database/schemas.ts
```diff
@@ -1,4 +1,37 @@
 import { TypesenseCollectionCreateSchema } from '../../@types/index.js'
+import fs from 'fs'
+import path, { dirname, resolve } from 'path'
+import { fileURLToPath } from 'url'
+import { DATABASE_LOGGER } from '../../utils/logging/common.js'
+import { LOG_LEVELS_STR } from '../../utils/logging/Logger.js'
+
+export function readJsonSchemas(): any[] {
+  const jsonDocuments: any[] = []
+  const pathToSchemaDir: string = '../../../schemas/v4'
```
alexcos20, line 10: can we have just schema folder  (instead of schemas/v4 ? )

alexcos20 (changes requested):

src/components/database/schemas.ts
mariacarmina: done

src/components/database/schemas.ts
```diff
@@ -1,4 +1,37 @@
 import { TypesenseCollectionCreateSchema } from '../../@types/index.js'
+import fs from 'fs'
+import path, { dirname, resolve } from 'path'
+import { fileURLToPath } from 'url'
+import { DATABASE_LOGGER } from '../../utils/logging/common.js'
+import { LOG_LEVELS_STR } from '../../utils/logging/Logger.js'
+
+export function readJsonSchemas(): any[] {
+  const jsonDocuments: any[] = []
```
bogdanfazakas, line 9: i think better to create a new type in src/@types with .json files fields and use that instead of any

src/components/database/schemas.ts
```diff
@@ -1,4 +1,37 @@
...
+  const pathToSchemaDir: string = '../../../schemas'
+  const currentModulePath = fileURLToPath(import.meta.url)
+
+  // Use dirname to get the directory name
+  const currentDirectory = dirname(currentModulePath)
+  const schemaFilePath = resolve(currentDirectory, pathToSchemaDir)
+  const jsonFiles = fs
```
bogdanfazakas, line 16: better to ensure also this one does not fail since not surounded with try catch block, at leas check if jsonFiles exists to avoind .forEach of undefined bellow

src/components/database/index.ts
```diff
@@ -614,11 +614,11 @@ export class Database {
         configureCustomDBTransport(this, DATABASE_LOGGER)
       }
 
-      this.ddo = await new DdoDatabase(config, schemas.ddoSchemas)
-      this.nonce = await new NonceDatabase(config, schemas.nonceSchemas)
-      this.indexer = await new IndexerDatabase(config, schemas.indexerSchemas)
-      this.logs = await new LogDatabase(config, schemas.logSchemas)
-      this.order = await new OrderDatabase(config, schemas.orderSchema)
+      this.ddo = await new DdoDatabase(this.config, schemas.ddoSchemas)
```
paulo-ocean, line 617: unnecessary change here, no need for the "this" keyword, not an issue but certainly unnecessary to change

mariacarmina: would you prefer to remove it?

paulo-ocean: any param marked as `private` in the constructor is automatically assigned (we don't need to create a matching property and assign on the class, typescript does that automatically)... its greyed out by the linter but we know better :-)
Anyway, i do not care at all, one way or the other is fine :+1:  .. just pointed it out because there was no need for that change

src/components/database/schemas.ts
```diff
@@ -16,6 +16,15 @@ export function readJsonSchemas(): any[] {
   const jsonFiles = fs
     .readdirSync(schemaFilePath)
     .filter((file) => path.extname(file) === '.json')
+
+  if (!jsonFiles || jsonFiles === undefined) {
```
paulo-ocean, line 20: this is still not exactly fine @mariacarmina 
this piece:
```
const jsonFiles = fs
    .readdirSync(schemaFilePath)
    .filter((file) => path.extname(file) === '.json')
```
is not part of the try catch as @bogdanfazakas  mentioned. Besides that, the filter() will return an array [], which could be empty []

With an empty array, this part:

`if (!jsonFiles || jsonFiles === undefined) {`

Is Not executed

And it still will do the forEach() part without logging the error message :-(

src/components/database/index.ts
mariacarmina: > its greyed out by the linter but we know better :-). It was being used
Yes, ofc it was used, otherwise it wasn't working properly. Thanks for pointing out, I have a habit to put `this` wherever needed within a class, good practice.

src/components/database/schemas.ts
mariacarmina: Ok then I will include in the try/catch block as well. Thank you

paulo-ocean (commented):
@mariacarmina please take a couple of minutes before rushing the commits :-)
If an exception is caught on the outer try/catch (have no idea why we need both but anyway)
The function returns **nothing**, it just logs a message, it should return empty array as well
tkx

mariacarmina:
> If an exception is caught on the outer try/catch (have no idea why we need both but anyway) The function returns **nothing**, it just logs a message, it should return empty array as well to be consistent tkx

I return now empty array in the outer try/catch block.

paulo-ocean (approved):
lgtm

paulo-ocean (approved):
lgtm

bogdanfazakas (approved):
lgtm

mariacarmina:
Thanks for approvals, but @alexcos20 needs to undo the requested changes status in order to merge it 😅

alexcos20 (approved):

PR #234, opened 2024-01-22
Expose validateDDO on http 

### LINKED ISSUE #47: Command:  validateDDO
We need to validate DDOs before that are published:

TO DO:

- [x]  implement core function validateDDO, according to available schacl schemas
- [ ]   expose it on http interface
- [x]   expose it on p2p interface

Command description:

{
command: "validateDDO",
ddo: Object,
}

Returns:
 - signed proof of validation
 ```
{
    "hash": "0x"  // ddo hash
    "nodeID": nodeID that signed the proof
    "v" :  signature.v,
    "r" :  signature.r,
    "s" :  signature.s
}

mariacarmina:
Fixes #47  .

Changes proposed in this PR:

- added DDO validation endpoint for http interface


jamiehewitt15 (approved):
Looks good to me

src/components/httpRoutes/aquarius.ts
```diff
@@ -131,3 +131,40 @@ aquariusRoutes.get('/state/ddo', async (req, res) => {
...
+      return
+    }
+    const ddo = JSON.parse(req.body)
+
+    if (!ddo.version) {
+      res.status(400).send('Missing DDO version')
+      return
+    }
+
+    if (!ddo.version) {
```
paulo-ocean, line 148: Duplicate line here

src/components/httpRoutes/validateCommands.ts
```diff
@@ -36,8 +36,9 @@ export function validateCommandAPIParameters(requestBody: any): ValidateParams {
           status: 400
         }
       }
-      // nonce
-    } else if (command === PROTOCOL_COMMANDS.NONCE) {
+    }
+    // nonce
+    else if (command === PROTOCOL_COMMANDS.NONCE) {
```
paulo-ocean, line 41: We are not adding any params validation for the directCommand (only on aquarius route) ..
Are we going to do that on [https://github.com/oceanprotocol/ocean-node/issues/47](https://github.com/oceanprotocol/ocean-node/issues/47) ?
thanks

mariacarmina: > We are not adding any params validation for the directCommand (only on aquarius route) ..
> Are we going to do that on https://github.com/oceanprotocol/ocean-node/issues/47 ?

From sprint planning discussion, @alexcos20 mentioned only the separate endpoint option in `aquarius.ts`. I have added before in `directCommand`. If I add the validation also here, that means that validateDDO command will be included within /directCommand route.

@alexcos20 can you confirm that `validateDDO` will be integrated in just a separate endpoint as it is in aquarius or can we have it also inside /directCommand route?

mariacarmina: > Also, i noticed that we are not logging anything (errors) on ValidateDDOHandler (also happens on some other handlers) , if something fails we have no idea of what happened.. Only the client that initiated the request will get some info on the status response.

Regarding this, besides `extraErrors` which are for DDO object validation, the logs handle internal server error (where basically print the error found with status 500), failure in detecting and reading shacl schemas, unexisting validation report. If there are not enough, I think there is already an issue for adding logs or something similar if I am not wrong. But in the meantime, I have a retrieve from DB and a call to `validateObject` which can bring error. What other logs can be suitable for these 2 functions if they are already have a specific log for each of them?

paulo-ocean: i don´t know, but this PR says it says it fixes #47 , and it includes http and p2p, that's why i'm asking..
from the description:
Command description:

```
{
command: "validateDDO",
ddo: Object,
}
```
this is clearly the directCommand, but i know if we are addressing this issue into separate tasks or not

mariacarmina: > this is clearly the directCommand, but i know if we are addressing this issue into separate tasks or not
I have added in [my first commit](https://github.com/oceanprotocol/ocean-node/pull/234/commits/bc47a8bad5f6cb89453760445d4d23e26f5bfa25#diff-9097c73b57fa2e60596f49b0309fdc0505e2cfb189f2d28f06782f716c353eefR39) from this PR :)
I can re-add here, no need to create a new issue.

src/components/httpRoutes/validateCommands.ts
```diff
@@ -36,6 +36,21 @@ export function validateCommandAPIParameters(requestBody: any): ValidateParams {
           status: 400
         }
       }
+    } else if (command === PROTOCOL_COMMANDS.VALIDATE_DDO) {
```
paulo-ocean, line 31: @mariacarmina , please check my previous comments. The `VALIDATE_DDO` has exactly the same required parameters of `FIND_DDO` and `GET_DDO` (the "id"), there is no need for another "if", its exactly the same code. You can aggregate the 3 in one. Also use `buildInvalidRequestMessage` lets not repeat again the same repetitive blocks that were removed before:
```
return {
          valid: false,
          reason: 'Missing or invalid required parameter: "id"',
          status: 400
        }
```
thanks

paulo-ocean (changes requested):

paulo-ocean (approved):
lgtm

PR #236, opened 2024-01-22
add api doc

### LINKED ISSUE #51: Add docs structure
We need two doc files, to detail:

- [x] Commands  (both HTTP API and P2p API)
- [x] Env vars

an7e:
Fixes #51

### DIFF
--- a/API.md
+++ b/API.md
@@ -2,9 +2,524 @@
 
 ---
 
+## State DDO
+
+### `HTTP` GET /api/aquarius/assets/metadata/query?
+
+#### Description
+
+returns ddo state
+
+#### Query Parameters
+
+| name    | type   | required | description                                            |
+|---------|--------|----------|--------------------------------------------------------|
+| did     | object |         | document id or did                                     |
+| chainId | object |         | chain id of network on which document is provided |
+| nft     | object |         | one or more field names that should be queried against |
+
+#### Response
+
+```
+123
+```
+
+---
+
+## Query DDO
+
+### `HTTP` POST /api/aquarius/assets/metadata/query
+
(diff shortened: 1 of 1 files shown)


jamiehewitt15:
Looks good on first impressions. Although, I am wondering - couldn't we have used an automated tool like [TypeDoc](https://www.npmjs.com/package/typedoc) for this? That way it would also be super easy to keep it up to date when any changes are made to the code...

an7e:
> Looks good on first impressions. Although, I am wondering - couldn't we have used an automated tool like [TypeDoc](https://www.npmjs.com/package/typedoc) for this? That way it would also be super easy to keep it up to date when any changes are made to the code.
> 
> Without automation, files like this often get out of date over time as it's easier to forget to update it. We use TypeDoc in Ocean.js and it works pretty well there.

@jamiehewitt15 here was set the task by @alexcos20  to create API.md

from my experience - I always use manual writing of docs, even if it is a swagger, for several reasons

- trying to generate documentation from TypeScript types has never worked well, often libraries contain any types or complex variant types with complex structure that cannot be parse properly, especially generic types
- things like OpenApi and others don't work with socket protocols, and if you use your own byte protocol, it's better to forget about it

paulo-ocean (approved):
LGTM. :+1:  Only think that would make more more happy is if we actually start putting comments on our code. I don't enjoy to read documentation that much  :-)

bogdanfazakas (approved):
LGTM, tho same like the other guys I think in the future we need to a more automatic way of generating the docs swagger, openApi, typedoc or at least have some comments with jsDoc on the code itself.

PR #239, opened 2024-01-24
Issue 205 ddo handling

### LINKED ISSUE #205: DDO Schema handling

jamiehewitt15:
Fixes #205 

Changes proposed in this PR:

- Ensuring that the right collection is used at all times, based on the version number in the ddo. 
- Searching all schemas when retrieving, updating or deleting DDOs.
- Ensuring that we don't have the DDO in multiple collections at the same time.
- Creating new schema for short DDOs which are used for deprecated assets.


src/components/database/index.ts
```diff
@@ -255,17 +299,36 @@ export class DdoDatabase {
...
-    } catch (error) {
-      const errorMsg = `Error when deleting DDO entry ${did}: ` + error.message
+    let isDeleted = false
+    for (const schema of this.schemas) {
+      try {
+        const response = await this.provider
+          .collections(schema.name)
+          .documents()
+          .delete(did)
+        isDeleted = true
```
paulo-ocean, line 309: hi @jamiehewitt15 
Have some doubts here..
are we sure that the document was deleted? i could be wrong but i think you have to analyse the response (not sure if it throws an error)
i had a look here, but can't tell for sure..
[https://typesense.org/docs/0.25.2/api/documents.html#delete-a-single-document](https://typesense.org/docs/0.25.2/api/documents.html#delete-a-single-document)
And do we want to delete in just one collections or all the collection where it could be?.. 
Cause we're "assuming" it was deleted, and return at first pass of the for loop:
```
isDeleted = true
return response
```
Again, i could be wrong here, just my understanding of it
thanks

jamiehewitt15: Yeah, good point. I'll take another look at it

jamiehewitt15: I added a check on the response to make sure that DDO was actually deleted.

> And do we want to delete in just one collection or all the collection where it could be?
It should only ever be in one collection at a time... If we want to be 100% sure on that I could add a check within the create method, that first looks for the did in all collections before creating it.

paulo-ocean (approved):
lgtm

src/components/database/index.ts
```diff
@@ -212,34 +227,63 @@ export class DdoDatabase {
...
-      return null
     }
+
+    return ddo
   }
 
   async update(ddo: Record<string, any>) {
+    // Find the schema based on the DDO version OR use the short DDO schema when state !== 0
+    let schemaName: string
+    if (ddo.version) {
```
bogdanfazakas, line 267: think we can move most of this logic to a new helper method, already saw we already use it in two places.

jamiehewitt15: Yeah, good idea. I've added that

mariacarmina (commented):
I have a question, looking at the code, I could not find the validation result when saving the DDO for metadata events (in the processor.ts). I think this issue should also include this part as well, not only at the database level, but correct me if I am wrong.

jamiehewitt15:
> I have a question, looking at the code, I could not find the validation result when saving the DDO for metadata events (in the processor.ts). I think this issue should also include this part as well, not only at the database level, but correct me if I am wrong.

The focus of this issue is making sure that we're using the right schema and collection when entering or retrieving data from the database. When the metadata is saved into a collection from `src/components/Indexer/processor.ts` it uses the `update` method on the ddoDatabase class.  The `update` method has been adjusted within this PR to ensure that the right collection is used, based on the version number in the schema.

jamiehewitt15:
Currently blocked by this issue as the short DDO can't pass the validation: https://github.com/oceanprotocol/ocean-node/issues/256

Could potentially implement a quick fix to bypass that issue: 

1.  Avoid doing any validation on the short DDO until the validation function has been updated
2. Add the extra missing fields back into the short DDO

mariacarmina (approved):
lgtm

PR #240, opened 2024-01-24
add C2C cluster env config

### LINKED ISSUE #226: C2D: add C2D_CLUSTERS env for ocean-node
Compared to provider, which only supports one C2D Cluster,  ocean-node will handle multiple ones

So, enhance our config to take an additional C2D_CLUSTERS config env.
Our config should look like:
```json
{
...
"c2dClusters": [
{ 
  url: "http:///...",
  hash: "0x123'
}
]
```

where hash  = sha256(url)

 C2D_CLUSTER env:
If it's a string -> there is only one cluster
If it's an array -> we have several clusters

paulo-ocean:
Fixes #226 

Changes proposed in this PR:

-  Add new config `OPERATOR_SERVICE_URL` (array of 1 or multiple cluster URLS)
-  Updated README

(when we start actually using this and with a c2d env ready, we will have to update the CI env as well, for instance)

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,12 @@ export IPFS_GATEWAY='https://ipfs.io/'
 export ARWEAVE_GATEWAY='https://arweave.net/'
 ```
 
+For configuring a C2D (Compute to Data) cluster(s), please set the following environment variable (array of 1 or multiple cluster URLS):
+
+```bash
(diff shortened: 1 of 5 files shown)


alexcos20 (approved):

alexcos20:
see https://github.com/oceanprotocol/ocean-node/pull/241

jamiehewitt15 (approved):
Looks good to me

PR #246, opened 2024-01-25
fix unit test on commands.ts - pick mismatches in both directions

### LINKED ISSUE #245: Fix test/unit/commands.test.ts , it should pick missing handlers and mismatch between handlers/commands
There is a test to check if any of the supported commands is missing the corresponding handler.
However this test should also pick mismatches in the opposite direction (having handler with the corresponding command being added to the `SUPPORTED_PROTOCOL_COMMANDS` list

paulo-ocean:
Fixes #245 

Changes proposed in this PR:

- Make sure unit test pick both unregistered/missing handlers or new commands (with handler) but not added to SUPPORTED_PROTOCOL_COMMANDS

### DIFF
--- a/src/components/core/coreHandlersRegistry.ts
+++ b/src/components/core/coreHandlersRegistry.ts
@@ -53,7 +53,7 @@ export class CoreHandlersRegistry {
   // eslint-disable-next-line no-use-before-define
   private static instance: CoreHandlersRegistry
   // map of handlers registered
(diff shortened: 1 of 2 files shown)


jamiehewitt15 (approved):
Good idea, this can help catch some mistakes

bogdanfazakas (approved):
lgtm

PR #247, opened 2024-01-26
Create purgatory

### LINKED ISSUE #213: Indexer: purgatory
Given two ENV vars that control lists of assets and accounts, indexer should prevent ingesting assets which are a match

Format of the lists: https://github.com/oceanprotocol/list-purgatory

mariacarmina:
Fixes #213  .

Changes proposed in this PR:

- create purgatory class
- added two env vars ACCOUNT_PURGATORY_URL and ASSET_PURGATORY_URL and build the logic around them
- tests


.github/workflows/ci.yml
```diff
@@ -162,6 +162,8 @@ jobs:
           DB_URL: 'http://localhost:8108/?apiKey=xyz'
           FEE_TOKENS: '{ "1": "0x967da4048cD07aB37855c090aAF366e4ce1b9F48", "137": "0x282d8efCe846A88B159800bd4130ad77443Fa1A1", "80001": "0xd8992Ed72C445c35Cb4A2be468568Ed1079357c8", "56": "0xDCe07662CA8EbC241316a15B611c89711414Dd1a" }'
           FEE_AMOUNT: '{ "amount": 1, "unit": "MB" }'
+          ASSET_PURGATORY_URL: 'https://raw.githubusercontent.com/oceanprotocol/list-purgatory/main/list-assets.json'
```
paulo-ocean, line 165: Hi @mariacarmina 
do we really need 2 separate variables? Can´t we simplify this like FEE_AMOUNT, FEE_TOKENS, RPCS, etc? In a json object instead.

IMO, much simpler if we have less variables, specially if they work together

Are they mandatory/not mandatory?  These should be added to the utils/constants/ENVIRONMENT_VARIABLES so we keep track of what we have

Also, is there any special reason why these variables are not treated on "`getEnvConfig()`" like all the others?
thanks

src/utils/purgatory.ts
```diff
@@ -0,0 +1,6 @@
+import { Purgatory } from '../components/Indexer/purgatory.js'
+import { getDatabase } from './database.js'
+
+export async function getPurgatory(): Promise<Purgatory> {
```
paulo-ocean, line 4: Why export an utils function that is always building a new instance of Purgatory() ? We could do this from anywhere :-)
Besides, on the processor, every time "`processEvent()`" is called it is calling this and building a new instance
`const purgatory = await getPurgatory()` ,  without need for it... And with every new instance, you have new properties for `bannedAccounts` and `bannedAssets`. None of these values  persist between 2 different calls of "processEvent()"

src/test/integration/purgatory.test.ts
```diff
@@ -0,0 +1,70 @@
...
+  let database: Database
+  let purgatory: Purgatory
+
+  before(async () => {
+    const dbConfig = {
+      url: 'http://localhost:8108/?apiKey=xyz'
+    }
+    database = await new Database(dbConfig)
+    // these are the default URLs
+    process.env.ASSET_PURGATORY_URL =
```
paulo-ocean, line 15: We had already seen that this is not the best way to override .env vars..
It affects local env and other tests. Please always use setup/tearDown IF we really need to override
thanks

.github/workflows/ci.yml
mariacarmina: Hello, in aquarius we had 2 env and in the issue there was specified to use those 2 envs for purgatory logic (@alexcos20 can provide an answer regarding changing the logic or keep those 2 envs separately as it was in aquarius).

> Also, is there any special reason why these variables are not treated on "getEnvConfig()" like all the others?

Thanks for catching that, maybe that's the reason that the tests failed in the beggining. I set the env vars only in the CI and not specifying in the env config and needed to export again in the tests. I'll add them.

src/test/integration/purgatory.test.ts
mariacarmina: They are not overriding the existing ones, I'll remove them to see if the tests fail due to missing envs in the test scope. Thanks for suggestion!

src/utils/purgatory.ts
mariacarmina: I think we need to keep it inside event processor class like a parameter and have one call. How does it sound @paulo-ocean ?

src/test/integration/purgatory.test.ts
paulo-ocean: yes, they are, you're explicitly setting
process.env.ASSET_PURGATORY_URL 
process.env.ACCOUNT_PURGATORY_URL 
so, yes, its overriding the values, whatever they are :-)

mariacarmina: I set them here as well, but yeah they have the same values.

src/utils/purgatory.ts
paulo-ocean: even so, its still creating a new instance on every call. There are 2 different issues here

src/test/integration/purgatory.test.ts
paulo-ocean: you don't need to set them here, as they are on CI... and if they are mandatory and not optional (i don't know details on this)... the config() should warn that... and if they are mandatory and are not set, the node should not even start ...

src/components/Indexer/purgatory.ts
```diff
@@ -0,0 +1,151 @@
...
+  private bannedAssets: Array<PurgatoryAssets>
+
+  constructor(db: Database) {
+    this.db = db
+    this.bannedAccounts = []
+    this.bannedAssets = []
+  }
+
+  getDb(): Database {
+    return this.db
```
bogdanfazakas, line 19: do we still need this since we now have the getDatabase method from singleton OceanNode?

src/components/Indexer/purgatory.ts
```diff
@@ -0,0 +1,151 @@
...
+  setBannedAccounts(newBannedAccounts: Array<PurgatoryAccounts>): void {
+    this.bannedAccounts = newBannedAccounts
+  }
+
+  setBannedAssets(newBannedAssets: Array<PurgatoryAssets>): void {
+    this.bannedAssets = newBannedAssets
+  }
+
+  async parsePurgatoryAssets(): Promise<Array<PurgatoryAssets>> {
+    if (!('ASSET_PURGATORY_URL' in process.env)) {
```
bogdanfazakas, line 39: there is a method that does this in other places of the codebase `existsEnvironmentVariable` and would be better to stay consistent and use that here as well

src/test/integration/purgatory.test.ts
bogdanfazakas: agreed with Paolo here lets use setup and tear down only when dealing with env vars in our tests

src/components/Indexer/processor.ts
```diff
@@ -133,8 +134,22 @@ export class MetadataEventProcessor extends BaseEventProcessor {
...
+        (await purgatory.isBannedAsset(ddo.id)) ||
+        (await purgatory.isBannedAccount(from))
+      ) {
+        ddo.purgatory = {
+          state: true
+        }
+      } else {
+        ddo.purgatory = {
+          state: false
+        }
```
bogdanfazakas, line 149: this whole logic above can be one method that receives the did and the owner and return a boolean value that we set in ddo.purgatory.state, or maybe receives the `ddo` and the `owner` does the update inside the method and returned the already processed DDO.

src/components/Indexer/purgatory.ts
paulo-ocean: definitely, agree. The purgatory env vars (either 1 or 2) should be processed and part of the config, as the others.. we could create a sort of Purgatory type with "assets" and "accounts", and read/load that on the config... The we only need create 1 instance of Purgatory class (better use a singleton), when we know from the config that we have it set... no need for any checks inside the class.. If there is no purgatory .env vars.. an instance of the class should not even be created in the first place

mariacarmina: done, thanks.

src/utils/purgatory.ts
mariacarmina: Implemented the singleton now

src/components/Indexer/processor.ts
mariacarmina: done, thanks.

.mocharc.json
```diff
@@ -2,7 +2,7 @@
   "full-trace": true,
   "bail": true,
   "exit": true,
-  "timeout": 10000,
+  "timeout": 20000,
```
jamiehewitt15, line 5: I'm wondering why you changed the timeout, do the tests take a long time to run? If the request to the json file from github takes a long time, we would use a local json file in the test folder for the purgatory when running the tests.

mariacarmina: let me do a check if it's still valid.

src/components/Indexer/processor.ts
```diff
@@ -133,8 +134,14 @@ export class MetadataEventProcessor extends BaseEventProcessor {
           return
         }
       }
-      const saveDDO = this.createOrUpdateDDO(ddo, eventName)
-      return saveDDO
+      const from = decodedEventData.args[0]
+      const purgatory = Purgatory.getInstance(await getDatabase())
+      const updatedDDO = await this.updatePurgatoryStateDdo(ddo, from, purgatory)
+      if (updatedDDO.purgatory.state === false) {
+        // TODO: insert in a different collection for purgatory DDOs
```
jamiehewitt15, line 141: > // TODO: insert in a different collection for purgatory DDOs

If we're not implementing it in this PR should we create an issue to implement it further down the line?

.mocharc.json
jamiehewitt15: ok thanks

src/components/Indexer/processor.ts
mariacarmina: @alexcos20 said that he will add in the future more details/issues regarding purgatory functionality. Maybe he can provide some additional context here.

.mocharc.json
mariacarmina: it worked with initial value. thanks.

src/components/Indexer/processor.ts
jamiehewitt15: ok yeah, would be great if Alex can create issues for that additional functionality. If it's just a comment in the code, it's likely to be forgotten about

jamiehewitt15 (approved):
Looks good to me

src/utils/config.ts
```diff
@@ -348,7 +348,15 @@ async function getEnvConfig(isStartup?: boolean): Promise<OceanNodeConfig> {
     },
     supportedNetworks,
     feeStrategy: getOceanNodeFees(supportedNetworks, isStartup),
-    c2dClusters: getC2DClusterEnvironment()
+    c2dClusters: getC2DClusterEnvironment(),
+    accountPurgatoryUrl: getEnvValue(
+      process.env.ACCOUNT_PURGATORY_URL,
+      'https://raw.githubusercontent.com/oceanprotocol/list-purgatory/main/list-accounts.json'
```
alexcos20, line 354: Let's not have any defaults.  It will make it harder for people to start without purgatory, if they want to

src/components/Indexer/purgatory.ts
```diff
@@ -0,0 +1,162 @@
...
+
+  setBannedAssets(newBannedAssets: Array<PurgatoryAssets>): void {
+    this.bannedAssets = newBannedAssets
+  }
+
+  async parsePurgatoryAssets(): Promise<Array<PurgatoryAssets>> {
+    try {
+      const response = await axios({
+        method: 'get',
+        url: process.env.ASSET_PURGATORY_URL,
```
alexcos20, line 54: we already have them in config, don't do process.env :)

src/components/Indexer/purgatory.ts
```diff
@@ -0,0 +1,162 @@
...
+
+export class Purgatory {
+  private db: Database
+  private bannedAccounts: Array<PurgatoryAccounts>
+  private bannedAssets: Array<PurgatoryAssets>
+  static instance: any
+
+  constructor(db: Database) {
+    if (
+      !existsEnvironmentVariable('ASSET_PURGATORY_URL', true) ||
```
alexcos20, line 16: we already have them in config, don't do process.env :)

src/components/Indexer/purgatory.ts
```diff
@@ -0,0 +1,162 @@
...
+      )
+    }
+  }
+
+  async parsePurgatoryAccounts(): Promise<Array<PurgatoryAccounts>> {
+    const purgatoryAccounts: Array<PurgatoryAccounts> = []
+    try {
+      const response = await axios({
+        method: 'get',
+        url: process.env.ACCOUNT_PURGATORY_URL,
```
alexcos20, line 93: we already have them in config, don't do process.env :)

src/components/Indexer/processor.ts
alexcos20: for now, in purgatory means: 
 - log the warning
 - do not add ddo to db

src/components/Indexer/purgatory.ts
```diff
@@ -0,0 +1,165 @@
...
+
+  constructor(db: Database, config: OceanNodeConfig) {
+    if (!config.accountPurgatoryUrl || !config.assetPurgatoryUrl) {
+      INDEXER_LOGGER.log(
+        LOG_LEVELS_STR.LEVEL_ERROR,
+        `Cannot instantiate Purgatory due to missing env ASSET_PURGATORY_URL or ACCOUNT_PURGATORY_URL.`
+      )
+      return
+    }
+    this.db = db
```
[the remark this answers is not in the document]
paulo-ocean: from the config, seems you only need the `accountPurgatoryUrl` and `assetPurgatoryUrl`, so probably would make more sense to pass those as param on the constructor, not the config itself, not the db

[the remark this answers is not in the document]
mariacarmina: I did the following:

- getInstance receives no parameter (calls `getConfiguration` inside the function)
- Purgatory class receives accountPurgatoryUrl and assetPurgatoryUrl and removed config and db
- changed type for instance from any into Purgatory

I got them covered? Thank you for review!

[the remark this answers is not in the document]
mariacarmina: > changed type for instance from any into Purgatory

It fails the lint :(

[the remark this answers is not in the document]
paulo-ocean: yes, we already use the comment `// eslint-disable-next-line no-use-before-define` on other places, no problem with that

src/components/Indexer/purgatory.ts
```diff
@@ -0,0 +1,153 @@
...
+    for (const asset of purgatoryAssets) {
+      if (asset.did === refDid) {
+        return true
+      }
+    }
+    return false
+  }
+
+  static async getInstance(): Promise<Purgatory> {
+    const config = await getConfiguration()
```
paulo-ocean, line 144: I'm afraid this still might not be correct @mariacarmina :-(
First, you only need read the configuration once, when there is no instance created.. all the other calls (when Purgatory.getInstance() is called from processor.ts, are useless
Second, from the config it looks like `ASSET_PURGATORY_URL` and `ACCOUNT_PURGATORY_URL` are not mandatory (they default to empty string)
But on your `axios` calls you're not checking the urls at all, so these would constantly throw errors every time  they get called from processor.ts code

Why don't you simply  read the configuration on processor
**and only if the config** is set for both ASSET_PURGATORY_URL and ACCOUNT_PURGATORY_URL
call
```
const purgatory = await Purgatory.getInstance(accountUrl, assetUrl)
const updatedDDO = await this.updatePurgatoryStateDdo(ddo, from, purgatory)
```
thanks

mariacarmina: Can you check now please if it's improved?

paulo-ocean: Unfortunately this is NOT ok @mariacarmina , seems you basically ignored my previous comments :-(

Here's what we have on config:
```
 accountPurgatoryUrl: getEnvValue(process.env.ACCOUNT_PURGATORY_URL, ''),
 assetPurgatoryUrl: getEnvValue(process.env.ASSET_PURGATORY_URL, '')
```
You assign default values (empty string) if the .env vars are not set (which is OK, they are not mandatory, neither should have defaults by previous comments)

every time the indexer runs you have this:
`
const purgatory = await Purgatory.getInstance()
const updatedDDO = await this.updatePurgatoryStateDdo(ddo, from, purgatory)
`
On the sigleton method there is the following:
```
      if (!config.accountPurgatoryUrl && !config.assetPurgatoryUrl) {
        INDEXER_LOGGER.logMessage(`Purgatory URLs are not provided`)
        return
      }
```
Basically, if the .env vars are not set (which is fine), the instance is not created at all.

So, back to the indexer, every time this code runs:

```
...
const updatedDDO = await this.updatePurgatoryStateDdo(ddo, from, purgatory)
...
async updatePurgatoryStateDdo(
    ddo: any,
    owner: string,
    purgatory: Purgatory
  ): Promise<any> {
if (
      (await purgatory.isBannedAsset(ddo.id)) ||
      (await purgatory.isBannedAccount(owner))
    ) {
...

```
We get exceptions like:

```
{
  component: 'INDEXER',
  moduleName: 'INDEXER',
  level: 'error',
  message: "[INDEXER] => Error processMetadataEvents: TypeError: Cannot read properties of undefined (reading 'isBannedAsset')",
  timestamp: '2024-02-20 14:05:26'
}
```

Because there is no instance of purgatory created whatsoever..

Apart from that, inside
` async parsePurgatoryAssets(): Promise<Array<PurgatoryAssets>> {`
OR
`async parsePurgatoryAccounts(): Promise<Array<PurgatoryAccounts>> {`
There is no check for the URLs (even if they were configured with .env vars), there is no check anywhere on the URLs...  so i we set wrong (badly formed urls)
to
ASSET_PURGATORY_URL and ACCOUNT_PURGATORY_URL
We get errors every time these get called as well

paulo-ocean: I' m gonna approve this AS IS and create another task to fix the remaining issues, this was open at the end of January, IMO it doesn't make much sense to keep loosing time back and fourth with it..
I'll merge it, than create another small PR with the fixes
thanks

paulo-ocean (approved):
approved with some comments that will be addressed on a separate PR :-)

PR #248, opened 2024-01-29
add config option for network interfaces, p2p and http

### LINKED ISSUE #243: Presence of HTTP and P2P interfaces should be configurable (but now are hardcoded as 'true')
As part of the the configuration we have:

```
...
hasHttp: true,
hasP2P: true,
...
```
Always hardcoded to 'true', while this was supposed to be configurable (via .env) as some nodes can have only HTTP interface, and others only P2P or both.

Many of the logic we have on the nodes is also relying that these components are always present (accessing certain properties), which at the moment is true because its wrongly hardcoded, but should be taken a look as well

As a suggestion, i think we could have just one extra .env variable for instance, like "INTERFACES", or "NETWORK_INTERFACES", something like this.. 
and basically would be an array that only accepts 2 possible values: "P2P & "HTTP" case insensitive.
ex:
`INTERFACES = ["P2P","HTTP"] `
`INTERFACES = ["P2P"] `
`INTERFACES = ["HTTP"] `
Maybe, by default (if the .env property is not set) we can assume to have both HTTP and P2P, but always check if there is anything different specified.. 
and update the code accordingly

paulo-ocean:
Fixes #243 .

Changes proposed in this PR:

- 
- 
-

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,12 @@ export PRIVATE_KEY=0x.....
 export RPCS="{ \"1\":{ \"rpc\":\"https://rpc.eth.gateway.fm\", \"chainId\": 1, \"network\": \"mainet\", \"chunkSize\": 100 }, \"137\": { \"rpc\": \"https://polygon.meowrpc.com\", \"chainId\": 137, \"network\": \"polygon\", \"chunkSize\": 100 }, \"80001\": { \"rpc\": \"https://rpc-mumbai.maticvigil.com\", \"chainId\": 80001, \"network\": \"polygon-mumbai\", \"chunkSize\": 100 }}"
 ```
 
+Network interfaces supported by the node ('http' and/or 'p2p'). By default, if not specified otherwise, both are supported. Case insensitive.
+
+```bash
+export INTERFACES=[\"HTTP\",\"P2P\"]
+```
+
 You need to define a database URL if you want to run a database as part of your node. This is required for the tests to pass.
 
 ```bash
--- a/src/utils/config.ts
+++ b/src/utils/config.ts
@@ -237,6 +237,35 @@ function getC2DClusterEnvironment(): C2DClusterInfo[] {
   return clusters
 }
 
+// connect interfaces (p2p or/and http)
+function getNodeInterfaces(isStartup: boolean = false) {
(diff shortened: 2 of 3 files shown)


src/test/unit/networking.test.ts
```diff
@@ -0,0 +1,29 @@
+import { getConfiguration } from '../../utils/index.js'
+import { expect } from 'chai'
+
+describe('Test available network interfaces', async () => {
```
jamiehewitt15, line 4: By using the if statement, we're not getting full test coverage on how the configuration can be set. Can we have 5 different unit tests here where we explicitly set the different potential values of the `INTERFACES` env?

1. nothing set, use default 
2. `INTERFACES=[]`
3. `INTERFACES = ["P2P","HTTP"]` 
4. `INTERFACES = ["P2P"]` 
5. `INTERFACES = ["HTTP"]`

paulo-ocean: I've tested it manually, not sure if we really need 5 tests for a simple .env configuration. IMO, we're focusing to much on unit tests and less on code, i'm not sure that is the path we want to follow. I've seen a handful of regressions on the code on the past weeks precisely because we're focusing on having tests passing instead of fixing the underlying code. 
Nevertheless, i can add the tests, no problem..just wanted to avoid messing up with .env variables as much as possible
tkx

jamiehewitt15: > I've seen a handful of regressions on the code on the past weeks precisely because we're focusing on having tests passing instead of fixing the underlying code.

Yeah that definitely shouldn't happen, the idea of having tests is the exact opposite - to prevent the code from regressing. 

5 tests might be overkill but it would be good to test at least two scenarios (1) default and (2) with envs set. The idea is just that these tests would prevent someone from breaking this feature in the future. The current test only checks one scenario in the if statement.

paulo-ocean: added a few more tests, and did a couple of improvements to better account for .env changes on tests

jamiehewitt15 (approved):
Looks good to me

alexcos20 (approved):

Issue #230, opened 2024-01-21
C2D:  initializeCompute


alexcos20:
Implement  initializeCompute command:

- [ ] HTTP:   POST /api/services/initializeCompute
- [ ] P2P Command:
```json
{
command: "initializeCompute",
node: "NodeB ID", // optional, if missing, it refers to current node
algorithm:  {},   //algorithm
datasets: {[]},  // array of datasets
compute: { 
      env: "xx",   // desired compute env
     validUntil: 123 // abs timestamp until compute resources are rented
},
consumerAddress: "0x123"
}
```
For algorithm and datasets objects definition, see https://github.com/oceanprotocol/ocean.js/blob/main/src/%40types/Compute.ts#L62

- [ ] implement core function initializeCompute (see below)

initializeCompute is responsible for:
 - making sure user has access to the datasets
 - datasets are either download type, or compute type (and then they allow that algo to be executed)
 - validUntil is a valid timestamp in the future
 - generate providerFee for one of the inputs (algo preferred)
 - if dataset is served by another provider, reach to that provider and getProviderFees

Provider implementation:  https://github.com/oceanprotocol/provider/blob/main/ocean_provider/routes/compute.py#L75

It returns a json object (see https://github.com/oceanprotocol/ocean.js/blob/main/src/%40types/Provider.ts#L25)

alexcos20:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/S1---DApps-63f8b8c26e838c0017f34d23/issues/oceanprotocol/ocean-node/230?planning-poker) with Zenhub @an7e @bogdanfazakas @jamiehewitt15 @mariacarmina @paulo-ocean

PR #252, opened 2024-01-31
Provider fees compute

### LINKED ISSUE #228: C2D:  ProviderFees
We need to write the provider fees logic for c2d, which is going to be used by both initializeCompute and startCompute.

In C2D, you can have:
 - one optional algo
 - one or more optional datasets
 
You can use any of that to pay the compute fees.  IE:  Pay them when you buy the algo.. Or pay them when you buy one of the datasets.

We should always try to have the providerFees in the algo tx, if possible

Functions to implement:

- [ ] calculateComputeProviderFee  - given an env and an absolute timestamp, generate the providerFee amount, data,signature  (see https://github.com/oceanprotocol/provider/blob/main/ocean_provider/utils/provider_fees.py#L178 and https://github.com/oceanprotocol/ocean-node/blob/develop/src/components/core/utils/feesHandler.ts#L21)
- [ ] validateComputeProviderFee  - given a tx and an compute env, check if providerFee is still valid

mariacarmina:
Fixes #228

Changes proposed in this PR:

- added calculate ProviderFees for compute and also validate provider fees for compute functions, which will be used for initialize compute endpoint
- fix tests
- improve error logs


src/@types/Fees.ts
```diff
@@ -16,7 +18,7 @@ export type FeeStrategy = {
 export interface ProviderFeeData {
   providerFeeAddress: string
   providerFeeToken: string
-  providerFeeAmount: number
+  providerFeeAmount: number | BigNumberish
```
jamiehewitt15, line 21: Does the `BigNumberish` type accept a `BigInt`?

src/components/core/utils/feesHandler.ts
```diff
@@ -1,11 +1,208 @@
...
+    }
+    return envs
+  } catch (error) {
+    CORE_LOGGER.log(LOG_LEVELS_STR.LEVEL_ERROR, `Error identifying C2D envs: ${error}`)
+  }
+}
+
+async function getEnv(asset: DDO, computeEnv: string): Promise<any> {
+  const computeEnvs = await getC2DEnvs(asset)
+  CORE_LOGGER.logMessage(`compute envs: ${computeEnvs}`)
```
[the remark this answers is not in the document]
jamiehewitt15: Usually it's safer to avoid logging any environmental variables and instead just log a message if something is missing. The same would also apply to the other logs below.

src/components/core/utils/feesHandler.ts
```diff
@@ -1,11 +1,208 @@
...
+  // For Solidity, we need the expanded-format of a signature
+  const signatureSplitted = ethers.Signature.from(signed32Bytes)
+  // console.log(
+  //   'verify message:',
+  //   await verifyMessage(
+  //     ethers.toBeArray(signableHash), // 32 bytes again
+  //     providerWallet.address,
+  //     signed32Bytes
+  //   )
+  // )
```
jamiehewitt15, line 132:
```suggestion
```

src/components/core/utils/feesHandler.ts
```diff
@@ -1,11 +1,208 @@
...
+  const iface = new ethers.Interface(abi)
+  const receipt = await provider.getTransactionReceipt(transactionHash)
+  const eventObj = {
+    topics: receipt.logs[0].topics as string[],
+    data: receipt.logs[0].data
+  }
+  return iface.parseLog(eventObj)
+}
+
+// TODO tests in initializeCompute issue
```
jamiehewitt15, line 167: Is this something that you're planning to do in this PR? If not, let's make sure there is an issue for it

src/components/core/utils/feesHandler.ts
```diff
@@ -15,6 +212,7 @@ export async function createFee(
   service: Service
   // provider: OceanProvider // this node provider
```
jamiehewitt15, line 213: Let's remove this commented-out parameter if it is not needed

```suggestion
```

src/components/core/utils/feesHandler.ts
```diff
@@ -40,21 +238,21 @@ export async function createFee(
   const providerFeeAmount: number = await getProviderFeeAmount() // TODO check decimals on contract?
```
jamiehewitt15, line 238: > TODO check decimals on contract?

That needs to be implemented, otherwise we will quickly run into issues when the fees are in USDC

src/components/core/utils/feesHandler.ts
```diff
@@ -285,6 +483,13 @@ export async function getProviderFeeToken(chainId: number): Promise<string> {
   return result.length ? result[0].token : ethers.ZeroAddress
 }
 
+export async function getProviderFeeTokenByArtifacts(chainId: number): Promise<string> {
+  if (chainId === 8996) {
+    return getOceanArtifactsAdresses().development.Ocean
+  }
```
jamiehewitt15, line 489: Is there any reason you didn't put this logic in the `getProviderFeeToken` function and then call that one directly? It would be simpler for others to work with if there is just one function and would help avoid mistakes or inconsistencies.

src/components/core/validateTransaction.ts
```diff
@@ -1,13 +1,44 @@
...
+    try {
+      CORE_LOGGER.logMessage(`fetching tx hash retry ${retries}....`)
+      const txReceipt = await provider.getTransactionReceipt(txId)
+      CORE_LOGGER.logMessage(`txReceipt: ${JSON.stringify(txReceipt)}`)
+      if (txReceipt) {
+        return txReceipt
+      }
+      CORE_LOGGER.logMessage(
+        `txReceipt not found #${retries}: ${JSON.stringify(txReceipt)}`
+      )
```
jamiehewitt15, line 29: To avoid creating a lot of noise in the logs when the retries number is set high, it would be nice to only log this message when we are on the final try (and then the other logs below can also be removed). Also, there is no point logging the `txReceipt` if one hasn't been recieved. For example: 

```suggestion
    retries--
    if(retries === 0){
    CORE_LOGGER.logMessage(
        `txReceipt not found after #${retries} tries.`
      )
    }
```

src/components/core/validateTransaction.ts
```diff
@@ -1,13 +1,44 @@
...
+      await sleep(30000)
```
jamiehewitt15, line 30: We shouldn't wait another 30 seconds on the last try, that's just delaying the time until we send the client an error message...

src/components/core/utils/feesHandler.ts
mariacarmina: I get the decimals for calculateProviderFeeCompute. That TODO was for provider fees for download.

src/components/core/utils/feesHandler.ts
mariacarmina: No, I'll delete it, I am assigned to initializeCompute task #230 just wanted to know the purpose of that function

src/components/core/utils/feesHandler.ts
mariacarmina: `getProviderFeeToken` function is used for download part, and in tests there are some places where the fee token is hardcoded and does not match the address of Ocean token from development network.

src/components/core/validateTransaction.ts
mariacarmina: Here in my opinion should be a separate task to wait for a tx receipt depending on the network processing blocks time, polling interval etc.. I'll do a check to avoid another sleep.

src/components/core/utils/feesHandler.ts
mariacarmina: Should we modify code which is not changed in this PR?

src/components/core/utils/feesHandler.ts
mariacarmina: It is not in scope of this task though and that function has not been changed in this PR.

src/components/core/utils/feesHandler.ts
jamiehewitt15: This is just a comment, so no harm in tidying it up

src/components/core/utils/feesHandler.ts
jamiehewitt15: If it's only applicable for download, then I'd suggest we just make sure we're tracking it in an issue

mariacarmina: I created separated functions for compute when calculating provider fees.

mariacarmina: https://github.com/oceanprotocol/ocean-node/issues/262 done

src/components/core/validateTransaction.ts
mariacarmina: #261 done

src/@types/Fees.ts
jamiehewitt15: ok, good to know

src/components/core/utils/feesHandler.ts
jamiehewitt15: Thanks

src/components/core/utils/feesHandler.ts
```diff
@@ -1,11 +1,208 @@
...
+    id: service.id
+  }
+  const providerWallet = await getProviderWallet(String(asset.chainId))
+  const providerFeeAddress: string = providerWallet.address
+  let providerFeeAmount: number
+  let providerFeeAmountFormatted: BigNumberish
+
+  const providerFeeToken: string = await getProviderFeeTokenByArtifacts(asset.chainId)
+
+  if (providerFeeToken === '0x0000000000000000000000000000000000000000') {
```
paulo-ocean, line 79: better use constants instead, like `ZeroAddress`, this  looks a bit noisy when appearing multiple times :-)
Also i would avoid logging stuff like:
`CORE_LOGGER.logMessage(`provider fee amount: ${providerFeeAmount}`)` OR
`CORE_LOGGER.logMessage(`provider fee amount formatted: ${providerFeeAmountFormatted}`)`
This seems more like debugging code... but there is no need to log these, too much noise
thanks

an7e (approved):

src/components/core/utils/feesHandler.ts
mariacarmina: There were debugging logs yes, sure I'll use ZeroAddress.

src/components/core/utils/feesHandler.ts
```diff
@@ -1,11 +1,198 @@
...
+        [`${cluster}api/v1/operator/environments?chain_id=${asset.chainId}`]: data
+      })
+    }
+    return envs
+  } catch (error) {
+    CORE_LOGGER.log(LOG_LEVELS_STR.LEVEL_ERROR, `Error identifying C2D envs: ${error}`)
+  }
+}
+
+async function getEnv(asset: DDO, computeEnv: string): Promise<any> {
```
paulo-ocean, line 41: on the config.ts there is already a property named `c2dClusters`:
parsed on:
` function getC2DClusterEnvironment(): C2DClusterInfo[] {` 
responsible for parsing the .env variable (if present)
we should use that instead of parsing .env again. This function here will also give an exception if the property is not set (its not required, so that is a common scenario)

jamiehewitt15 (approved):

src/components/core/utils/feesHandler.ts
mariacarmina: done

src/components/core/utils/feesHandler.ts
```diff
@@ -1,11 +1,203 @@
...
+import { findEventByKey } from '../../Indexer/utils.js'
+import axios from 'axios'
+import { getOceanArtifactsAdresses } from '../../../utils/address.js'
+import ERC20Template from '@oceanprotocol/contracts/artifacts/contracts/templates/ERC20TemplateEnterprise.sol/ERC20TemplateEnterprise.json' assert { type: 'json' }
+import { C2DClusterInfo } from '../../../@types'
+
+export async function getC2DEnvs(asset: DDO): Promise<Array<any>> {
+  try {
+    const envs: Array<any> = []
+    const clustersURLS: string[] = JSON.parse(
```
paulo-ocean, line 26: Ok, i'm gonna approve this as it had already 2 approvals, but it might need some refactor afterwards 
IMO, we should try to take a deep breath and second look when reviewing PRs.. :-)
For instance...
We are still processing `process.env.OPERATOR_SERVICE_URL` on `getC2DEnvs`
We only changed on `getEnv`
but on `getEnv` we have:
```
const computeEnvs = await **getC2DEnvs**(asset)  
  const clustersURLS: string[] = []
  const clustersInfo: C2DClusterInfo[] = (await getConfiguration()).c2dClusters
```
And on `getC2DEnvs` called by the previous one, we have again:
const clustersURLS: string[] = JSON.parse(
      process.env.OPERATOR_SERVICE_URL
    ) as string[]
    
 So, we are parsing it twice, instead of just 1 time....  and in 2 different ways :-)

paulo-ocean (approved):

src/components/core/utils/feesHandler.ts
alexcos20: agreed, let's refactor now, not later

mariacarmina: I'll change right now, it's not an issue

mariacarmina: done

PR #254, opened 2024-02-02
229 c2d validation logic

### LINKED ISSUE #229: C2D:  validation logic
We need to write the validation logic for c2d, which is going to be used by both initializeCompute and startCompute

Functions to implement:

- [ ] checkC2DEnvExists (env)  - check if env exists, by getting envs from all cluster.  User cannot start a compute using an invalid environment
- [ ] getAlgoChecksums  - given a algo did or algo ddo, get algo file checksum (by calling FileInfo) and compute container checksum (see https://github.com/oceanprotocol/provider/blob/main/ocean_provider/validation/algo.py#L325).   This will be used to match dataset requirments
- [ ] validateAlgoForDataset  - given an algo DID + algoChecksums(from above), validate that dataset has no restrictions (see https://docs.oceanprotocol.com/developers/compute-to-data/compute-options#compute-options).  This way, we can ensure that only allowed algos can be run on top of a dataset
- [ ] validateConsumerParameters  - given an input array of keys->values, validate that we have required fields or defaults are filled  (see https://docs.oceanprotocol.com/developers/compute-to-data/compute-options#consumer-parameters)

### LINKED ISSUE #253: Storage:  missing checksums
When https://github.com/oceanprotocol/ocean-node/issues/105 was defined,  the requirements were to add checksum as well.

By looking in the code while debugging something with Ilia, we discovered that there is no checksum for files.

So, let's implement in the following way:

- define an ENV  "MAX_CHECKSUM_LENGTH ": define the maximum length for a file if checksum is required. If file size is greater, we do not compute checksum.  Default value: 10 Mb
- update function storage.getFileInfo  to take extra parameter:  forceChecksum (default false)
- update function storage.getFileInfo  and underlying storage classes to add checksum to output:
    - if ( forceChecksum==false and filesize=<MAX_CHECKSUM_LENGTH) || (forceChecksum ) -> compute hash
    - else  do not compute hash

In this way, we can have checksums for datasets only if they are smaller than a predefined size, but always compute checksums for algos

an7e:
Fixes #229 
Fixes #253

Changes proposed in this PR:

- 
- 
-


src/test/integration/completeFlow.test.ts
```diff
@@ -155,7 +155,7 @@ describe('Should run a complete node flow.', () => {
     const fileInfo = await streamToObject(response.stream as Readable)
 
     assert(fileInfo[0].valid, 'File info is valid')
-    expect(fileInfo[0].contentLength).to.equal('417')
+    expect(fileInfo[0].contentLength).to.equal('946')
```
jamiehewitt15, line 158: Why did the content length change when there haven't been any other changes to this file?

src/test/integration/c2d.test.ts
```diff
@@ -0,0 +1,351 @@
...
+      '0x00',
+      metadata,
+      '0x' + hash,
+      []
+    )
+    const txReceipt = await setMetaDataTx.wait()
+    assert(txReceipt, 'set metadata failed')
+  })
+
+  delay(50000)
```
jamiehewitt15, line 264: Is it really necessary to have such a long delay?

an7e: @jamiehewitt15 unfortunately yes, there are two publishing in the test, dataset ddo and algo ddo, sometimes 30000 wasn't enough time

src/test/integration/completeFlow.test.ts
jamiehewitt15: Ok cool, yeah that makes sense

src/test/integration/c2d.test.ts
jamiehewitt15: And we can't use the `waitToIndex` function to make sure that we only wait as long as necessary?

src/components/core/fileInfoHandler.ts
```diff
@@ -58,13 +58,18 @@ async function formatMetadata(file: ArweaveFileObject | IpfsFileObject | UrlFile
       ? (file as IpfsFileObject).hash
       : null
 
-  const { contentLength, contentType } = await fetchFileMetadata(url)
+  const { contentLength, contentType, contentChecksum } = await fetchFileMetadata(
```
alexcos20, line 61: you wait for contentChecksum , but the function is not returning that..

src/components/c2d/index.ts
```diff
@@ -0,0 +1,215 @@
...
+
+function checkBoolean(value: any) {
+  return typeof value === 'boolean' || value instanceof Boolean
+}
+
+function checkNumber(value: any) {
+  return typeof value === 'number' || value instanceof Number
+}
+
+export function validateConsumerParameters(consumerParameters: any) {
```
alexcos20, line 159: what is the purpose of this?

if we want to validate something, we need to validate an user object of responses against the consumer parameter requirments.

IE: Are all mandatory params given? Are types ok? etc

src/test/integration/c2d.test.ts
```diff
@@ -0,0 +1,351 @@
...
+      algoDDOTest.id,
+      algoChecksums,
+      datasetDDOTest.id,
+      datasetDDOTest.services[0].id,
+      oceanNode
+    )
+    expect(result).to.equal(true)
+  })
+
+  it('should validateConsumerParameters', async () => {
```
alexcos20, line 299: see my other comment regarding validateConsumerParameters

src/test/integration/c2d.test.ts
an7e: @jamiehewitt15 i checked your suggestion and removed delay, thank you

src/components/c2d/index.ts
an7e: fixed ConsumerParameter type

src/test/integration/c2d.test.ts
an7e: fixed ConsumerParameter type

src/test/integration/c2d.test.ts
paulo-ocean: we need to get rid of the delays, we should listen events or use block confirmations instead 
something to improve later for sure :-)

src/components/c2d/index.ts
alexcos20: again, this function needs two parameters...

```
validateConsumerParameters(DDOconsumerParameters, userSentObject)
```

where:
 - DDOconsumerParameters:  is the object defined in the DDO, which defines what parameters are required, defaults, types
  - userSentObject:  is the actual object send by the user
  
  We need to make sure that:
   - check if user is missing some required params
   - whatever user sent complies with definition (ie: string is a string, select box option exists, etc)
   - put defaults if needed
   
   The flow is the following:
    - download:       Publisher defines some params, like 'days' , which represents how many days in the past the dataset should contain.   When consumes, user specifies  '5'  (ie:  last 5 days of data).
    - compute:       Publisher defines some params for algo, like 'iterations' (how many iterations to perform, number, default 10) and 'output' (which is a select box with 'pdf', 'csv'  options).  When user starts a c2d, we need to validate, fill default, and send that to c2d

alexcos20: also, this is not c2d specific, please move it to utils  (it's used by both download and compute)

alexcos20 (changes requested):

src/test/integration/c2d.test.ts
```diff
@@ -0,0 +1,350 @@
...
+    datasetDDO = { ...publishDatasetDDO }
+    factoryContract = new ethers.Contract(
+      artifactsAddresses.development.ERC721Factory,
+      ERC721Factory.abi,
+      publisherAccount
+    )
+    process.env.PRIVATE_KEY =
+      '0xc594c6e5def4bab63ac29eed19a134c130388f74f019bc74b8f4389df2837a58'
+    process.env.RPCS = JSON.stringify(mockSupportedNetworks)
+    process.env.AUTHORIZED_DECRYPTERS = JSON.stringify([publisherAddress])
```
jamiehewitt15, line 73: Let's avoid overwritting the environmental variables as that can easily cause problems. Instead we should use `setupEnvironment()` / `tearDownEnvironment()`

jamiehewitt15: https://github.com/oceanprotocol/ocean-node/tree/develop?tab=readme-ov-file#unit-and-integration-environments

src/test/integration/c2d.test.ts
```diff
@@ -0,0 +1,401 @@
...
+  getMockSupportedNetworks,
+  setupEnvironment,
+  tearDownEnvironment
+} from '../utils/utils.js'
+
+describe('C2D functions', async () => {
+  let config: OceanNodeConfig
+  let database: Database
+  let oceanNode: OceanNode
+  let indexer: OceanIndexer
```
paulo-ocean, line 51: Seems Indexer is not used, we can safely delete it
same for line:
`indexer = new OceanIndexer(database, mockSupportedNetworks)`

src/test/integration/c2d.test.ts
```diff
@@ -0,0 +1,401 @@
...
+  let datasetDDO: any
+
+  const chainId = 8996
+  const mockSupportedNetworks: RPCS = getMockSupportedNetworks()
+
+  let previousConfiguration: OverrideEnvConfig[]
+
+  before(async () => {
+    const artifactsAddresses = getOceanArtifactsAdresses()
+    provider = new JsonRpcProvider('http://127.0.0.1:8545')
```
paulo-ocean, line 66: Just a side note:
This is the kind of code that only works on ganache/local network
```
publisherAccount = (await provider.getSigner(0)) as Signer
publisherAddress = await publisherAccount.getAddress()
```
Only with ganache/barge running we have a signer from `provider.getSigner(0)` because it creates some default accounts.. We had a lot of broken things on the indexer because we we're relying on this kind of code, inside it.. 
Outside ganache this would throw "No Account" errors
We should start  using the **blockchain** class instead... which associates the wallet from the PRIVATE_KEY to the provider/RPC 

Not an issue for the tests here, but better keep that in mind for other code :-)

PS: there are things from develop branch to be merged first
thanks

src/test/integration/c2d.test.ts
paulo-ocean: ok, nevermind... just saw that we are waiting for index in some tests :-)

src/components/c2d/index.ts
```diff
@@ -0,0 +1,144 @@
...
+import { createHash } from 'crypto'
+import { FindDdoHandler } from '../core/ddoHandler.js'
+import { decrypt } from '../../utils/crypt.js'
+import { Storage } from '../storage/index.js'
+import { getConfiguration } from '../../utils/config.js'
+import { GetEnvironmentsHandler } from '../core/compute.js'
+import { PROTOCOL_COMMANDS } from '../../utils/constants.js'
+import { streamToObject } from '../../utils/util.js'
+import { Readable } from 'stream'
+import { ConsumerParameter } from '../../@types/DDO/ConsumerParameter.js'
```
jamiehewitt15, line 12: > 'ConsumerParameter' is defined but never used
'ERC721Factory' is defined but never used   
'ERC721Template' is defined but never used   

Can you remove all of the unused dependencies in this file?

src/components/c2d/index.ts
```diff
@@ -0,0 +1,144 @@
...
+import ERC721Factory from '@oceanprotocol/contracts/artifacts/contracts/ERC721Factory.sol/ERC721Factory.json'
+import ERC721Template from '@oceanprotocol/contracts/artifacts/contracts/templates/ERC721Template.sol/ERC721Template.json'
+
+export async function checkEnvironmentExists(envId: string, oceanNode: OceanNode) {
```
mariacarmina, line 16: I think @paulo-ocean also have a function that checks for env vars `existsEnvironmentVariable` from config.ts. Let's use that.

mariacarmina (commented):
small comment from my side

src/components/c2d/index.ts
an7e: @jamiehewitt15 yeah, code was moved

src/components/c2d/index.ts
an7e: @mariacarmina how should I use it? could you show an example or link to example?

mariacarmina: https://github.com/oceanprotocol/ocean-node/pull/247/commits/f7a3816d992b64a17be1ad91c0e65a5dbe0f7c0e#diff-b31c00271d46ab655ddf80b2ec12ac06d6e5376ff2fe743fc099cb4a44e457dcL16 I have here an example how I used

mariacarmina: But for here it was not applicable, name confusion.

src/utils/validators.ts
```diff
@@ -0,0 +1,81 @@
...
+  return typeof value === 'number' || value instanceof Number
+}
+
+function checkObject(value: any) {
+  return typeof value === 'object' && value !== null && !Array.isArray(value)
+}
+
+export function validateConsumerParameters(
+  ddoConsumerParameters: ConsumerParameter[],
+  userSentObject: any[]
```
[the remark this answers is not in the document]
paulo-ocean: after another look, i guess it could make some sense the array, if we want to pass 1 or + objects, and even for 1, it might be easier to work with an array, at least the way the validation was coded... 
Regrading the throw-> try/catch , well.. if it was me, i would probably had chosen  in a different way, i'm not a big fan of throwing errors just to catch them, but overall i think it does the job its supposed to do, and its just a different approach

paulo-ocean:
fyi, there is an integration test that fails randomly on multiple branches, seems related with the ddo state DB
there is already another open PR for investigating it
[https://github.com/oceanprotocol/ocean-node/pull/389](url)

bogdanfazakas (approved):
LGTM,
small remarks but should be good

src/test/integration/compute.test.ts
```diff
@@ -559,6 +594,221 @@ describe('Compute', () => {
     console.log(jobs)
   })
 
+  // algo and checksums related
+  it('should publish AlgoDDO', async () => {
```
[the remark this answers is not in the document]
paulo-ocean: moved the test and update the test description

paulo-ocean:
hey @alexcos20 do you want to have another look at it?
thanks

paulo-ocean:
merging as all previous questions/changes have been addressed

PR #258, opened 2024-02-08
Feat/develop dashboard

tom1145:
Changes proposed in this PR:

- added node dashboard made in nextJS


dashboard/yarn.lock
jamiehewitt15, line 1: Let's use `npm` rather than yarn to keep it consistent with the rest of the project

dashboard/README.md
```diff
@@ -0,0 +1,40 @@
+This is a [Next.js](https://nextjs.org/) project bootstrapped with [`create-next-app`](https://github.com/vercel/next.js/tree/canary/packages/create-next-app).
+
+## Getting Started
+
```
jamiehewitt15, line 4: We need to update the readme to explain how to set up the config. Otherwise, users end up seeing this:

```
Unhandled Runtime Error
TypeError: Failed to fetch

Source
src/components/Dashboard/index.tsx (62:4) @ fetch

  60 | useEffect(() => {
  61 |   setLoading(true)
> 62 |   fetch(config.apiURL.status, {
     |  ^
  63 |     headers: {
  64 |       Accept: 'application/json',
  65 |       'Content-Type': 'application/json'
```

dashboard/README.md
```diff
@@ -0,0 +1,40 @@
...
+First, run the development server:
+
+```bash
+npm run dev
+# or
+yarn dev
+# or
+pnpm dev
+# or
+bun dev
```
jamiehewitt15, line 14:
```suggestion
npm i
npm run dev
```

dashboard/README.md
jamiehewitt15: Even after setting the apiURL for my local node and the nodeId, I'm still getting this error. is there anything that I'm missing from the way you are setting it up? There's no issue with my node as I'm able to make the status request fine when using postman.

dashboard/config.ts
```diff
@@ -0,0 +1,19 @@
...
+    logs: string
+    status: string
+  }
+  nodeId: string
+}
+
+const config: ConfigType = {
+  apiURL: {
+    logs: 'http://localhost:8010/proxy/logs',
+    status: 'http://localhost:8010/proxy/directCommand',
```
jamiehewitt15, line 12: Is there any particular reason why you are using the direct command endpoint, rather than the status endpoint?

dashboard/config.ts
```diff
@@ -0,0 +1,19 @@
+export type ConfigType = {
+  apiURL: {
+    logs: string
+    status: string
+  }
+  nodeId: string
+}
+
+const config: ConfigType = {
+  apiURL: {
```
jamiehewitt15, line 10: Rather than having this object of URLs, I'd suggest we just set one env for the `apiURL` and then we can add the paths as and when they are needed e.g. `${apiURL}/logs`

dashboard/config.ts
```diff
@@ -0,0 +1,19 @@
...
+    logs: 'http://localhost:8010/proxy/logs',
+    status: 'http://localhost:8010/proxy/directCommand',
+  },
+  nodeId:
+    process.env.NEXT_PUBLIC_NODE_ID ||
+    '16Uiu2HAm6G73f3wiPtrmT7BNLqA3PCzQGGtpShu9qNFjAJgAM4R7',
```
jamiehewitt15, line 16: This default value doesn't make any sense. It would be better to send no `nodeId` rather than a random one as it is not a required parameter. The node will by default you make the request to will give you the status of itself by default

dashboard/package.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "ocean-nodes",
```
jamiehewitt15, line 2:
```suggestion
  "name": "ocean-node-dashboard",
  "author": "Ocean Protocol Foundation",
  "license": "Apache-2.0",
```

dashboard/README.md
tom1145: Hello! I also got the same error on my end. From what I investigated, it is a CORS error. I used [this](https://github.com/garmeeh/local-cors-proxy) for a workaround. And it worked well after.

dashboard/config.ts
tom1145: I checked here but couldn't find the route you mentioned, so I used the `directCommand` instead.
Screenshot 2024-02-16 at 15 17 51

dashboard/config.ts
tom1145: I've added apiURL for the base API URL and apiRoutes object for each API path.

dashboard/config.ts
tom1145: I forgot about this one, my bad. I removed it and added a default empty string.

dashboard/config.ts
jamiehewitt15: A `GET` request to the root endpoint `/` returns the status: 

image

dashboard/config.ts
jamiehewitt15: Thank you

dashboard/config.ts
jamiehewitt15: Thank you

dashboard/additional.d.ts
```diff
@@ -0,0 +1,3 @@
+interface Window {
```
jamiehewitt15, line 1: > 'Window' is defined but never used

If it's not being used, let's remove it

jamiehewitt15:
Pull the latest changes from `origin develop` to fix the tests

jamiehewitt15:
We also need to update the build process to include building this next.js app and including it in the express app.

jamiehewitt15:
Can you also please fix the linting issues, thanks

paulo-ocean (commented):
.

dashboard/.env.sample
```diff
@@ -0,0 +1,2 @@
+NEXT_PUBLIC_NODE_ID=
```
paulo-ocean, line 1: Is this a sample? What exactly should we put here?
some example/explanation would be nice
tkx

dashboard/app.js
```diff
@@ -0,0 +1,22 @@
+const express = require('express')
+
+const app = express()
```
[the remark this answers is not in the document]
paulo-ocean: maybe an extra script on package.json to build/start both the node and the dashboard ?
we might not want the dashboard running for instance, but the dashboard alone doesn't make sense either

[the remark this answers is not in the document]
jamiehewitt15: It's not mandatory but I'd suggest we start it by default as it should be pretty lightweight. And then it would be nice to give the user the option to run the node without a dashboard which could be set via an `env`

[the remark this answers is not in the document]
tom1145: Does this mean we must also move the `package.json` from the dashboard to the main one?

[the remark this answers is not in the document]
jamiehewitt15: No, you should keep the separate `package.json` in the `dashboard` directory. We don't need dependencies like React and Next being installed there when they are not being used. In the `dashboard` directory it's just like working with another react/ next.js project but the main repo only needs the static html/ css/ javascript files. 

So we just need a short script in the main repo that runs `npm run build` in the dashboard directory, Next will then handle building the files. You need to change the output directory of the build process to then serve the static files from the express app in the main repo.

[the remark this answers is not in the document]
tom1145: Thx for the tips, I'm working now on it. For Next.js to build the files we also need to install the dependencies first.

dashboard/.env.sample
paulo-ocean: yep, that's true indeed :+1:

jamiehewitt15: Yeah, I'd just say we don't start the dashboard if the node does not expose http. We can add a note in the readme saying that the dashboard requires http to be exposed on the node for it to run

jamiehewitt15 (approved):
This works now. There is still more work to be done but I suggest that we merge this and then move the remaining improvements into separate issues and create new PRs.

paulo-ocean (approved):
looks good. Of course there is still work to do, but the skeleton is there. Builds & starts alongside with node  which is also nice :+1:

PR #259, opened 2024-02-08
257 ddo metadata proof

### LINKED ISSUE #257: DDO Metadata proof
Now, we are indexing all assets published on chain.  But for some use cases, you need to index only assets validated by somebody.

The flow is the following:
 - as a publisher, I will call validateDDO on some node.
 - node will validate it (with RBAC, or other means) and will decide that I'm a trusted publisher, belonging to a certain group (or whatever)
 - if valid, node will output a signature, called metadataProof 
 - as a publisher, I will also get some validations from other nodes if needed
 - when publishing my assets, I will include those proofs (https://github.com/oceanprotocol/contracts/blob/main/contracts/templates/ERC721Template.sol#L258-L292)
 - all ocean-nodes will catch that ddo and they can check metadataProofs

TO DO:
Validate flow:
  -[ ] add signature (metadataProof) on validateDDO API endpoint  
 Returns:
```
{   
    hash:    sha256(ddo_object),   (0xAB...)
    publicKey:  eth address of node
    v,
    r,
    s
}
```
Reference:   https://github.com/oceanprotocol/aquarius/blob/main/aquarius/app/assets.py#L310

Index flow:
 -[ ] add an optional ENV, called ALLOWED_VALIDATORS, as an array of strings. 
 -[ ] when node indexes the asset and ALLOWED_VALIDATORS is not empty, check if we have a signature from an address from this list
 -[ ] if yes, index the asset
 -[ ] otherwise, log the error

an7e:
Fixes #257

Changes proposed in this PR:

- added signature in ValidateDDOHandler
- added check for MetadataValidated in crawler


src/components/core/utils/validateDdoHandler.ts
```diff
@@ -156,32 +157,24 @@ export async function validateObject(
...
-    const signatureSplitted = ethers.Signature.from(signed)
+    const signatureSplitted = ethers.Signature.from(signed32Bytes)
     const v = signatureSplitted.v <= 1 ? signatureSplitted.v + 27 : signatureSplitted.v
     const r = ethers.hexlify(signatureSplitted.r) // 32 bytes
     const s = ethers.hexlify(signatureSplitted.s)
-
-    values = { hash: hashedRaw, publicKey: wallet.address, r, s, v }
+    return { hash: hashedDDO, publicKey: providerWallet.address, r, s, v }
   } catch (error) {
     console.error(error)
```
paulo-ocean, line 177: cosmetic refactor here, as nothing really changed but ok :-)
you can remove the console.log(error)
thanks

an7e: @paulo-ocean yep, that's right, the key change is to use Uint8Array for the sign and wallet of the provider

src/utils/config.ts
```diff
@@ -99,6 +99,30 @@ function getAuthorizedDecrypters(): string[] {
   }
 }
 
+export function getAllowedValidators(): string[] {
+  try {
+    const allowedValidators: string[] = JSON.parse(process.env.ALLOWED_VALIDATORS)
```
jamiehewitt15, line 104: Can you add a short explanation of this `ALLOWED_VALIDATORS` env to the readme. Otherwise people will get the `Missing or Invalid ALLOWED_VALIDATORS env variable format` error and not know what to set for this value...

jamiehewitt15 (approved):
Looks good to me

src/utils/config.ts
```diff
@@ -99,6 +99,30 @@ function getAuthorizedDecrypters(): string[] {
...
+        'Missing or Invalid ALLOWED_VALIDATORS env variable format',
+        true,
+        GENERIC_EMOJIS.EMOJI_CROSS_MARK,
+        LOG_LEVELS_STR.LEVEL_ERROR
+      )
+      return []
+    }
+    return allowedValidators.map((address) => getAddress(address))
+  } catch (error) {
+    CONFIG_LOGGER.logMessageWithEmoji(
```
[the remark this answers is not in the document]
an7e: @paulo-ocean getAddress throws an Error if the address is invalid, why shouldn't we log this as ERROR ?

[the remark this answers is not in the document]
paulo-ocean: please read my comment again @an7e :-)
What happens if i do not set the .env var?

src/components/Indexer/utils.ts
```diff
@@ -147,6 +153,42 @@ export const processChunkLogs = async (
           `-- ${event.type} -- triggered for ${log.transactionHash}`,
           true
         )
+        if (
+          event.type === EVENTS.METADATA_CREATED ||
+          event.type === EVENTS.METADATA_UPDATED ||
+          event.type === EVENTS.METADATA_STATE
+        ) {
+          if (checkMetadataValidated) {
```
paulo-ocean, line 161: IMO, we could have written this in a much simpler way, and easier to read, with maybe just one 'if' block, without 'continue'... and without the almost duplicate 'if' statement
like: 
- if Validation required 
  - validate it
      - if OK index it
- else index it
      
but anyway, not a blocker..

src/utils/config.ts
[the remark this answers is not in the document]
an7e: @paulo-ocean because this error provided by another check - if (!Array.isArray(allowedValidators))
it needs additional check if env optional with warning, i'll add it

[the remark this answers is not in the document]
paulo-ocean: i know... but if is not mandatory, you could first check if it exists... (if not, you can log as warning for instance... or not log at all) 
and if exists... then anything wrong with it would be an error
Just mentioned this because we have done the same for other "non mandatory" variables... and its misleading when you see the errors in red, for something that is not mandatory, neither needs defaults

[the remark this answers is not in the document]
an7e: @paulo-ocean yep, now it check for optional env, then if it valid

src/components/Indexer/utils.ts
an7e: @paulo-ocean it doesn't matter if we check the event type first and then check the metadata or vice versa, there will still be two checks

paulo-ocean: that's not my point @an7e 
Before we add the following case:
```
if( event.type === EVENTS.METADATA_CREATED ||
event.type === EVENTS.METADATA_UPDATED)
```
And we processed the event

Now we have an extra additional case:
```
if (
          event.type === EVENTS.METADATA_CREATED ||
          event.type === EVENTS.METADATA_UPDATED ||
          event.type === EVENTS.METADATA_STATE
        ) 
```
That only does metadata check/validation OR nothing if `checkMetadataValidated` is false
IMO, there was no need for this additional case, because its pretty much the same

There is the extra:
`event.type === EVENTS.METADATA_STATE`
But why do we need it?, its only taken in account for the block: `if (checkMetadataValidated) {...`
But nothing happens with this state... when the indexing happens.. it will only happen for case:
```
if (
          event.type === EVENTS.METADATA_CREATED ||
          event.type === EVENTS.METADATA_UPDATED
        ) 
```
So, the extra block seems unnecessary

paulo-ocean: that's not my point @an7e  but ok :-)

paulo-ocean (approved):
lgtm

src/utils/config.ts
mariacarmina: Can we add allowed validators to the config as well and invoke the config here instead of relying on existsEnvironmentVariable function?

paulo-ocean: actually it is already part of the config:
```
const config: OceanNodeConfig = {
 authorizedDecrypters: getAuthorizedDecrypters(),
 allowedValidators: getAllowedValidators(),
...
```
he just called the same function used there :-) 
in any case i would avoid calling this all the time.. once the node starts the config does not change (in theory, it should not)
we're calling this every 30 seconds on the index crawler.. 
Anyway.. something to improve later i guess

mariacarmina: aha ok, I understand now what was the approach. thanks!

PR #266, opened 2024-02-14
Issue 265 encrypt file

### LINKED ISSUE #265: HTTP Endpoint:  encryptFile
So, we need to implement encrypt file HTTP Endpoint  (not to be confused with #40)

There are two ways of encrypting a file:

1. Sending the file object.
     - Content-type header must be set to application/json
     - Body contains file object
     - Query.encryptMethod can be aes or ecies (if missing, defaults to aes)
 

2. Sending the raw content
     - Content-type header must be set to application/octet
     - Body contains raw file content (can be huge)
     - Query.encryptMethod can be aes or ecies (if missing, defaults to aes)

3. TBD:  Should support  content-type: multipart as well ?

For all cases, we need to return:
 - Body:  encrypted file content
 - Headers  (so caller knows how to construct his files object for publish the asset)
     - X-Encrypted-By:  our_node_id
     - X-Encrypted-Method: aes or ecies

jamiehewitt15:
Fixes #265

Changes proposed in this PR:

- 
- 
-


jamiehewitt15:
Something weird going on with the tests right now, not just in this PR. I would still be grateful for reviews and feedback on this while the tests are being sorted out

README.md
```diff
@@ -35,9 +35,16 @@ git checkout feature/nodes
 
 ### 5. Open terminal 1 and run a node
 
+Make sure the you are using the same `PRIVATE_KEY` value as the CI for the decryption tests to work locally:
```
paulo-ocean, line 38: For someone outside OPF reading this, i think they might get the idea that this PRIVATE_KEY is a MUST.
But people can/should use whatever key they want, and most will not care about CI or tests at all.
I think we should rephrase this to make it more clear

src/components/storage/index.ts
```diff
@@ -60,6 +64,83 @@ export abstract class Storage {
     }
     return response
   }
+
+  async encrypt(encryptionType: 'AES' | 'ECIES' = 'AES') {
```
paulo-ocean, line 68: Not a big deal, just a suggestion, but IMO we should start using constants for these kind of strings..
for instance, right now, only for 'ECIES' we have it hardcoded in more than 20 places
Its a lot easier to have string constants in case we need to change something later. 
But anyway, not for this PR

src/@types/fileObject.ts
```diff
@@ -4,20 +4,23 @@ export interface HeadersObject {
   [key: string]: string
 }
 
-export interface UrlFileObject {
+export interface BaseFileObject {
   type: string
+  encryptedBy?: string
+  encryptMethod?: 'AES' | 'ECIES' | ''
```
bogdanfazakas, line 10: as also suggested by Paolo maybe worth using enums or constants instead of the strings and we can have them in types and use them in all the places

bogdanfazakas (commented):
looks good in general, some small remarks that are not mandatory but nice to have

src/components/storage/index.ts
jamiehewitt15: Yeah, good point

README.md
jamiehewitt15: Yeah, I agree. I'll remove this. I wrote this in the readme when I had a test present that relied on a certain private key to work but that test was proving to be unreliable so I have removed it for now

src/@types/fileObject.ts
jamiehewitt15: Yeah I agree, an enums or constant would be better

paulo-ocean: @jamiehewitt15 not a big issue, as we mentioned before, but i noticed you already created the enum

```
export enum EncryptMethod {
  AES = 'AES',
  ECIES = 'ECIES',
  NONE = ''
}
```
and you replaced it in a few places already, but since you started the job, why not go to the end and finish it?  :-)
I mean, there are a still a lot of hardcoded references to 'AES' and 'ECIES' (just search that on visual studio)
You could replace those ones (or most of them at least) ... while the subject is hot .
tkx

paulo-ocean (approved):
lgtm

bogdanfazakas (approved):

PR #268, opened 2024-02-15
40 command encryptfile

### LINKED ISSUE #40: Command:  encryptFile
So, we need to implement encrypt file command.

Command description:

{
  command: "encryptFile",
  node: "NodeB ID",  // optional, if missing, it refers to current node
  encryptionType:   [aes, ecies] // type of encryption required, default aes
  files: filesObject //  an object (NOT an ARRAY!) of type https://docs.oceanprotocol.com/developers/storage
 }

If successful, returns encrypted file content

an7e:
Fixes #40


src/@types/fileObject.ts
```diff
@@ -8,7 +8,7 @@ export interface UrlFileObject {
   type: string
   url: string
   method: string
-  headers?: [HeadersObject]
+  headers?: HeadersObject
```
alexcos20, line 11: Why only one?
We can have several headers for the same storage object

src/components/core/encryptHandler.ts
```diff
@@ -32,3 +35,56 @@ export class EncryptHandler extends Handler {
     }
   }
 }
+
+export class EncryptFileHandler extends Handler {
+  async handle(task: EncryptFileCommand): Promise<P2PCommandResponse> {
+    try {
+      if (task.files.type === 'url') {
```
alexcos20, line 42: please, we have Storage clases, use that.
Two weeks from now when we are going to add S3 storage, we have to search for hardcoded things..
And then do it again, four weeks later, when adding Azure Blob :)

alexcos20 (changes requested):

src/@types/fileObject.ts
an7e: @alexcos20 What do you mean by "several"?
like this ?
`headers: [ { "header1": "qwe" }, { "header2": "asd" } ]`
because HeadersObject is a key-value object, and usually headers are passed like this
`headers: { "header1": "qwe", "header2": "asd" }`

paulo-ocean: it was done

src/components/core/encryptHandler.ts
paulo-ocean: it was done

alexcos20 (approved):
lgtm

paulo-ocean (approved):

PR #269, opened 2024-02-16
Removed hardcoded checks from DDO validationon & fix DDO validation

### LINKED ISSUE #256: Short DDO validation & remove hardcoded validation
We are using the `validateObject` function to validate DDOs. However, it is currently doing this validation without checking what fields are required by the schema. So for example, the short DDO is unable to pass the validation as there are hardcoded checks for `@context` and `metadata`. The validation should only check for fields that are required by the schema. This could also cause problems for other DDO formats from other organisations, or ones added by us in the future.

```Typescript

export async function validateObject(
  obj: Record<string, any>,
  chainId: number,
  nftAddress: string
): Promise<[boolean, Record<string, string>]> {
  CORE_LOGGER.logMessage(`Validating object: ` + JSON.stringify(obj), true)
  const extraErrors: Record<string, string> = {}
  if (!('@context' in obj)) {
    extraErrors['@context'] = 'Context is missing.'
  }
  if ('@context' in obj && !Array.isArray(obj['@context'])) {
    extraErrors['@context'] = 'Context is not an array.'
  }
  if (!('metadata' in obj)) {
    extraErrors.metadata = 'Metadata is missing or invalid.'
  }
  ;['created', 'updated'].forEach((attr) => {
    if ('metadata' in obj && attr in obj.metadata && !isIsoFormat(obj.metadata[attr])) {
      extraErrors.metadata = `${attr} is not in ISO format.`
    }
  })

  if (!chainId) {
    extraErrors.chainId = 'chainId is missing or invalid.'
  }

...
```

mariacarmina:
Fixes #256  .

Changes proposed in this PR:

- 
- 
-


mariacarmina:
This can be closed.

PR #270, opened 2024-02-19
Initialize compute

### LINKED ISSUE #230: C2D:  initializeCompute

mariacarmina:
Fixes #230  .

Changes proposed in this PR:

- created initializeCompute 
- add test


src/components/core/compute.ts
```diff
@@ -53,3 +61,143 @@ export class GetEnvironmentsHandler extends Handler {
...
+      }
+    }
+    return false
+  }
+
+  getServiceById(ddo: DDO, serviceId: string) {
+    try {
+      return ddo.services.filter((service) => service.id === serviceId)[0]
+    } catch (err) {
+      CORE_LOGGER.logMessage(`Service was not found: ${err}`)
```
jamiehewitt15, line 87: It would be better to log an error rather than just a message: 

```suggestion
      CORE_LOGGER.error(`Service was not found: ${err}`)
```

src/components/core/utils/feesHandler.ts
```diff
@@ -170,34 +170,38 @@ export async function validateComputeProviderFee(
...
+            provider
+          )
+          return [false, newProviderFee]
+        } else {
+          return [true, decodedEventData.args]
+        }
       }
     }
+  } catch (err) {
+    CORE_LOGGER.logMessage(`Validation for compute provider fees failed due to: ${err}`)
```
jamiehewitt15, line 204: It would make sense to log an error here rather than a message `CORE_LOGGER.error()`

src/test/integration/computeFees.test.ts
```diff
@@ -164,12 +218,68 @@ describe('Compute provider fees', async () => {
     assert(trxReceipt, 'set metada failed')
   })
 
+  delay(1000)
```
jamiehewitt15, line 221: Paulo has a seperate PR removing all the delays so I would avoid adding another one here

src/test/integration/computeFees.test.ts
```diff
@@ -164,12 +218,68 @@ describe('Compute provider fees', async () => {
...
+          type: 'url',
+          url: 'https://raw.githubusercontent.com/tbertinmahieux/MSongsDB/master/Tasks_Demos/CoverSongs/shs_dataset_test.txt',
+          method: 'GET'
+        }
+      ]
+    }
+
+    const data = Uint8Array.from(Buffer.from(JSON.stringify(files)))
+    const encryptedData = await encrypt(data, 'ECIES')
+    // const encryptedDataString = encryptedData.toString('base64')
```
jamiehewitt15, line 238:
```suggestion
```

src/@types/commands.ts
```diff
@@ -49,6 +49,35 @@ export interface GetEnvironmentsCommand extends Command {
   chainId: number
 }
 
+export interface Dataset {
```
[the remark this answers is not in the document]
alexcos20: see https://github.com/oceanprotocol/ocean-node/pull/279/files#diff-c19a166d9cbd78d7cfe5a671f3a2d151ea75b829ed97b46f989bc9c4945f9235

src/components/core/compute.ts
```diff
@@ -53,3 +61,143 @@ export class GetEnvironmentsHandler extends Handler {
...
+        return {
+          stream: null,
+          status: {
+            httpStatus: 400,
+            error: errorMsg
+          }
+        }
+      }
+
+      const listOfAssest = [...task.datasets, ...[task.algorithm]]
```
alexcos20, line 134: this should be a separate function, because startCompute will use this as well to validate the request

alexcos20 (changes requested):

src/test/integration/computeFees.test.ts
mariacarmina: no more delays after merging with develop

mariacarmina (commented):
I also need to expose initializeCompute via http.

src/components/core/compute.ts
alexcos20: yes, we need to check in startCompute if fees were paid

mariacarmina: I have created a separate function but for the moment if the order is expired or there is no transaction Id for an order, I calculate another provider fees.

src/components/core/compute.ts
```diff
@@ -53,3 +61,143 @@ export class GetEnvironmentsHandler extends Handler {
...
+                  error: errorMsg
+                }
+              }
+            }
+          }
+          const service = this.getServiceById(ddo, asset.serviceId)
+          const networkUrl = (await getConfiguration()).supportedNetworks[
+            task.chainId.toString()
+          ].rpc
+          const provider = new JsonRpcProvider(networkUrl)
```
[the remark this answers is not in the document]
paulo-ocean: if we use the existing `Blockchain` class we already have it
`constructor(rpc: string, chaindId: number) {`
...
```
public getProvider(): JsonRpcApiProvider {
    return this.provider
  }
```

mariacarmina:
Still need to test via ocean-cli, but if you have other suggestions regarding the code, it will be open for review.

src/components/httpRoutes/compute.ts
```diff
@@ -44,3 +44,42 @@ computeRoutes.get('/api/services/computeEnvironments', async (req, res) => {
...
+      res.status(400).send('Missing algorithm did')
+      return
+    }
+    body.command = PROTOCOL_COMMANDS.INITIALIZE_COMPUTE
+    const result = await new InitializeComputeHandler(req.oceanNode).handle(body)
+    if (result.stream) {
+      const queryResult = JSON.parse(await streamToString(result.stream as Readable))
+      res.json(queryResult)
+    } else {
+      res.status(result.status.httpStatus).send(result.status.error)
```
bogdanfazakas, line 79: maybe worth adding a log here as well with the result error

alexcos20 (approved):

alexcos20 (approved):
lgtm

bogdanfazakas (approved):
besides some small remarks looks good

src/components/core/compute.ts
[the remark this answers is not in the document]
mariacarmina: I created an util function, not inside the class. Do we want to create a blockchain instance in order to create JsonRpcProvider? Is still valid suggestion? @paulo-ocean I can include if you want, no worries, but let me know if you are satisfied with the current changes

PR #271, opened 2024-02-19
Fix tests

mariacarmina:
Fixes # .

Changes proposed in this PR:

- 
- 
-

### DIFF
--- a/src/test/integration/computeFees.test.ts
+++ b/src/test/integration/computeFees.test.ts
@@ -125,7 +125,7 @@ describe('Compute provider fees', async () => {
       files: [
         {
           type: 'url',
-          url: 'https://github.com/datablist/sample-csv-files/raw/main/files/organizations/organizations-100.csv',
+          url: 'https://raw.githubusercontent.com/tbertinmahieux/MSongsDB/master/Tasks_Demos/CoverSongs/shs_dataset_test.txt',
           method: 'GET'
         }
       ]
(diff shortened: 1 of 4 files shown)


paulo-ocean (approved):
lgtm

src/test/integration/computeFees.test.ts
```diff
@@ -125,7 +125,7 @@ describe('Compute provider fees', async () => {
       files: [
         {
           type: 'url',
-          url: 'https://github.com/datablist/sample-csv-files/raw/main/files/organizations/organizations-100.csv',
+          url: 'https://raw.githubusercontent.com/tbertinmahieux/MSongsDB/master/Tasks_Demos/CoverSongs/shs_dataset_test.txt',
```
jamiehewitt15, line 128: why don't we use a file hosted in our own repo so that we have more control and can ensure that it doesn't change or get taken down?

mariacarmina: I saw that we used on provider and for other py repos, what file are you talking about exactly?

jamiehewitt15: Any file will work, you can put the same file that you're already using in `src/test/data` and then use the github link to it

paulo-ocean: i agree with @jamiehewitt15 we should use a file we control, but i also think that for now we should merge this, and maybe open another issue for changing the file, otherwise we will loose more time here

alexcos20 (approved):

jamiehewitt15 (approved):
Thanks Maria, this is a good quick fix for now

src/test/integration/computeFees.test.ts
jamiehewitt15: yeah agreed

PR #277, opened 2024-02-20
fix purgatory instance + additional checks

### LINKED ISSUE #275: Fix Purgatory instance and run from indexer
right now, the purgatory is called every time the indexer crawler runs:
 
```
const purgatory = await Purgatory.getInstance() 
const updatedDDO = await this.updatePurgatoryStateDdo(ddo, from, purgatory)
```
However the Purgatory.getInstance() might not even exist if the .env vars are not set

```
static async getInstance(): Promise<Purgatory> {
      if (!config.accountPurgatoryUrl && !config.assetPurgatoryUrl) {
        INDEXER_LOGGER.logMessage(`Purgatory URLs are not provided`)
        return
  }
```
So every time we call
`const updatedDDO = await this.updatePurgatoryStateDdo(ddo, from, purgatory)`
we pass 'undefined' reference.. and we get exceptions ahead
Apart from that...

There are some axio calls made on the Purgatory class that do not do any check (neither it is done previously) on the purgatory URLs... and these also throw exceptions if the .env vars are not set properly

More details on the comments of the PR
[https://github.com/oceanprotocol/ocean-node/pull/247](https://github.com/oceanprotocol/ocean-node/pull/247)

paulo-ocean:
Fixes #275 

Changes proposed in this PR:

- Some fixes on purgatory instance + additional checks


src/test/unit/utils.test.ts
```diff
@@ -1,5 +1,5 @@
 import { expect, assert } from 'chai'
-import { sleep, getEventFromTx } from '../../utils/util.js' // replace './yourModule' with the actual path of your module
+import { sleep, getEventFromTx, isValidUrl } from '../../utils/util.js' // replace './yourModule' with the actual path of your module
```
jamiehewitt15, line 2: Can you remove the pointless comment:
 
```suggestion
import { sleep, getEventFromTx, isValidUrl } from '../../utils/util.js'
```

paulo-ocean: strange.. that's already on develop branch, no idea where it comes from, seems automatic stuff

src/utils/util.ts
```diff
@@ -90,3 +90,14 @@ export async function readStream(stream: Stream): Promise<string> {
     })
   })
 }
+
+// basic url check using URL constructor
+export function isValidUrl(urlString: string): boolean {
```
jamiehewitt15, line 95: We might be better off using a library for this, for slightly more comprehensive validation. In the market we were using `is-url-superb`

paulo-ocean: i have to disagree on that one, why do we keep adding dependencies for simple stuff with 1 or 2 lines of code??
makes no sense IMO

bogdanfazakas: i've added `valid-url` library few weeks ago for a similar check, i think we should use in both places either the newly created method or use that library (it a bit too old but should do the work https://github.com/ogt/valid-url/blob/master/index.js) so we have a unified way of dealing with the url validations.
I'm also a fan of having small methods in our codebase than having a library as a dependancy, at that time i've use valid-url because it seemed to have a pretty complex way of checking the url string.

paulo-ocean: yeah, but that library is 11 years old :-)
things evolved a bit, and we don't need much  of that, a simple URL constructor is enough

jamiehewitt15: There are arguments in favor of either approach, the benefit of using a library is typically that it's been battle-tested by many other users over several years so we can be confident that we're not introducing any bugs. In this instance, it's a pretty simple method so I don't have a strong opinion either way. I agree with Bogdan though, we should avoid having both. If you're adding the `isValidUrl` method, can you remove the `valid-url` dependency?

paulo-ocean: ok, will remove that dependency then... the URL [https://developer.mozilla.org/en-US/docs/Web/API/URL/URL](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL) is enough these days

paulo-ocean: done

jamiehewitt15 (approved):
Looks good to me

PR #279, opened 2024-02-21
Feature/c2d start

alexcos20:
Fixes # .

Changes proposed in this PR:

- 
- 
-


paulo-ocean (approved):
overall, LGTM :+1: 
i couldn't test it in detail. I'm having some issues with barge and c2d locally (i think its something with my docker setup, as it worked before, and now seems i've messed up with it somehow),  but tests are green & code wise looks good too, so lets move on :-)

PS: I think you're missing some pulls from develop
tkx

jamiehewitt15:
As @mariacarmina pointed out, we're only running the `completeFlow.test.js` test on the CI at the moment. Can we merge this change into this PR to get all the tests to run: https://github.com/oceanprotocol/ocean-node/pull/282

mariacarmina:
Discussed with @alexcos20, we will merge first #230 then #279

src/components/core/compute/initialize.ts
```diff
@@ -0,0 +1,211 @@
...
+        index = index + 1
+      }
+      if (!foundValidCompute) {
+        // edge case, where all assets have valid orders and valid provider fees (for download)
+        // unfortunatelly, none have valid compute provider fees.  let's create for the first asset that is published on a chainId that matches our env
+        // just take any asset and create provider fees with compute
+        console.log('TO DO!!!!')
+      }
+      return {
+        stream: Readable.from(JSON.stringify(allFees)),
```
mariacarmina, line 195: Here shouldn't it be `result` instead of `allFees`? I mean you want to return this type of response:

```json
{
    "datatoken": "0x21fa3ea32892091...",
    "nonce": 23,
    "providerFee": {
        "providerFeeAddress": "0xabc123...",
        "providerFeeToken": "0xabc123...",
        "providerFeeAmount": "200",
        "providerData": "0xabc123...",
        "v": 27,
        "r": "0xabc123...",
        "s": "0xabc123...",
        "validUntil": 123456,
    }
    "computeAddress": "0x8123jdf8sdsa..."
}
```
It is from [API.md](https://github.com/oceanprotocol/provider/blob/main/API.md#initialise-a-compute-service).
When I first implemented I also put only datasets and algorithm fees but the bresponse format was different so I changed to match API.md from provider.

src/components/core/utils/feesHandler.ts
```diff
@@ -155,53 +134,97 @@ export async function calculateComputeProviderFee(
...
-    CORE_LOGGER.logMessage(`Validation for compute provider fees failed due to: ${err}`)
   }
+  if (errorMsg) {
+    CORE_LOGGER.logMessage(errorMsg)
+    retMsg.message = errorMsg
+  }
+  return retMsg
 }
+
+// TO DO - delete functions below, as they are used in the tests
```
mariacarmina, line 225: Should be tackled in another issue?

mariacarmina (commented):
Another comment: TODO comments will be tackled in this PR or separate issues?

src/components/core/compute/initialize.ts
alexcos20: we agregate results for algo & all datasets in allFees var, and then we return that

src/components/core/utils/feesHandler.ts
alexcos20: yes, I will create a new one

PR #281, opened 2024-02-23
Issue 265 encrypt endpoint

### LINKED ISSUE #265: HTTP Endpoint:  encryptFile

paulo-ocean:
Fixes #265 

Changes proposed in this PR:

- Add encryptFile endpoint (supports fileObject and raw data on body)
- update express.raw() to support larger files (default is only 100kb)
- Some refactor on endpoints, add some constants for routes
- adapt EncryptFileHandler to support Buffer
- return proper headers


bogdanfazakas (approved):
lgtm

mariacarmina (approved):
lgtm

PR #284, opened 2024-02-27
New ESLint rule: Require await

### LINKED ISSUE #283: Remove unnecessary async keywords
Remove unnecessary async keywords. Currently, we have 70+ occasions where async is used without an `await`, most of these are not necessary.

jamiehewitt15:
Fixes #283 

Changes proposed in this PR:

- Adding New ESLint rule: Require await to highlight when async keyword could be removed
- Removing unnecessary async keywords
- Adding ESLint ignore when it is not appropriate.


src/components/Indexer/crawlerThread.ts
```diff
@@ -132,7 +132,7 @@ async function processReindex(): Promise<void> {
   }
 }
 
-export async function checkNewlyIndexedAssets(events: BlocksEvents): Promise<void> {
+export function checkNewlyIndexedAssets(events: BlocksEvents): void {
```
paulo-ocean, line 135: I'm not 100% sure about this ESlint rule. Personally i would be more in favour of us doing the right thing when creating the functions than relaying on these automated rules..
For instance, here's one example:
This function is :
`export function checkNewlyIndexedAssets(events: BlocksEvents): void {`
(not async or returning any promise)
but it is still called like this:
`await checkNewlyIndexedAssets(processedBlocks.foundEvents)`
Which is not exactly an error, but doesn't make sense to have `await` expressions on stuff that do not return Promises..
So, it looks to me the ESLint only does half the job, and in particular cases like this can be even more confusing IMO

src/components/Indexer/index.ts
```diff
@@ -109,7 +109,7 @@ export class OceanIndexer {
     }
   }
 
-  static async addReindexTask(reindexTask: ReindexTask): Promise<void> {
```
paulo-ocean, line 112: Same thing here for instance.. We got rid of the Promise (OK)
But we still call this with `await` sintax

src/components/Indexer/crawlerThread.ts
jamiehewitt15: Yeah, you're right, it's doing half the job. It's only helping reduce the unnecessary "async" keywords. That's still useful though, as it's much easier for developers to spot an unnecessary `await` if the unnecessary `async` has already been removed and less likely that more unnecessary `await` will be added. I can also look to see if there is another rule that might help spot unnecessary `await` keywords

paulo-ocean: no worries, not a big deal

paulo-ocean (approved):
lgtm, added couple comments but minor stuff

mariacarmina (approved):
lgtm

bogdanfazakas (approved):
Lgtm

PR #301, opened 2024-02-29
Remove logs

### LINKED ISSUE #291: Logs retention
- [ ] Define an env variable which sets how much time in history do we keep the logs
- [ ] have an internal cron job that deletes all logs older then necessary

Without this, we are going to fill the db pretty quickly

jamiehewitt15:
Fixes #291

Changes proposed in this PR:

- 
- 
-


src/components/database/index.ts
```diff
@@ -706,6 +709,60 @@ export class LogDatabase {
...
+        GENERIC_EMOJIS.EMOJI_CROSS_MARK,
+        LOG_LEVELS_STR.LEVEL_ERROR
+      )
+      throw error
+    }
+  }
+
+  async deleteOldLogs() {
+    const currentTime = new Date().getTime()
+    const xTime = parseInt(process.env.LOG_RETENTION_TIME || '2592000000') // Default to 30 days in milliseconds
```
alexcos20, line 738: instead of using process.env , can you pls use src/utils/constants.ts ?

alexcos20 (changes requested):

paulo-ocean (commented):
@jamiehewitt15 
I think maybe its worth to have a look at winston log rotation capabilities, there are projects like:
[https://www.npmjs.com/package/winston-daily-rotate-file](url)
and others.

maybe we can have this automatically integrated in winston transports in a more smoother way ( But i'm not sure how / if this applies to custom transports like DB or just files)

Also, don't forget to update ENVIRONMENT_VARIABLES on constants.ts

thanks

src/components/database/index.ts
```diff
@@ -681,6 +682,9 @@ export class LogDatabase {
       if (level) {
         filterConditions += ` && level:${level}`
       }
+      if (maxLogs > 250) {
+        maxLogs = 250
```
bogdanfazakas, line 686: would it be helpful to have a log message here to inform the users that the logs number was capped to 250 ?

src/components/database/index.ts
```diff
@@ -706,6 +710,60 @@ export class LogDatabase {
...
+        GENERIC_EMOJIS.EMOJI_CROSS_MARK,
+        LOG_LEVELS_STR.LEVEL_ERROR
+      )
+      throw error
+    }
+  }
+
+  async deleteOldLogs() {
+    const currentTime = new Date().getTime()
+    const xTime = parseInt(ENVIRONMENT_VARIABLES.LOG_RETENTION_TIME.value || '2592000000') // Default to 30 days in milliseconds
```
bogdanfazakas, line 739: maybe worth storing the default 2592000000 value in a const

src/index.ts
```diff
@@ -163,4 +164,17 @@ if (config.hasHttp) {
   app.listen(config.httpPort, () => {
     OCEAN_NODE_LOGGER.logMessage(`HTTP port: ${config.httpPort}`, true)
   })
+
+  // Schedule the cron job to run daily at midnight
+  cron.schedule('0 0 * * *', async () => {
```
bogdanfazakas, line 169: think would be cleaner to have a helper that sets this cron and have just one call in the index.ts file

src/components/database/index.ts
```diff
@@ -681,6 +682,9 @@ export class LogDatabase {
       if (level) {
         filterConditions += ` && level:${level}`
       }
+      if (maxLogs > 250) {
```
alexcos20, line 685: can we move '250' to a env as well ?

alexcos20 (changes requested):

src/components/database/index.ts
jamiehewitt15: We could but it's just going to lead to users getting an error. If they want more logs they will change the env above 250 and get the `Only up to 250 hits can be fetched per page.` error. If someone wants fewer logs they can already use the `maxLogs` parameter when they make a request.

If we need to request more than 250 logs then we should introduce pagination.

jamiehewitt15: I have created a separate issue for introducing pagination as applies to all our collections, not just the logs database: https://github.com/oceanprotocol/ocean-node/issues/330

bogdanfazakas (approved):
lgtm

src/components/database/index.ts
jamiehewitt15: @alexcos20 can you have another look at this PR? As agreed in the meeting, we will introduce pagination rather than having an env here for `maxLogs`

mariacarmina (approved):
lgtm

PR #318, opened 2024-03-04
Issue 316 address file optional

### LINKED ISSUE #316: Bug:   Node needs addresses.json file for remote deployments
Node looks for ADDRESS_FILE env  and tries to open $HOME/.ocean/.... address.json

But this should be optional.
If there is no such env, use addresses.json with is imported by ocean-contracts npm dep

paulo-ocean:
Fixes #316 

Changes proposed in this PR:

- don't just read local file, use ADDRESS_FILE (if set) or imported addresses from npm package 
- refactor indexer calls to getAddressFile, use existing utils function


alexcos20 (approved):

bogdanfazakas (approved):
lgtm

Issue #288, opened 2024-02-28
BE:  Authentication function


alexcos20:
Some actions on the dashboard will required admin privileges.
Let's build the base for that

- [ ] research how to do the signature.   Proposal:
      - Users signs a message that looks like   String(nonce + '-' + expiryTimestamp)
      - For every request, users adds:
           - nonce
           - expiryTimestamp
           - signature
 - [ ] define env var:  ALLOWED_ADMINS,  it can be a string or array of strings
 - [ ] implement function to check auth for admins

alexcos20:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/S1---DApps-63f8b8c26e838c0017f34d23/issues/oceanprotocol/ocean-node/288?planning-poker) with Zenhub @bogdanfazakas @jamiehewitt15 @mariacarmina @paulo-ocean

jamiehewitt15:
Couple of things I'll add on this: 

1. The HTTP endpoint only needs to respond `true` or `false` as I will send another signature from the front end when we call one of the endpoints that requires user authentication. Those other endpoints can then reverify the signature.
2. For the implementation we can follow [the approach that we took in the uploader ](https://github.com/oceanprotocol/uploader_filecoin/blob/74d08f75dd6146c8c66539adb18469def3611103/src/api/upload/upload.service.js#L20)and update it to use ethers 6 and typescrpt.
3. We don't need to check a nonce if we are including an `expiryTimestamp` we just need to verify that the signature has come from an admin account and that it hasn't expired. 

So the implementation would be like this: 

```
const message = expiryTimestamp
const signerAddress = utils.verifyMessage(message, signature).toLowerCase();

if (signerAddress === adminAddress && currentTimestamp < expiryTimestamp){
return true
}
```

mariacarmina:
Thank you for clarifying @jamiehewitt15. Btw how do you prefer to send the `nonce` and `expiryTimestamp` as types? numbers or strings?

jamiehewitt15:
The `expiryTimestamp` would be sent as a number in unix timestamp format. And then if we agree the nonce isn't needed, I won't send it.

jamiehewitt15:
Had a chat with @alexcos20 about this and we agreed on the following: 

1. No need for an HTTP endpoint for authenticating the user. We will just have an HTTP endpoint that returns the public addresses of the admins. 
[...]

PR #319, opened 2024-03-05
Auth admin route

### LINKED ISSUE #296: Authenticate admin
Some actions on the dashboard will required admin privileges.
Let's build the base for that in the UI.   Depends on #288

TO DO:

- [ ] prepare signature function. Message = String(nonce + '-' + expiryTimestamp)
- [ ] store expiryTimestamp,nonce and signature to browser localdb
- [ ] send expiryTimestamp,nonce and signature whenever is needed for auth calls

TBD: expiryTimestamp should be hardcoded (ie:  valid for next 24h) or allow user to define?

mariacarmina:
Fixes #296 .

Changes proposed in this PR:

- add admin route for auth
- add test
- added handler for testing


jamiehewitt15:
Either in this PR or another one, we also need a simple endpoint that will return the array of admin addresses.

src/test/integration/dashboard.test.ts
```diff
@@ -0,0 +1,104 @@
...
+  tearDownEnvironment
+} from '../utils/utils.js'
+import {
+  DEVELOPMENT_CHAIN_ID,
+  getOceanArtifactsAdresses,
+  getOceanArtifactsAdressesByChainId
+} from '../../utils/address.js'
+import axios, { AxiosResponse } from 'axios'
+
+describe('Should run a complete node flow.', async () => {
```
jamiehewitt15, line 22:
```suggestion
describe('Should run the authentication flow.', async () => {
```

src/components/httpRoutes/dashboard/utils/utils.ts
```diff
@@ -0,0 +1,35 @@
...
+export function validateSignature(
+  nonce: number,
+  expiryTimestamp: number,
+  signature: string
+): boolean {
+  try {
+    const message = sha256(
+      toUtf8Bytes(nonce.toString() + '-' + expiryTimestamp.toString())
+    )
+    HTTP_LOGGER.logMessage(`Message for signature validation: ${message}`)
```
jamiehewitt15, line 14: I don't think there is much value in logging this

src/components/httpRoutes/dashboard/utils/utils.ts
```diff
@@ -0,0 +1,35 @@
+import { sha256, toUtf8Bytes, verifyMessage } from 'ethers'
+import { HTTP_LOGGER } from '../../../../utils/logging/common'
+import { existsEnvironmentVariable } from '../../../../utils/config.js'
+
+export function validateSignature(
```
jamiehewitt15, line 5: This is a very generalized function that can also used by P2P commands if they need authentication. I think it would be better placed in the `src/utils` folder rather than `src/components/httpRoutes/dashboard/utils`

src/components/httpRoutes/dashboard/utils/utils.ts
```diff
@@ -0,0 +1,35 @@
...
+  nonce: number,
```
jamiehewitt15, line 6:
```suggestion
```
no need for a nonce

src/components/httpRoutes/dashboard/utils/utils.ts
```diff
@@ -0,0 +1,35 @@
...
+  expiryTimestamp: number,
+  signature: string
+): boolean {
+  try {
+    const message = sha256(
+      toUtf8Bytes(nonce.toString() + '-' + expiryTimestamp.toString())
```
jamiehewitt15, line 12:
```suggestion
      toUtf8Bytes(expiryTimestamp.toString())
```

.github/workflows/ci.yml
```diff
@@ -164,6 +164,7 @@ jobs:
           FEE_AMOUNT: '{ "amount": 1, "unit": "MB" }'
           ASSET_PURGATORY_URL: 'https://raw.githubusercontent.com/oceanprotocol/list-purgatory/main/list-assets.json'
           ACCOUNT_PURGATORY_URL: 'https://raw.githubusercontent.com/oceanprotocol/list-purgatory/main/list-accounts.json'
+          ALLOWED_ADMINS: '[]'
```
paulo-ocean, line 167: If this is not required, it should not have a default value (a missing value could be interpreted as an empty array as well). Also we should have the same setting on `test/.env.test` (in this case, if no default, then should not be set anywhere)
tkx

src/components/httpRoutes/dashboard/admin.ts
```diff
@@ -0,0 +1,52 @@
+import express from 'express'
+import { HTTP_LOGGER } from '../../../utils/logging/common.js'
+import { validateSignature } from './utils/utils.js'
+import { LOG_LEVELS_STR } from '../../../utils/logging/Logger.js'
+
+export const aquariusRoutes = express.Router()
```
paulo-ocean, line 6: this is not another `aquariusRoutes` :-) this name is clashing with the real aquariousRoutes on `aquarius.ts`
besides, it should be imported on `httpRoutes/index.ts` otherwise its not used at all

paulo-ocean (changes requested):

src/test/integration/dashboard.test.ts
```diff
@@ -0,0 +1,104 @@
...
+  const provider = new JsonRpcProvider('http://127.0.0.1:8545')
+
+  const publisherAccount = (await provider.getSigner(0)) as Signer
+  const consumerAccount = (await provider.getSigner(1)) as Signer
+  const consumerAddress = await consumerAccount.getAddress()
+  const publisherAddress = await publisherAccount.getAddress()
+
+  before(async () => {
+    // override and save configuration (always before calling getConfig())
+    console.log(`envs: ${JSON.stringify(ENVIRONMENT_VARIABLES)}`)
```
paulo-ocean, line 37: better clean the console.log()
besides, do we really need to override all those ENV variables on setupEnvironment? I think only ALLOWED_ADMINS would be necessary here.
I also don't understand the purpose of this part:
```
//  eslint-disable-next-line no-unused-vars
    const indexer = new OceanIndexer(database, mockSupportedNetworks)

    let network = getOceanArtifactsAdressesByChainId(DEVELOPMENT_CHAIN_ID)
    if (!network) {
      network = getOceanArtifactsAdresses().development
    }
```
Why do we need the database, indexer, and the 'network' check?

Overall i think we should instead have a 'unit' test and not an 'integration' test, to check the signature validation with the allowed admins.. not the axios HTTP call for the endpoint
thanks

src/components/httpRoutes/dashboard/utils/utils.ts
```diff
@@ -0,0 +1,35 @@
...
+    )
+    HTTP_LOGGER.logMessage(`Message for signature validation: ${message}`)
+
+    const signerAddress = verifyMessage(message, signature).toLowerCase()
+    HTTP_LOGGER.logMessage(`Resolved signer address: ${signerAddress}`)
+
+    if (!existsEnvironmentVariable('ALLOWED_ADMINS')) {
```
paulo-ocean, line 19: that's wrong usage of the function, should be:
`!existsEnvironmentVariable(ENVIRONMENT_VARIABLES.ALLOWED_ADMINS)`

paulo-ocean (changes requested):

src/components/httpRoutes/dashboard/admin.ts
jamiehewitt15: Yeah true, although the easier option would just be to remove HTTP endpoint altogether. In standup yesterday we agreed the dashboard wouldn't use an HTTP endpoint for authentication. We just need a simple endpoint that returns the array of public addresses for admin users. The `validateSignature` function will be used by the other endpoints (yet to be added) that require validation. 

The updated flow is outlined [here](https://github.com/oceanprotocol/ocean-node/issues/288#issuecomment-1978948960)

src/components/httpRoutes/dashboard/utils/utils.ts
```diff
@@ -0,0 +1,35 @@
...
+      HTTP_LOGGER.logMessage(`Missing env var for ALLOWED_ADMINS`)
+    }
+    const currentTimestamp = new Date().getTime()
+    for (const address of JSON.parse(process.env.ALLOWED_ADMINS)) {
```
paulo-ocean, line 23: Overall, i think its better to read this on getConfiguration() and make this part of the config, 
check if the variable exists: 
`existsEnvironmentVariable(ENVIRONMENT_VARIABLES.ALLOWED_ADMINS)`
if not or there are errors parsing it, set it as an empty array, 
if it exists, parse the array and make sure we have valid addresses and not something else..
expose it on config
tkx

mariacarmina:
> Either in this PR or another one, we also need a simple endpoint that will return the array of admin addresses.

I'll do it in this PR.

src/components/httpRoutes/dashboard/utils/utils.ts
mariacarmina: I'll use this information for admin list endpoint that @jamiehewitt15 reffered to. Thank you!

src/components/httpRoutes/dashboard/admin.ts
mariacarmina: Created only validateSignature and `adminList` endpoint.

mariacarmina:
Postman response for `directCommand` calling status handler:
Screenshot 2024-03-07 at 15 59 50

src/components/httpRoutes/dashboard/admin.ts
```diff
@@ -0,0 +1,50 @@
+import express from 'express'
+import { HTTP_LOGGER } from '../../../utils/logging/common.js'
+import { LOG_LEVELS_STR } from '../../../utils/logging/Logger.js'
+import { getConfiguration } from '../../../utils/index.js'
+
+export const adminRoutes = express.Router()
+const regex: RegExp = /^(0x)?[0-9a-fA-F]{40}$/
+
+adminRoutes.get(`/adminList`, async (req, res) => {
```
jamiehewitt15, line 9: Seeing as we're now sending the array of admin addresses in the status response, we don't need this endpoint

mariacarmina: Missed that spot, deleted now. Thanks

src/test/integration/download.test.ts
```diff
@@ -51,6 +51,7 @@ describe('Should run a complete node flow.', () => {
   let assetDID: string
   let publishedDataset: any
   let actualDDO: any
+  // const publisherAddress = await publisherAccount.getAddress()
```
jamiehewitt15, line 54:
```suggestion
```

src/components/core/utils/statusHandler.ts
```diff
@@ -27,6 +27,23 @@ export async function status(
   }
   const config = await getConfiguration()
   const { indexer: indexerDatabase } = oceanNode.getDatabase()
+
+  const validAddresses = []
+  if (config.allowedAdmins) {
+    const regex: RegExp = /^(0x)?[0-9a-fA-F]{40}$/
+    for (const admin of JSON.parse(config.allowedAdmins)) {
+      if (regex.test(admin) === true) {
```
bogdanfazakas, line 35: do we need to have our own regular expression and test the address against it? 
is there a reason we are not using `isAddress()` from ethers utils ?

src/@types/commands.ts
```diff
@@ -51,6 +51,8 @@ export interface ValidateDDOCommand extends Command {
   ddo: DDO
 }
 
+export interface GetAdminListCommand extends Command {}
```
bogdanfazakas, line 54: do we still need this ?

src/components/core/utils/statusHandler.ts
mariacarmina: I checked the docs for ethers and I have seen that it accepts whatever format as a string. No need to have that regex expr, I'll use `isAddress`. Thanks!

src/@types/commands.ts
mariacarmina: Nope, I'll delete it

jamiehewitt15 (approved):
I've tested locally and it's working well

bogdanfazakas (approved):
Lgtm

src/utils/auth.ts
```diff
@@ -0,0 +1,27 @@
...
+    const message = sha256(toUtf8Bytes(expiryTimestamp.toString()))
+
+    const signerAddress = verifyMessage(message, signature).toLowerCase()
+    HTTP_LOGGER.logMessage(`Resolved signer address: ${signerAddress}`)
+
+    if (!existsEnvironmentVariable('ALLOWED_ADMINS')) {
+      HTTP_LOGGER.logMessage(`Missing env var for ALLOWED_ADMINS`)
+    }
+    const currentTimestamp = new Date().getTime()
+    for (const address of JSON.parse(process.env.ALLOWED_ADMINS)) {
```
mariacarmina, line 16: I should also add here status handler call instead of getting the envs.

src/utils/config.ts
```diff
@@ -421,7 +421,8 @@ async function getEnvConfig(isStartup?: boolean): Promise<OceanNodeConfig> {
     feeStrategy: getOceanNodeFees(supportedNetworks, isStartup),
     c2dClusters: getC2DClusterEnvironment(isStartup),
     accountPurgatoryUrl: getEnvValue(process.env.ACCOUNT_PURGATORY_URL, ''),
-    assetPurgatoryUrl: getEnvValue(process.env.ASSET_PURGATORY_URL, '')
+    assetPurgatoryUrl: getEnvValue(process.env.ASSET_PURGATORY_URL, ''),
+    allowedAdmins: getEnvValue(process.env.ALLOWED_ADMINS, [])
```
paulo-ocean, line 425: This is not OK @mariacarmina .
`getEnvValue()` is parsing Strings..
If you set a valid .env var like for instance:
`export ALLOWED_ADMINS=["0xe2DD09d719Da89e5a3D0F2549c7E24566e947260"]`
The node crashes when getting the command Status (from status handler)
Because you're trying to parse an array but the value was initially read as a plain String
tkx

src/components/core/utils/statusHandler.ts
```diff
@@ -27,6 +28,22 @@ export async function status(
   }
   const config = await getConfiguration()
   const { indexer: indexerDatabase } = oceanNode.getDatabase()
+
+  const validAddresses = []
+  if (config.allowedAdmins) {
+    for (const admin of JSON.parse(config.allowedAdmins)) {
```
paulo-ocean, line 34: This line crashes the node, because the .env var was initially read as String, instead of Array of strings..
Besides, all the parsing & validation (including address check)  should be done on the config, when we first read the setting..
Not here... here its too late, and its doing the same checks every time it gets called, while it should already be fine on config, and here is just has to read it from there. Even if if it's an empty array, no problem.. You just need to add whatever is there to the response
tkx

paulo-ocean: Anyway, lets close this as is.
I created another issue: [https://github.com/oceanprotocol/ocean-node/issues/328](url)
and will address the refactor as well
tkx

paulo-ocean (approved):

src/utils/config.ts
mariacarmina: Yes, I had some troubles with that, but in the end I convert within status handler and the list of addresses appears fine. How would you prefer? To parse in the config or to keep parsing logic inside status handler?

mariacarmina: I have added for config.

mariacarmina:
Fixed the comments and tested again on postman, seems to work fine:

Screenshot 2024-03-11 at 13 24 30
I'll merge it, thanks for reviews

PR #324, opened 2024-03-07
Issue 306 command validation

### LINKED ISSUE #306: Security:   Command input parameter validation
Right now, we are doing validation of command's input parameters in http endpoints section.

But, a P2P command sent by a malicious node is handled  by the handlers, without any type of validation

Task:
 - move all validation from http endpoints section to handlers section.  Each handler should have it's own validation
 - http endpoints:  try to pass null if a params is missing, handlers should handle that, don't do any kind of validation, as it's redundant

paulo-ocean:
Fixes #306 

Changes proposed in this PR:

- Adds validation at handler level, all handlers do validation prior to handle commands
- returns appropriate responses
- refactor of some existing param checks methods

### DIFF
--- a/src/components/core/compute/initialize.ts
+++ b/src/components/core/compute/initialize.ts
@@ -3,7 +3,7 @@ import { P2PCommandResponse } from '../../../@types/index.js'
 import { CORE_LOGGER } from '../../../utils/logging/common.js'
 import { Handler } from '../handler.js'
 import { DDO } from '../../../@types/DDO/DDO.js'
-import { ComputeInitializeCommand } from '../../../@types/commands.js'
+import { Command, ComputeInitializeCommand } from '../../../@types/commands.js'
 import { ProviderComputeInitializeResults } from '../../../@types/Fees.js'
 import { AssetUtils } from '../../../utils/asset.js'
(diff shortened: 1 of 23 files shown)


alexcos20 (approved):

src/components/core/compute/initialize.ts
```diff
@@ -12,37 +12,50 @@ import { validateOrderTransaction } from '../utils/validateOrders.js'
...
+          `Invalid compute environment: ${command.compute.env}`
+        )
       }
+    }
+
+    return validation
+  }
+
+  async handle(task: ComputeInitializeCommand): Promise<P2PCommandResponse> {
+    const validation = this.validate(task)
```
mariacarmina, line 53: Can we keep logs when accesing a new handler after the validation?

paulo-ocean: We have the logs of what every handler/command receives, its part of `validateCommandParameters`, so it doesn't make sense to have them duplicated... the other logs are still there, nothing changed

jamiehewitt15 (approved):
Looks good to me

bogdanfazakas (approved):

PR #329, opened 2024-03-11
Stop node handler

### LINKED ISSUE #292: BE: Allow admin to stop node
Expose an endpoint for admin that will stop the node  (process.exit), after doing necessary cleanups.

When node is started with something like pm2  or docker using restart:always, this will behave like a forced restart.

Of course, admin needs to auth itself

mariacarmina:
Fixes #292  .

Changes proposed in this PR:

- 
- 
-

### DIFF
--- a/src/components/core/dashboard.ts
+++ b/src/components/core/dashboard.ts
@@ -1,0 +1,20 @@
+import { Handler } from './handler.js'
+import { P2PCommandResponse } from '../../@types/OceanNode.js'
+import { StopNodeCommand } from '../../@types/commands.js'
+import { CORE_LOGGER } from '../../utils/logging/common.js'
+import { ReadableString } from '../P2P/handleProtocolCommands.js'
+
+export class StopNodeHandler extends Handler {
+  handle(task: StopNodeCommand): Promise<P2PCommandResponse> {
+    CORE_LOGGER.logMessage(`Stopping node execution...`)
+    setTimeout(() => {
+      process.exit()
+    }, 2000)
+    return new Promise<P2PCommandResponse>((resolve, reject) => {
+      resolve({
+        status: { httpStatus: 200 },
+        stream: new ReadableString('EXIT OK')
+      })
+    })
+  }
+}
--- a/src/components/httpRoutes/validateCommands.ts
+++ b/src/components/httpRoutes/validateCommands.ts
@@ -1,6 +1,8 @@
 import { isAddress } from 'ethers'
 import { SUPPORTED_PROTOCOL_COMMANDS, PROTOCOL_COMMANDS } from '../../utils/constants.js'
 import { EncryptMethod } from '../../@types/fileObject.js'
+import { validateSignature } from '../../utils/auth.js'
+import { OceanNode } from '../../OceanNode.js'
 
 export type ValidateParams = {
   valid: boolean
@@ -8,14 +10,20 @@ export type ValidateParams = {
   status?: number
 }
 
-export function validateBroadcastParameters(requestBody: any): ValidateParams {
+export function validateBroadcastParameters(
+  requestBody: any,
+  oceanNode: OceanNode
+): ValidateParams {
   // for now we can use the same validation function,
   // but later we might need to have separate validation functions
   // if we many different commands of each type
-  return validateCommandAPIParameters(requestBody)
+  return validateCommandAPIParameters(requestBody, oceanNode)
 }
 // add others when we add support
-export function validateCommandAPIParameters(requestBody: any): ValidateParams {
+export function validateCommandAPIParameters(
+  requestBody: any,
+  oceanNode: OceanNode
+): ValidateParams {
   // eslint-disable-next-line prefer-destructuring
   const command: string = requestBody.command as string
 
@@ -145,6 +153,17 @@ export function validateCommandAPIParameters(requestBody: any): ValidateParams {
         return buildInvalidRequestMessage(
           'Missing required parameter(s): "chaindId","datasets", "algorithm","compute", "consumerAddress"'
         )
+      }
+    } else if (command === PROTOCOL_COMMANDS.STOP_NODE) {
+      if (!requestBody.expiryTimestamp || !requestBody.signature) {
+        return buildInvalidRequestMessage(
+          'Missing required parameter(s): "expiryTimestamp","signature"'
+        )
+      }
+      if (
+        !validateSignature(requestBody.expiryTimestamp, requestBody.signature, oceanNode)
+      ) {
+        return buildInvalidRequestMessage('Expired authentication or invalid signature')
       }
     }
     // only once is enough :-)
--- a/src/@types/OceanNode.ts
+++ b/src/@types/OceanNode.ts
@@ -48,6 +48,7 @@ export interface OceanNodeConfig {
   accountPurgatoryUrl: string
   assetPurgatoryUrl: string
   codeHash?: string
+  allowedAdmins?: string
 }
 
 export interface P2PStatusResponse {
@@ -90,6 +91,7 @@ export interface OceanNodeStatus {
   platform: any
   uptime?: number // seconds since start
   codeHash?: string
+  allowedAdmins?: string[]
 }
 
 export interface P2PBroadcastResponse {
--- a/src/@types/commands.ts
+++ b/src/@types/commands.ts
@@ -53,6 +53,11 @@ export interface ValidateDDOCommand extends Command {
 
 export interface StatusCommand extends Command {}
 
+export interface StopNodeCommand extends Command {
+  expiryTimestamp: number
+  signature: string
+}
+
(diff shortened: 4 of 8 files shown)


src/components/core/dashboard.ts
```diff
@@ -0,0 +1,20 @@
...
+import { StopNodeCommand } from '../../@types/commands.js'
+import { CORE_LOGGER } from '../../utils/logging/common.js'
+import { ReadableString } from '../P2P/handleProtocolCommands.js'
+
+export class StopNodeHandler extends Handler {
+  handle(task: StopNodeCommand): Promise<P2PCommandResponse> {
+    CORE_LOGGER.logMessage(`Stopping node execution...`)
+    setTimeout(() => {
+      process.exit()
+    }, 2000)
```
mariacarmina, line 12: Should we have this `2000` as env var or we hardcode it?

mariacarmina (commented):
cc @alexcos20

src/components/core/dashboard.ts
bogdanfazakas: I don't think we need an env var just for this

src/components/core/dashboard.ts
```diff
@@ -0,0 +1,20 @@
+import { Handler } from './handler.js'
+import { P2PCommandResponse } from '../../@types/OceanNode.js'
+import { StopNodeCommand } from '../../@types/commands.js'
+import { CORE_LOGGER } from '../../utils/logging/common.js'
+import { ReadableString } from '../P2P/handleProtocolCommands.js'
+
+export class StopNodeHandler extends Handler {
```
jamiehewitt15, line 7: I would recommend avoiding naming this file `/dashboard.ts` as it doesn't give any indication about what this handler does.

paulo-ocean: agree, maybe this would make more sense on some "admin" file or something? or  "adminOperations"? i don't know, just some thoughts, but i agree that is is more node administration than dashboard (we might want to stop the node by sending the command in some other form)

src/components/httpRoutes/validateCommands.ts
```diff
@@ -1,21 +1,29 @@
...
-export function validateBroadcastParameters(requestBody: any): ValidateParams {
+export function validateBroadcastParameters(
+  requestBody: any,
+  oceanNode: OceanNode
+): ValidateParams {
   // for now we can use the same validation function,
   // but later we might need to have separate validation functions
   // if we many different commands of each type
-  return validateCommandAPIParameters(requestBody)
+  return validateCommandAPIParameters(requestBody, oceanNode)
```
paulo-ocean, line 20: overall, looks good , it stops and sends the response back as expected.. 
but we should merge and fix conflicts... validation is now part of the handlers 
So we should use the `validate() `method on the handler, as we do  for others  (check develop branch)
thanks

mariacarmina: And in `validate` method I'll tackle the signature as well.

src/components/core/adminOperations.ts
```diff
@@ -0,0 +1,51 @@
...
+    if (
+      !validateSignature(command.expiryTimestamp, command.signature, this.getOceanNode())
+    ) {
+      return buildInvalidRequestMessage('Expired authentication or invalid signature')
+    }
+  }
+
+  handle(task: StopNodeCommand): Promise<P2PCommandResponse> {
+    const validation = this.validate(task)
+    if (!validation.valid) {
```
paulo-ocean, line 35: this breaks the response when the validation passes: "Cannot read properties of undefined (reading 'valid')"
Reason:
On the validate() method above you're only returning a response if case the validation fails..  so you're missing a return at the end, otherwise, on this line its undefined :-)
`const validation = this.validate(task)`
thanks

mariacarmina: I missed that spot, I'll add it back, thank you!

src/components/core/adminOperations.ts
```diff
@@ -0,0 +1,51 @@
...
+      'signature'
+    ])
+    if (!commandValidation.valid) {
+      CORE_LOGGER.logMessage(
+        `Command validation failed: ${JSON.stringify(commandValidation)}`
+      )
+      return commandValidation
+    }
+    if (
+      !validateSignature(command.expiryTimestamp, command.signature, this.getOceanNode())
```
paulo-ocean, line 27: there is something else Not Ok here, we can still stop the node if the `validateSignature` fails.

mariacarmina: > we can still stop the node if the validateSignature fails.

you mean simple `return` statement? From what I understood, if authentication fails you cannot do anything.
_"Of course, admin needs to auth itself"_

paulo-ocean: i mean, AS IS..
Even when the authentication fails (because there was an issue on validateSignature) , the exit() part is still being executed... So anyone issuing the command can stop the node (with any wrong signature)
Futhermore the `validateSignature` function, that was already merged on another PR, is wrong..
It makes use of StatusHander for nothing
tkx

mariacarmina: > i don't understand the purpose of the StatusHandler usage. Its not doing anything

I retrieve allowed admins from status handler, because we agreed to expose on status command. `validateSignature` returns a Promise and in the validation code, it needed to be an `await` as well

paulo-ocean: @mariacarmina expose the allowedAdmins property on the Status Command is not the same of using the StatusHandler to validate a signature :-)

mariacarmina: The `validateSignature` is not wrong, we've previously use it and it worked fine on other microservices.

mariacarmina: > expose the allowedAdmins property on the Status Command is not the same of using the StatusHandler to validate a signature :-)

I wanted to retrieve the allowed admins addresses, if you have other option, please propose it and I'm happy to implement, but we need to find solutions.

paulo-ocean: getConfiguration() ?

mariacarmina: A solution can be to retrieve the env vars and use them in the signature code to avaoid async code. What do you think @paulo-ocean?

mariacarmina: getConfiguration is async

mariacarmina: getAllowedAdmins() should be a good fit

paulo-ocean: what's the problem of being async? It can stay async, for reading the configuration or you can pass the configuration as param (instead of the node for iinstance).. There are several ways of doing it.

When calling it you could also use other syntax like `.then()` , it doesn't necessary need to use await sintax

The  `handle()` method is usually async on other handlers as well, maybe you can just validate the params on the validate() and the signature at the beginning of the handle() as well (outside the validate but after the params check)

mariacarmina: it solved it faster than having `.then()` blocks and 1000 paranthesis.

> maybe you can just validate the params on the validate() and the signature at the beginning of the handle() as well (outside the validate but after the params check)

That's also valid right, but since we have a dedicated function for validation, I thought that it fits better in the validation, the idea was to check the signature before accessing the handler IMO.

paulo-ocean: just checked, and i think you can call getAllowedAdmins() , which is exported function..
So, you get rid of all the async code... in this case you can keep the signature check inside validate()
All issues with async will be gone

mariacarmina: That's what I proposed and committed as well, yes, I thought that it can make all problems with async code dissapear for auth function :))

paulo-ocean: yes, its better to keep on validate() if we can... and without async calls, we can :-)
but if there was not other easy option, its valid as well, as long as we do it :-)

src/components/httpRoutes/validateCommands.ts
```diff
@@ -9,12 +10,16 @@ export type ValidateParams = {
   status?: number
 }
 
-export function validateBroadcastParameters(requestBody: any): ValidateParams {
+export function validateBroadcastParameters(
+  requestBody: any,
+  oceanNode: OceanNode
```
paulo-ocean, line 15: we can remove this unused param

tsconfig.json
```diff
@@ -23,6 +23,6 @@
   "ts-node": {
     "esm": true
   },
-  "include": ["src/**/*"],
+  "include": ["src/**/*", "src/components/core/adminOperations.ts"],
```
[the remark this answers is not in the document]
mariacarmina: to be honest, when I fixed conflicts and renamed the file to adminOperations that appeared, I'll delete it cause it looks strange.

jamiehewitt15 (approved):
Looks good to me

paulo-ocean (approved):
lgtm

PR #332, opened 2024-03-12
Reindex tx handler.

### LINKED ISSUE #313: BE: Allow admin to force reindex some tx
The admin needs to be able to force node to reindex a tx.

Let's expose that endpoint, which is only allowed for admins

Params: 
 - chain id
 - txId

mariacarmina:
Fixes #313  .

Changes proposed in this PR:

- refactor of admin handlers structure => new admin folder and handlers in separate files
- created reindex tx handler + reindex chain handler (merged from #336 )
- added tests


jamiehewitt15:
Ideally it would be nice to see a test written for this, is there any reason that you didn't add one?

src/@types/commands.ts
```diff
@@ -106,6 +106,13 @@ export interface GetFeesCommand extends Command {
   validUntil?: number // this allows a user to request a fee that is valid only for a limited period of time, less than service.timeout
 }
 
+export interface ReindexTxCommand extends Command {
```
paulo-ocean, line 109: if the admin commands all have `expiryTimestamp` & `signature` in common, probably is better to extend a common AdminCommand or something.
For instance, we already have:
```
export interface StopNodeCommand extends Command {
  expiryTimestamp: number
  signature: string
}
```
and

```
export interface ReindexTxCommand extends Command {
  expiryTimestamp: number
  signature: string
  chainId: number
  txId: string
}
```
We already have 2 fields in common on both commands and we're extending the base interface instead

thanks

src/components/core/adminOperations.ts
```diff
@@ -49,3 +52,72 @@ export class StopNodeHandler extends Handler {
...
+  validate(command: ReindexTxCommand): ValidateParams {
+    const commandValidation = validateCommandParameters(command, [
+      'expiryTimestamp',
+      'signature',
+      'chainId',
+      'txId'
+    ])
+    if (!commandValidation.valid) {
+      const errorMsg = `Command validation failed: ${JSON.stringify(commandValidation)}`
+      CORE_LOGGER.logMessage(errorMsg)
```
paulo-ocean, line 66: i'm not sure we need all this logging.. the command input/message is already completely logged upper in validateCommandParameters, and we're returning a response with all the details in case something fails.
In any case, if its an error, we should use `CORE_LOGGER.error(message)`, otherwise we're just logging with the default level (Info), which also looks different (and looks a good message by the colours)
tkx

src/components/core/adminOperations.ts
```diff
@@ -49,3 +52,72 @@ export class StopNodeHandler extends Handler {
...
+      return buildInvalidRequestMessage(errorMsg)
+    }
+    return commandValidation
+  }
+
+  async handle(task: ReindexTxCommand): Promise<P2PCommandResponse> {
+    const validation = this.validate(task)
+    if (!validation.valid) {
+      return new Promise<P2PCommandResponse>((resolve, reject) => {
```
paulo-ocean, line 80: there's no need for the Promise<> sintax... we already have async on the function signature. Its one or the other...

src/components/core/adminOperations.ts
```diff
@@ -49,3 +52,72 @@ export class StopNodeHandler extends Handler {
...
+      if (!ret) {
+        CORE_LOGGER.log(
+          LOG_LEVELS_STR.LEVEL_ERROR,
+          `Reindex tx for txId ${task.txId} failed on chain ${task.chainId}.`,
+          true
+        )
+        return
+      }
+
+      return new Promise<P2PCommandResponse>((resolve, reject) => {
```
paulo-ocean, line 113: same as above

src/components/core/adminOperations.ts
```diff
@@ -49,3 +52,72 @@ export class StopNodeHandler extends Handler {
...
+    const validation = this.validate(task)
+    if (!validation.valid) {
+      return new Promise<P2PCommandResponse>((resolve, reject) => {
+        resolve(buildInvalidParametersResponse(validation))
+      })
+    }
+    CORE_LOGGER.logMessage(`Reindexing tx...`)
+    const config = await getConfiguration()
+    const blockchain = new Blockchain(
+      config.supportedNetworks[task.chainId.toString()].rpc,
```
paulo-ocean, line 87: what happens if i send a `chainId` that is not supported?

src/@types/commands.ts
mariacarmina: I agree

src/components/core/adminOperations.ts
mariacarmina: > we already have async on the function signature.

We removed that from stop node command PR

src/components/core/adminOperations.ts
mariacarmina: we need also a check before creating Blockchain instance creation.

src/components/core/adminOperations.ts
mariacarmina: for this particular I'll remove this log to avoid the noise, we already return validation error within handler.

mariacarmina: For errors, I'll change to `CORE_LOGGER.error(message)`

src/components/core/adminOperations.ts
```diff
@@ -10,27 +11,27 @@ import {
...
+      return buildInvalidRequestMessage('Expired authentication or invalid signature')
     }
     return commandValidation
   }
 
-  handle(task: StopNodeCommand): Promise<P2PCommandResponse> {
+  handle(task: AdminStopNodeCommand): Promise<P2PCommandResponse> {
     const validation = this.validate(task)
     if (!validation.valid) {
       return new Promise<P2PCommandResponse>((resolve, reject) => {
```
paulo-ocean, line 37: its much cleaner if we simply mark the handle function as "async"  as we have on other handlers..
we don't need the Promise resolve/reject syntax, which always add more noise
tkx

jamiehewitt15:
As I suggest on [your other PR](https://github.com/oceanprotocol/ocean-node/pull/336#discussion_r1533758032) I think we can do a refactoring to prevent this `src/components/core/adminOperations.ts` file getting too big.

src/components/core/adminOperations.ts
paulo-ocean: not sure if i understand, but we still have:
```
return new Promise<P2PCommandResponse>((resolve, reject) => {
        resolve(buildInvalidParametersResponse(validation))
      })
```
and
```
return new Promise<P2PCommandResponse>((resolve, reject) => {
        resolve({
          status: { httpStatus: 200 },
          stream: new ReadableString('REINDEX TX OK')
        })
on `ReindexTxHandler`
      })
```
and there's no need for that syntax

mariacarmina:
> As I suggest on [your other PR](https://github.com/oceanprotocol/ocean-node/pull/336#discussion_r1533758032) I think we can do a refactoring to prevent this `src/components/core/adminOperations.ts` file getting too big.

It is integrated if you want to take a look

src/components/core/admin/adminHandler.ts
```diff
@@ -0,0 +1,34 @@
...
+    }
+    if (!validateSignature(command.expiryTimestamp, command.signature)) {
+      return buildInvalidRequestMessage('Expired authentication or invalid signature')
+    }
+    return {
+      valid: true
+    }
+  }
+
+  abstract handle(task: Command): Promise<P2PCommandResponse>
```
paulo-ocean, line 29: these are not needed @mariacarmina :
```
abstract handle(task: Command): Promise<P2PCommandResponse>

  getOceanNode(): OceanNode {
    return super.getOceanNode()
  }
}
```
You can delete this piece... its still an abstract class that derives from Handler, we just need to re-implement `validate()`, the rest we don't need to redo
thanks

src/components/core/admin/reindexTxHandler.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  async handle(task: AdminReindexTxCommand): Promise<P2PCommandResponse> {
+    const validation = this.validate(task)
+    if (!validation.valid) {
+      return buildInvalidParametersResponse(validation)
+    }
+    CORE_LOGGER.logMessage(`Reindexing tx...`)
+    const checkChainId = await checkSupportedChainId(task.chainId)
+    if (!checkChainId.validation) {
+      CORE_LOGGER.error(`Chain ID ${task.chainId} is not supported in config.`)
```
paulo-ocean, line 34: I think we can skip this log... the `checkSupportedChainId` its already logging the same message, so its duplicated.. and maybe its worth to have a look at other logs as well, and see if we can clean something... we are logging too much stuff in general...
by the way, one other thing, all the commands should return something like: `Promise<P2PCommandResponse> ` so we shoul not simply do a "return", but always return something like:
```
return {
        status: xxx
        stream: xxx
      }
```
(maybe have a default structure to respond in case of an error, to avoid duplicating much code)...
Same on the `catch` block..
```
} catch (error) {
      CORE_LOGGER.log(LOG_LEVELS_STR.LEVEL_ERROR, `REINDEX tx: ${error.message} `, true)
    }
```
There is no answer here
tkx

mariacarmina: Sure, integrated.

mariacarmina:
For ReindexTxHandler, I used addReindexTask method to post message to worker thread and add the reindex in indexing queue -> ended up with CI infinite loop for contract deployment, neither installation of package.json worked at all. I kept `processLogs` for now, should I create another static method in indexer which reindexes the tx in that moment for admin?
cc @bogdanfazakas

src/test/utils/assets.ts
```diff
@@ -96,6 +96,7 @@ export async function publishAsset(genericAsset: any, publisherAccount: Signer)
   genericAsset.services[0].datatokenAddress = datatokenAddress
 
   const stringDDO = JSON.stringify(genericAsset)
+  console.log(`string DDO: ${stringDDO}`)
```
bogdanfazakas, line 99: missed this :)

src/test/integration/download.test.ts
```diff
@@ -166,11 +166,12 @@ describe('Should run a complete node flow.', () => {
...
     const fileInfoTask = {
       command: PROTOCOL_COMMANDS.FILE_INFO,
-      did: publishedDataset.ddo.id,
-      serviceId: publishedDataset.ddo.services[0].id
+      did: actualDDO.id,
+      serviceId: actualDDO.services[0].id
     }
 
     const response = await new FileInfoHandler(oceanNode).handle(fileInfoTask)
+    console.log(`response file info: ${JSON.stringify(response)}`)
```
bogdanfazakas, line 174: same :)

src/test/integration/operationsDashboard.test.ts
```diff
@@ -0,0 +1,197 @@
...
+    assert(handlerResponse, 'handler resp does not exist')
+    assert(handlerResponse.status.httpStatus === 200, 'incorrect http status')
+
+    assert(
+      (await dbconn.ddo.retrieve(publishedDataset.ddo.id)) === null,
+      'ddo does not exist'
+    )
+    const searchParameters = {
+      q: '*',
+      quey_by: `chainId:${DEVELOPMENT_CHAIN_ID}`
```
bogdanfazakas, line 178: shouldn't this look something like this
`{
  "q": "chainId:{chain_id}",
  "query_by": "chainId"
}`

src/components/database/index.ts
```diff
@@ -379,6 +382,21 @@ export class DdoDatabase {
       )
     }
   }
+
+  async deleteAllAssetsFromChain(chainId: number) {
+    const searchParameters = {
+      q: '*',
+      quey_by: `chainId:${chainId}`
+    }
```
bogdanfazakas, line 390: should we use something like this for the query? 
{
  "q": "chainId:{chain_id}",
  "query_by": "chainId"
}

src/components/database/index.ts
```diff
@@ -354,6 +354,9 @@ export class DdoDatabase {
           .delete(did)
         if (response.id === did) {
           isDeleted = true
+          DATABASE_LOGGER.logMessage(
+            `response for deleting the ddo: ${response}, isDeleted: ${isDeleted}`
```
bogdanfazakas, line 358: let's reduce a bit the log pollution and for eg here keep just 
```suggestion
            `response for deleting the ddo: ${response.id }, isDeleted: ${isDeleted}`
```
and not log entire objects, that kind of log would work imo only for logs with `DEBUG` level

src/components/core/fileInfoHandler.ts
```diff
@@ -128,6 +137,7 @@ export class FileInfoHandler extends Handler {
         })
       } else if (task.did && task.serviceId) {
         const fileArray = await getFile(task.did, task.serviceId, oceanNode)
+        CORE_LOGGER.logMessage(`fileArray: ${fileArray}`)
```
bogdanfazakas, line 140: consider deleting this

src/components/Indexer/crawlerThread.ts
```diff
@@ -170,4 +172,8 @@ parentPort.on('message', (message) => {
       REINDEX_QUEUE.push(message.reindexTask)
     }
   }
+  if (message.method === 'reset-crawling') {
+    const deployedBlock = getDeployedContractBlock(message.chainId)
+    updateLastIndexedBlockNumber(deployedBlock)
```
bogdanfazakas, line 177: as it is now, we can still end up with the race condition issue we debated on.

The approach should be something like the following:
`
    const deployedBlock = getDeployedContractBlock(message.chainId)
    // set a global variable that you should check later on 
    after  checkNewlyIndexedAssets(processedBlocks.foundEvents) inside proccesNetworkData's while loop
    lastIndexedBlock = (your-env-var) ? your-env-var  : processedBlocks.lastBlock 
`

src/components/Indexer/index.ts
```diff
@@ -128,6 +128,13 @@ export class OceanIndexer {
     }
   }
 
+  static resetCrawling(chainId: number): void {
+    const worker = OceanIndexer.workers[chainId]
+    if (worker) {
+      worker.postMessage({ method: 'reset-crawling', chainId })
```
bogdanfazakas, line 134: no need to specify chainId since the worker has that data already from the creation

bogdanfazakas (commented):
Please also do some log cleanups, did not added comments to all specific locations where we can reduce a bit or keep as low as possible the `INFO` level logs

src/components/Indexer/crawlerThread.ts
mariacarmina: created the global var and set/unset when reset-crawling message is posted to the thread otherwise continue indexing normally.

src/components/database/index.ts
mariacarmina: modified with "filter_by" according to [docs ](https://typesense.org/docs/0.24.0/api/search.html#query-parameters)

src/components/Indexer/index.ts
mariacarmina: ok, but then in the bind parentPort.on(), to get the start block from contracts with specific chaniId from message -> should I use chainId from supported network from ThreadData?

src/components/database/index.ts
mariacarmina: I wanted to delete it but, will keep it on debug level

paulo-ocean: i agree with @bogdanfazakas , its hard to keep track of all the changes (they are many :-) )
but generally speaking it think you are logging too much stuff @mariacarmina . It can get pretty heavy really quit (keep in mind that these will be going to a database (apart from files) in production env. 
We already log all the input received by the commands, including any validation errors.. i think is enough to log any errors that might happen. We don't need to log "good" responses or log any action we are taking...
tkx

src/components/core/admin/reindexChainHandler.ts
```diff
@@ -0,0 +1,54 @@
...
+} from '../../httpRoutes/validateCommands.js'
+import { P2PCommandResponse } from '../../../@types/OceanNode.js'
+import { CORE_LOGGER } from '../../../utils/logging/common.js'
+import { checkSupportedChainId } from '../../../utils/blockchain.js'
+import { ReadableString } from '../../P2P/handleProtocolCommands.js'
+import { OceanIndexer } from '../../Indexer/index.js'
+
+export class ReindexChainHandler extends AdminHandler {
+  validate(command: AdminReindexChainCommand): ValidateParams {
+    if (!validateCommandParameters(command, ['chainId', 'txId'])) {
```
paulo-ocean, line 18: Hi @mariacarmina 
I don´t think we need 'txId' on this command, `AdminReindexChainCommand` only takes 'chainId',
If you try to `validate()` this command according to the interface specified AdminReindexChainCommand, 
validation will fail without 'txId' wich should not be part of if
tkx

src/components/core/admin/reindexTxHandler.ts
```diff
@@ -0,0 +1,52 @@
...
+  ValidateParams
+} from '../../httpRoutes/validateCommands.js'
+import { AdminReindexTxCommand } from '../../../@types/commands.js'
+import { P2PCommandResponse } from '../../../@types/OceanNode.js'
+import { CORE_LOGGER } from '../../../utils/logging/common.js'
+import { ReadableString } from '../../P2P/handleProtocolCommands.js'
+import { checkSupportedChainId } from '../../../utils/blockchain.js'
+import { OceanIndexer } from '../../Indexer/index.js'
+
+export class ReindexTxHandler extends AdminHandler {
```
paulo-ocean, line 16: I see we have some common logic on `ReindexTxHandler` and `ReindexHandler` ...
These 2 handers have exactly the same purpose, only difference is that one requires admin permissions and other don't ... we probably need to refactor in the future (not for this PR)... probably we don't want to keep the one without  the "admin" checks (the older one)
tkx

src/test/unit/commands.test.ts
```diff
@@ -209,19 +207,6 @@ describe('Commands and handlers', () => {
     }
     expect(echoHandler.validate(echoCommand).valid).to.be.equal(true)
     // -----------------------------------------
-    // ReindexHandler
-    const reindexHandler: ReindexHandler = CoreHandlersRegistry.getInstance(
```
paulo-ocean, line 213: Don't forget to also add the validations for the other 2 handlers here, since we removed this one
and to delete the corresponding references to PROTOCOL_COMMANDS.REINDEX
tkx

src/components/core/admin/reindexTxHandler.ts
mariacarmina: removed the one without auth

src/test/unit/commands.test.ts
mariacarmina: yes, ofc, thanks

mariacarmina: In order to not include barge dep (create signer, generate signature) for unit test => I've just created the handlers and make sure that they exist, I already test validation on `operationsDashboard.test.ts` file, so we are covered.

mariacarmina:
created https://github.com/oceanprotocol/ocean-node/issues/387 for ref

src/components/database/index.ts
mariacarmina: Already implemented in the latest commits.

src/components/database/index.ts
mariacarmina: > think you are logging too much stuff @mariacarmina .

Yes for debug I'll do everything :). I am not planning to leave them in the develop branch, that log was for debugging only. Read my comment above please :)

src/components/core/adminOperations.ts
mariacarmina: this is outdated, it was changed.

src/components/database/index.ts
mariacarmina:
```
message: '[DATABASE] => ❌ [DATABASE] => Error when searching by query {"q":8996,"query_by":"chainId"}: Field `chainId` should be a string or a string array.'

```

mariacarmina:
Reindex chain command works on Optimisim Sepolia and you can see from the logs:
New last indexed block : 3870603', and after it settles to deployed block from contracts 3867703
``` console
{
  component: 'INDEXER',
  moduleName: 'INDEXER',
  level: 'info',
  message: '[INDEXER] => [INDEXER] => New last indexed block : 3870603',
  timestamp: '2024-04-15 14:13:00'
}
{
  component: 'CORE',
  moduleName: 'CORE',
  level: 'info',
  message: '[CORE] => Checking received command data for Command "reindexChain": {\n' +
    '    "signature": "0x09f28ce29c6ce6c42d3595df5727ade2df0b5ca40d103dbd534443942d010323243f8d9d6168f2888c7ffc3a41798e2619dbfe83040b77bafc7739df136ee6151c",\n' +
    '    "command": "reindexChain",\n' +
    '    "chainId": 11155420,\n' +
    '    "expiryTimestamp": 1744664400000\n' +
    '}',
  timestamp: '2024-04-15 14:13:03'
}
{
  component: 'HTTP',
  moduleName: 'HTTP',
  level: 'info',
  message: '[HTTP] => [HTTP] => Sending command : {"signature":"0x09f28ce29c6ce6c42d3595df5727ade2df0b5ca40d103dbd534443942d010323243f8d9d6168f2888c7ffc3a41798e2619dbfe83040b77bafc7739df136ee6151c","command":"reindexChain","chainId":11155420,"expiryTimestamp":1744664400000}',
  timestamp: '2024-04-15 14:13:03'
}
{
  component: 'OCEANNODE',
  moduleName: 'OCEANNODE',
  level: 'info',
  message: '[OCEANNODE] => [OCEANNODE] => Incoming direct command for ocean peer',
  timestamp: '2024-04-15 14:13:03'
}
{
  component: 'OCEANNODE',
  moduleName: 'OCEANNODE',
  level: 'info',
  message: '[OCEANNODE] => [OCEANNODE] => Performing task: {"signature":"0x09f28ce29c6ce6c42d3595df5727ade2df0b5ca40d103dbd534443942d010323243f8d9d6168f2888c7ffc3a41798e2619dbfe83040b77bafc7739df136ee6151c","command":"reindexChain","chainId":11155420,"expiryTimestamp":1744664400000}',
  timestamp: '2024-04-15 14:13:03'
}
{
  component: 'CORE',
  moduleName: 'CORE',
  level: 'info',
  message: '[CORE] => Checking received command data for Command "reindexChain": {\n' +
    '    "signature": "0x09f28ce29c6ce6c42d3595df5727ade2df0b5ca40d103dbd534443942d010323243f8d9d6168f2888c7ffc3a41798e2619dbfe83040b77bafc7739df136ee6151c",\n' +
    '    "command": "reindexChain",\n' +
    '    "chainId": 11155420,\n' +
    '    "expiryTimestamp": 1744664400000\n' +
    '}',
  timestamp: '2024-04-15 14:13:03'
}
{
  component: 'CORE',
  moduleName: 'CORE',
  level: 'info',
  message: '[CORE] => Checking received command data for Command "reindexChain": {\n' +
    '    "signature": "0x09f28ce29c6ce6c42d3595df5727ade2df0b5ca40d103dbd534443942d010323243f8d9d6168f2888c7ffc3a41798e2619dbfe83040b77bafc7739df136ee6151c",\n' +
    '    "command": "reindexChain",\n' +
    '    "chainId": 11155420,\n' +
    '    "expiryTimestamp": 1744664400000\n' +
    '}',
  timestamp: '2024-04-15 14:13:03'
}
{
  component: 'HTTP',
  moduleName: 'HTTP',
  level: 'info',
  message: 'Resolved signer address: 0xf96da2e8b85482f8ed870fd8dbb2374fc74d0779',
  timestamp: '2024-04-15 14:13:03'
}
{
  component: 'CORE',
  moduleName: 'CORE',
  level: 'info',
  message: 'Reindexing chain command called',
  timestamp: '2024-04-15 14:13:03'
}
{
  component: 'CORE',
  moduleName: 'CORE',
  level: 'info',
  message: 'Assets from chain 11155420 were deleted from db, now starting to reindex...',
  timestamp: '2024-04-15 14:13:03'
}
{
  component: 'INDEXER',
  moduleName: 'INDEXER',
  level: 'info',
  message: 'found worker: [object Object]',
  timestamp: '2024-04-15 14:13:03'
}
{
  component: 'INDEXER',
  moduleName: 'INDEXER',
  level: 'info',
  message: 'REINDEX BLOCK: 3867703',
  timestamp: '2024-04-15 14:13:03'
}
```

src/utils/auth.ts
```diff
@@ -1,24 +1,36 @@
-import { sha256, toUtf8Bytes, verifyMessage } from 'ethers'
+import { ethers } from 'ethers'
 import { HTTP_LOGGER } from './logging/common.js'
 import { getAllowedAdmins } from './index.js'
 
-export function validateSignature(expiryTimestamp: number, signature: string): boolean {
+export function validateSignature(
```
paulo-ocean, line 5: Hi @mariacarmina 
couple of comments here. I think it might be a good idea to use common types. we already have many different return types for validation define, all pretty similar.. for instance the return type of this function could be (instead of an array) something like a `CommonValidation` object which is defined as
```
export interface CommonValidation {
  valid: boolean
  error?: string
}
```
(maybe we could move the interface definition to another file than the current one)..

Apart from that i would suggest maybe change the function name so we know this is related **only** with admin signatures (we have a check for allowed admins), and is not to be used for other signature checks... maybe something like `validateAdminSignature` would be more meaningful . And for the logs, i think CORE_LOGGER would be a better choice since we're not logging HTTP requests here
thanks

src/components/core/handler/fileInfoHandler.ts
```diff
@@ -40,15 +40,24 @@ async function getFile(
       throw new Error(msg)
     }
     // 3. Decrypt the url
+    CORE_LOGGER.logMessage(
```
paulo-ocean, line 43: Hi @mariacarmina 
for sure we don't want to log and disclose all this info, please remove these logs, this seems to be debug stuff

src/components/core/admin/adminHandler.ts
```diff
@@ -0,0 +1,30 @@
...
+export abstract class AdminHandler extends Handler {
+  validate(command: AdminCommand): ValidateParams {
+    const commandValidation = validateCommandParameters(command, [
+      'expiryTimestamp',
+      'signature'
+    ])
+    if (!commandValidation.valid) {
+      return buildInvalidRequestMessage(commandValidation.reason)
+    }
+    if (!validateSignature(command.expiryTimestamp, command.signature)[0]) {
```
paulo-ocean, line 19: Hi @mariacarmina ,
There is no need to call `validateSignature` twice here, it doesn't make much sense. Just call it once and get the result object (maybe `CommonValidation` type object). Then check the fields,if not valid call `buildInvalidRequestMessage` with the error, otherwise just return true
tkx

src/components/Indexer/crawlerThread.ts
```diff
@@ -131,14 +132,19 @@ export async function proccesNetworkData(): Promise<void> {
...
+          } else {
+            await updateLastIndexedBlockNumber(processedBlocks.lastBlock)
+          }
           chunkSize = chunkSize !== 1 ? chunkSize : rpcDetails.chunkSize
         } catch (error) {
-          INDEXER_LOGGER.log(
-            LOG_LEVELS_STR.LEVEL_ERROR,
-            `Processing event from network failed network: ${rpcDetails.network} Error: ${error.message} `,
-            true
+          INDEXER_LOGGER.error(`network: ${rpcDetails.network} Error: ${error.message} `)
```
paulo-ocean, line 144: Hi @mariacarmina 
why do we need this change here? the chunk size reduction was only on the upper try/catch, i don't quite understand why we're doing it again here.. Besides, we are logging too much again. Imo, we should keep the catch block as it was before (maybe improve the message)... but just log 1 error message
thanks

src/components/Indexer/crawlerThread.ts
```diff
@@ -213,4 +220,7 @@ parentPort.on('message', (message) => {
       REINDEX_QUEUE.push(message.reindexTask)
     }
   }
+  if (message.method === 'reset-crawling') {
+    REINDEX_BLOCK = getDeployedContractBlock(message.chainId)
```
paulo-ocean, line 224: i think it might be a good idea to `lockProccessing` before doing this change and enable it right after. (the lockProccessing variable would need to be at a higgher scope).
Maybe not urgent or critical, but i think it would could some unexpected behaviours
thanks

src/utils/auth.ts
mariacarmina: Ok, I'll create the return type, change name function and use CORE LOGGER. At first implementation was in httpRoutes and I kept HTTP LOGGER as well :)

src/components/core/handler/fileInfoHandler.ts
mariacarmina: sure, it was for debugging, deleting it now.

src/components/Indexer/crawlerThread.ts
```diff
@@ -131,8 +133,13 @@ export async function proccesNetworkData(): Promise<void> {
             startBlock,
             blocksToProcess
           )
-          await updateLastIndexedBlockNumber(processedBlocks.lastBlock)
           checkNewlyIndexedAssets(processedBlocks.foundEvents)
+          if (REINDEX_BLOCK) {
```
paulo-ocean, line 137: hi @mariacarmina you have to do this update `updateLastIndexedBlockNumber(REINDEX_BLOCK)` when you receive the 'reset-crawling' message.. otherwise this code might never get executed... That is the problem for instance on local network..
At the beginning of the while() we have these checks:
```
const indexedBlock = await getLastIndexedBlock()
const networkHeight = await getNetworkHeight(provider)
```
As an Example: if the last indexed block is 42 and network height is 42 (meaning we are up to date)..
The entire block after:
```
if (networkHeight > startBlock) {
...
```
is never executed...  which means pretty nothing will change for this network
So, we probably need to update the block here: 
```
if (message.method === 'reset-crawling') {
    lockProccessing = true
    REINDEX_BLOCK = getDeployedContractBlock(message.chainId)
    updateLastIndexedBlockNumber(REINDEX_BLOCK).then(lockProccessing = false)`
```
The same error would happen for other networks... but as they are more 'busy', the network height is basically always greater than the last indexed block , so it will execute the block in most cases
tkx

mariacarmina: I can confirm that @paulo-ocean's solution works on ganache _at this moment_. Can we refactor the processNetworkData() in order to see if REINDEX_BLOCK is set and then update the block number at the beginning of the while() where we have those checks

```javascript
const indexedBlock = await getLastIndexedBlock()
const networkHeight = await getNetworkHeight(provider)
```

 and after we can unset REINDEX_BLOCK var?

paulo-ocean: > if (message.method === 'reset-crawling') {
>     lockProccessing = true
>     REINDEX_BLOCK = getDeployedContractBlock(message.chainId)
>     updateLastIndexedBlockNumber(REINDEX_BLOCK).then(lockProccessing = false)`

yes, we're locking the process:
```
if (message.method === 'reset-crawling') {
    **lockProccessing = true**
    REINDEX_BLOCK = getDeployedContractBlock(message.chainId)
    updateLastIndexedBlockNumber(REINDEX_BLOCK).then(lockProccessing = false)`
```
it doesn't matter "where" in the loop we are... the current loop will finish normally.. then the loop is blocked, and only unblocked after the change of the block number

paulo-ocean: > REINDEX_BLOCK

we don't even need `REINDEX_BLOCK` outside
 `if (message.method === 'reset-crawling') {`
 in fact, we don't need this variable at all
 just reset the lastIndexedBlock to getDeployedContractBlock when you get the message

bogdanfazakas: yep sorry missed that line in the comment, i was talking about the issue you also mentioned bellow, regargind race conditions that should be also handled 
> there can be some race condition, so when we are updating the block number inside the loop, maybe we should only do it, if the loop is not blocked in the meantime... might need some tweaks here indeed

paulo-ocean: no worries, thread is getting long :-D hard to follow for all..
I think there were basically 2 ways of handling this task..
1 ) Have a mechanism to stop the thread execution normally, delete it, update the block number, then just start a new worker thread with the new data..
2 ) change the thread data while running

so 1) would be a lot easier IMO, but we went for option 2), no big problem with that, but its trickier ..

so, going with 2)  IMO, we need to:
a) stop processing (locking it)
b) update the block number
c) restart the processing

I would do it all on the message handler and not complicate even more the while() loop..
Afaik, the problematic part is when we call `updateLastIndexedBlockNumber` inside the loop...
Cause even if we blocked processing on the message handler.. its possible that we are in the middle of the loop somewhere already, so the variable new value is not read in time... and the call to `updateLastIndexedBlockNumber(processedBlocks.lastBlock)` inside the loop could mess/override the call to `updateLastIndexedBlockNumber(reindexBlock)` done from the reindex message

So, i would probably do something like (just an example/suggestion)
```
if (message.method === 'reset-crawling') {
    **lockProccessing = true**
    const reindexBlock = getDeployedContractBlock(message.chainId)
    updateLastIndexedBlockNumber(reindexBlock).then(lockProccessing = false)
```
And inside the loop make sure we do something like:
```
if (!lockProccessing) {
  await updateLastIndexedBlockNumber(processedBlocks.lastBlock)
```
(so never update if we are 'locked', which could have been triggered by the reindex message in the meantime)

mariacarmina: basically both methods are valid. Either my method with global var and check if that exists then update last block before while() loop -> adds functionality in processNetworkData(), either keeping .then() functionality in parentPort binding as @paulo-ocean suggested and no extra global will be involved. Mention that in both cases you need lockProcessing set to true.

mariacarmina: ok this thread is too long and the comments are delayed in response. all good.

src/components/Indexer/crawlerThread.ts
```diff
@@ -16,12 +16,15 @@ import { INDEXER_LOGGER } from '../../utils/logging/common.js'
 import { getDatabase } from '../../utils/database.js'
 import { Log } from 'ethers'
 
+let lockProccessing: boolean = null
```
paulo-ocean, line 19: hi @mariacarmina  we don´t need this variable here,
it should be inside processNetworkData() as it was before:
```
...
let { chunkSize } = rpcDetails
let lockProccessing = false
...
```

src/components/Indexer/crawlerThread.ts
```diff
@@ -34,7 +37,7 @@ const blockchain = new Blockchain(rpcDetails.rpc, rpcDetails.chainId)
 const provider = blockchain.getProvider()
 const signer = blockchain.getSigner()
 
-async function updateLastIndexedBlockNumber(block: number): Promise<number> {
+export async function updateLastIndexedBlockNumber(block: number): Promise<number> {
```
paulo-ocean, line 40: we should update this function to return 'null' if we catch an error (like we did for `getLastIndexedBlock() `)

src/components/Indexer/crawlerThread.ts
```diff
@@ -80,7 +89,7 @@ export async function proccesNetworkData(): Promise<void> {
   // we can override the default value of 30 secs, by setting process.env.INDEXER_INTERVAL
   const interval = getCrawlingInterval()
   let { chunkSize } = rpcDetails
-  let lockProccessing = false
```
paulo-ocean, line 83: this was ok as it was

src/components/Indexer/crawlerThread.ts
```diff
@@ -152,6 +160,15 @@ export async function proccesNetworkData(): Promise<void> {
       )
     }
     await sleep(interval)
+    if (REINDEX_BLOCK) {
+      await deleteAllAssetsFromChain()
```
paulo-ocean, line 164: we should check the results of the calls to `deleteAllAssetsFromChain` and `updateLastIndexedBlockNumber`. We're logging success messages that we can't tell for sure that we're successful

src/components/Indexer/crawlerThread.ts
```diff
@@ -152,6 +160,15 @@ export async function proccesNetworkData(): Promise<void> {
...
+      INDEXER_LOGGER.logMessage(`Assets deleted from db for chain ${rpcDetails.chainId}`)
+      await updateLastIndexedBlockNumber(REINDEX_BLOCK)
+      INDEXER_LOGGER.logMessage(`Block updated for reindexing chain ${REINDEX_BLOCK}`)
+      lockProccessing = true
```
paulo-ocean, line 168: this is not needed:
`lockProccessing = true`
neither
`lockProccessing = false`
we're already past the `sleep(interval) `call, besides the changes on `lockProccessing` value are being done **AFTER** all the major operations (delete ddos and update block), so its kind of useless
Instead we should just do something like:
```
if (REINDEX_BLOCK && !lockProccessing) {
deleteAllAssetsFromChain()
...
updateLastIndexedBlockNumber()
...
REINDEX_BLOCK = null
..
}
```

src/components/Indexer/crawlerThread.ts
```diff
@@ -213,4 +231,9 @@ parentPort.on('message', (message) => {
       REINDEX_QUEUE.push(message.reindexTask)
     }
   }
+  if (message.method === 'reset-crawling') {
+    lockProccessing = true
```
paulo-ocean, line 235: same thing here... 
the changes to `lockProcessing` are not needed anymore, and the variable should not be global
thanks

src/components/Indexer/index.ts
```diff
@@ -128,6 +129,28 @@ export class OceanIndexer {
...
+    }
+  }
+
+  public async getLastIndexedBlock(network: number): Promise<number> {
```
bogdanfazakas, line 139: do we really need this method ? think it was removed few commits behind

src/components/Indexer/index.ts
bogdanfazakas: the worker is already created for that specific chain so the message for reindex command will go the the proper  worker

src/components/Indexer/crawlerThread.ts
bogdanfazakas, line 168: not needed here as it is maybe think about something like this:
`
  lockProccessing = true
  updateLastIndexedBlockNumber(REINDEX_BLOCK).then(lockProccessing = false)
`

src/components/Indexer/crawlerThread.ts
```diff
@@ -152,6 +160,15 @@ export async function proccesNetworkData(): Promise<void> {
       )
     }
     await sleep(interval)
```
bogdanfazakas, line 162: not really sure if this need to be done after the sleep, i would put it before the sleep but i'm open to suggestions why this should be done after

src/components/database/index.ts
```diff
@@ -564,6 +566,21 @@ export class DdoDatabase {
...
+  async deleteAllAssetsFromChain(chainId: number) {
+    const searchParameters: TypesenseSearchParams = {
+      q: `${chainId}`,
+      query_by: 'chainId'
+    }
+    const results = await this.search(searchParameters)
+    for (const res of results) {
+      if (res && res.hits) {
+        for (const h of res.hits) {
+          await this.delete(h.document.id)
```
bogdanfazakas, line 579: there should be a better way to do this, so instead of getting all ddo for the chainId then go through each of them and delete one by one think we should try something like this https://typesense.org/docs/0.24.0/api/documents.html#delete-by-query to reduce the complexity of this method

src/components/Indexer/crawlerThread.ts
paulo-ocean: didn't we agreed yesterday to do it after the sleep() ?.
before the sleep adds more complexity, because we have to deal with locking, after the sleep() we have full control of the flow

src/components/Indexer/index.ts
paulo-ocean: we don't , its not even used, we have the same inside crawlerThread

src/components/Indexer/crawlerThread.ts
paulo-ocean: we don´t need to do any locking we we do it after the sleep(), check my comments above :-)

bogdanfazakas: yep got it

src/components/Indexer/crawlerThread.ts
bogdanfazakas: i was not remembering the certain reason for having it after the sleep call. thanks

jamiehewitt15:
Hi @mariacarmina I merged the dashboard PR, can you pull the latest changes from develop?

src/test/integration/operationsDashboard.test.ts
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    provider = new JsonRpcProvider('http://127.0.0.1:8545')
+    publisherAccount = (await provider.getSigner(0)) as Signer
+  })
+
+  async function getSignature(message: string) {
+    // signing method for ganache
+    const jsonRpcSigner = new JsonRpcSigner(provider, await publisherAccount.getAddress())
+    return await jsonRpcSigner._legacySignMessage(message)
```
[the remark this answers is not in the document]
mariacarmina: ADDRESS_FILE is set in GITHUB_ENV, please check ci.yml...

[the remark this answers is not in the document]
paulo-ocean: "Publisher address is set in ALLOWED ADMINS list in setupEnv at the beginning of the tests, that's it's address, otherwise the signature won't work -> validation works."

You hardcoded the provider.getSigner(0) address. Again, this is not what i said, my point is still valid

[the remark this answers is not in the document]
paulo-ocean: "We had a discussion on it and told you that I cannot use signMessage on ganache and use that message as it is generated in signature. That's why ganache works with _legacySignMessage . On op sepolia works with signMessage."
Why not?, since every other piece of code on nodes use it? also on other tests.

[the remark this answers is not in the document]
mariacarmina: From daily discussion:
- all integration tests need update for accounts #402 
- check signMessage if works and remove _legacySignMessage

mariacarmina:
Tested on OP sepolia with the latest modification of the code (ddo db deletion + reindexing block)
```console
{
  component: 'CORE',
  moduleName: 'CORE',
  level: 'info',
  message: '[CORE] => Checking received command data for Command "reindexChain": {\n' +
    '    "signature": "0x09f28ce29c6ce6c42d3595df5727ade2df0b5ca40d103dbd534443942d010323243f8d9d6168f2888c7ffc3a41798e2619dbfe83040b77bafc7739df136ee6151c",\n' +
    '    "command": "reindexChain",\n' +
    '    "chainId": 11155420,\n' +
    '    "expiryTimestamp": 1744664400000\n' +
    '}',
  timestamp: '2024-04-24 15:12:33'
}
{
  component: 'HTTP',
  moduleName: 'HTTP',
  level: 'info',
  message: '[HTTP] => [HTTP] => Sending command : {"signature":"0x09f28ce29c6ce6c42d3595df5727ade2df0b5ca40d103dbd534443942d010323243f8d9d6168f2888c7ffc3a41798e2619dbfe83040b77bafc7739df136ee6151c","command":"reindexChain","chainId":11155420,"expiryTimestamp":1744664400000}',
  timestamp: '2024-04-24 15:12:33'
}
{
  component: 'OCEANNODE',
  moduleName: 'OCEANNODE',
  level: 'info',
  message: '[OCEANNODE] => [OCEANNODE] => Incoming direct command for ocean peer',
  timestamp: '2024-04-24 15:12:33'
}
{
  component: 'OCEANNODE',
  moduleName: 'OCEANNODE',
  level: 'info',
  message: '[OCEANNODE] => [OCEANNODE] => Performing task: {"signature":"0x09f28ce29c6ce6c42d3595df5727ade2df0b5ca40d103dbd534443942d010323243f8d9d6168f2888c7ffc3a41798e2619dbfe83040b77bafc7739df136ee6151c","command":"reindexChain","chainId":11155420,"expiryTimestamp":1744664400000}',
  timestamp: '2024-04-24 15:12:33'
}
{
  component: 'CORE',
  moduleName: 'CORE',
  level: 'info',
  message: '[CORE] => Checking received command data for Command "reindexChain": {\n' +
    '    "signature": "0x09f28ce29c6ce6c42d3595df5727ade2df0b5ca40d103dbd534443942d010323243f8d9d6168f2888c7ffc3a41798e2619dbfe83040b77bafc7739df136ee6151c",\n' +
    '    "command": "reindexChain",\n' +
    '    "chainId": 11155420,\n' +
    '    "expiryTimestamp": 1744664400000\n' +
    '}',
  timestamp: '2024-04-24 15:12:33'
}
{
  component: 'CORE',
  moduleName: 'CORE',
  level: 'info',
  message: '[CORE] => Checking received command data for Command "reindexChain": {\n' +
    '    "signature": "0x09f28ce29c6ce6c42d3595df5727ade2df0b5ca40d103dbd534443942d010323243f8d9d6168f2888c7ffc3a41798e2619dbfe83040b77bafc7739df136ee6151c",\n' +
    '    "command": "reindexChain",\n' +
    '    "chainId": 11155420,\n' +
    '    "expiryTimestamp": 1744664400000\n' +
    '}',
  timestamp: '2024-04-24 15:12:33'
}
{
  component: 'CORE',
  moduleName: 'CORE',
  level: 'info',
  message: 'Resolved signer address: 0xf96da2e8b85482f8ed870fd8dbb2374fc74d0779',
  timestamp: '2024-04-24 15:12:33'
}
{
  component: 'CORE',
  moduleName: 'CORE',
  level: 'info',
  message: 'Reindexing chain command called',
  timestamp: '2024-04-24 15:12:33'
}
{
  component: 'INDEXER',
  moduleName: 'INDEXER',
  level: 'info',
  message: 'Assets deleted from db for chain 11155420',
  timestamp: '2024-04-24 15:12:40'
}
{
  component: 'INDEXER',
  moduleName: 'INDEXER',
  level: 'info',
  message: '[INDEXER] => [INDEXER] => New last indexed block : 3867703',
  timestamp: '2024-04-24 15:12:40'
}
{
  component: 'INDEXER',
  moduleName: 'INDEXER',
  level: 'info',
  message: 'Block updated for reindexing chain 3867703',
  timestamp: '2024-04-24 15:12:40'
}
```

src/test/integration/operationsDashboard.test.ts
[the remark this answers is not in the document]
paulo-ocean:
```
const wallet = new ethers.Wallet(
    '0xc594c6e5def4bab63ac29eed19a134c130388f74f019bc74b8f4389df2837a58', **provider**
  )
```
you have to set the provider here as well

mariacarmina:
When retrieving an existent DDO in the database after reindex chain:
```console
{
  component: 'DATABASE',
  moduleName: 'DATABASE',
  level: 'error',
  message: '[DATABASE] => ❌ [DATABASE] => DDO entry with ID did:op:963e0d732881a0655ecd4f26612c92c258d378b75f1cc3d2338b22be1f405e03 not found in any schema.',
  timestamp: '2024-04-24 16:04:12'
}
{
  component: 'CORE',
  moduleName: 'CORE',
  level: 'info',
  message: '[CORE] => Checking received command data for Command "getDDO": {\n' +
    '    "id": "did:op:963e0d732881a0655ecd4f26612c92c258d378b75f1cc3d2338b22be1f405e03",\n' +
    '    "command": "getDDO"\n' +
    '}',
  timestamp: '2024-04-24 16:04:13'
}
{
  component: 'DATABASE',
  moduleName: 'DATABASE',
  level: 'error',
  message: '[DATABASE] => ❌ [DATABASE] => DDO entry with ID did:op:963e0d732881a0655ecd4f26612c92c258d378b75f1cc3d2338b22be1f405e03 not found in any schema.',
  timestamp: '2024-04-24 16:04:13'
}
```
on ocean-cli:

``` console
Resolving Asset with DID: did:op:963e0d732881a0655ecd4f26612c92c258d378b75f1cc3d2338b22be1f405e03
Error fetching Asset with DID: did:op:963e0d732881a0655ecd4f26612c92c258d378b75f1cc3d2338b22be1f405e03.  Does this asset exists?
```

jamiehewitt15:
When you tested locally with postman how did you generate the signature?

src/components/Indexer/crawlerThread.ts
```diff
@@ -152,6 +158,25 @@ export async function proccesNetworkData(): Promise<void> {
...
+    if (REINDEX_BLOCK && !lockProccessing) {
+      const res = await deleteAllAssetsFromChain()
+      if (res === null) {
+        INDEXER_LOGGER.error(
+          `Assets could not be deleted for chain ${rpcDetails.chainId}. Continue indexing normally...`
+        )
+        continue
+      }
+      INDEXER_LOGGER.logMessage(`Assets deleted from db for chain ${rpcDetails.chainId}`)
+      // TODO: if block couldn't be updated and DDOs are deleted => should set in DDO mapping a field for archieved DDOs.
```
jamiehewitt15, line 171: Can you make sure we have an issue for this? Leaving a `TODO` in the code isn't a great way to track tasks as we'll forget that it's here.

src/components/database/typesense.ts
```diff
@@ -69,6 +69,13 @@ class TypesenseDocuments {
     return this.api.delete<TypesenseDocumentSchema>(path)
   }
 
+  // eslint-disable-next-line require-await
+  async deleteByChainId(filterCondition: string) {
+    // TODO: Use the batch_size parameter to control the number of documents. A larger value will speed up deletions, but will impact performance.
```
jamiehewitt15, line 74: Please also make sure there is an issue for tracking this. Ideally, the `TODO` comment should contain a link to the github issue.

mariacarmina:
> When you tested locally with postman how did you generate the signature?

Hello @jamiehewitt15,
 I attach here the code, the signature generation is in the tests  in `getsignature` function in `operationsDashboard.test.ts`.
In the code below I generated the signature for one year.
 
 ```javascript
 import { JsonRpcProvider, Signer, Wallet } from 'ethers'
const currentDate = new Date()
const expiryTimestamp = new Date(
  currentDate.getFullYear() + 1,
  currentDate.getMonth(),
  currentDate.getDate()
).getTime()
const providerRemote = new JsonRpcProvider('https://sepolia.optimism.io/')
const me = new Wallet(
  <PRIVATE_KEY>,
  providerRemote
) as Signer
console.log(await me.signMessage(expiryTimestamp.toString()))
 ```

src/components/Indexer/crawlerThread.ts
mariacarmina: Created #405, will update the comment as well.

src/components/database/typesense.ts
mariacarmina: created #406, update the code now.

mariacarmina:
tests fail due to timeout for waiting on contracts and c2d deployment

jamiehewitt15:
> tests fail due to timeout for waiting on contracts and c2d deployment

Yeah, that's a separate issue. Not related to this PR

jamiehewitt15:
When I call the endpoint, I get `REINDEXING CHAIN...` as the response. How do we know when the reindexing is complete? I'm asking because it would be nice in the dashboard to show "Reindexing started" and "Reindexing finished"

src/components/Indexer/crawlerThread.ts
```diff
@@ -58,11 +59,17 @@ async function getLastIndexedBlock(): Promise<number> {
...
-      LOG_LEVELS_STR.LEVEL_ERROR,
-      'Error retrieving last indexed block',
-      true
-    )
+    INDEXER_LOGGER.error(`Error retrieving last indexed block: ${err}`)
+    return null
+  }
+}
+
+async function deleteAllAssetsFromChain(): Promise<void> {
```
paulo-ocean, line 67: it would be nice if these functions actually return something, so we know if we succeeded or not..
we're not returning anything in case of "non error", and returning 'null' if we catch an exception..
we should probably also have the DB function `deleteAllAssetsFromChain` return something if both cases, either number of deleted assets or -1 if nothing was deleted or we had an error... or even true or false. These functions are not much coherent as they are

src/components/Indexer/crawlerThread.ts
```diff
@@ -152,6 +160,30 @@ export async function proccesNetworkData(): Promise<void> {
...
+    }
+  }
+}
+
+async function reindexChain(currentBlock: number): Promise<void> {
+  const block = await updateLastIndexedBlockNumber(REINDEX_BLOCK)
+  if (block !== null) {
+    INDEXER_LOGGER.logMessage(`Block updated for reindexing chain ${REINDEX_BLOCK}`)
+    REINDEX_BLOCK = null
+    const res = await deleteAllAssetsFromChain()
```
paulo-ocean, line 175: `const res = await deleteAllAssetsFromChain()`
this only returns 'null' in case of an exception,  
The block on the called function:
```
try {
    await ddo.deleteAllAssetsFromChain(rpcDetails.chainId)
  } catch (err) {
    INDEXER_LOGGER.error(`Error deleting all assets: ${err}`)
    return null
  }
```
will return 'undefined' if not exception is thrown... 

Also this part:

```
const block = await updateLastIndexedBlockNumber(REINDEX_BLOCK)
  if (block !== null) {
```
It will never return 'null'

again, this is not much consistent

src/components/Indexer/crawlerThread.ts
mariacarmina: Ok, but this is a general issue with the majority of the db methods. I kept consistent with the current version of db operations (in case of error, we return null). I'll modify for this function particulary, but I think it's best for future to modify for all in a separate issue, in order to maintain consistency.

mariacarmina: my comment was addressed including to CRUD database/index.ts file, not only from Indexer/crawlerThread.

src/components/Indexer/index.ts
bogdanfazakas: we don't need this here since we have now the metod inside the class you mentioned it's used

src/components/Indexer/crawlerThread.ts
```diff
@@ -58,30 +60,38 @@ async function getLastIndexedBlock(): Promise<number> {
...
 }
 
+async function deleteAllAssetsFromChain(): Promise<number> {
+  const { ddo } = await getDatabase()
+  try {
+    const res = await ddo.deleteAllAssetsFromChain(rpcDetails.chainId)
+    return res.num_deleted
+  } catch (err) {
+    INDEXER_LOGGER.error(`Error deleting all assets: ${err}`)
+    return -1
```
bogdanfazakas, line 75: Why returning -1 instead of throwing a propper error to the upper level method used to call this

src/components/Indexer/crawlerThread.ts
```diff
@@ -152,6 +162,30 @@ export async function proccesNetworkData(): Promise<void> {
...
     }
     await sleep(interval)
+    // reindex chain command called
+    if (REINDEX_BLOCK && !lockProccessing) {
+      await reindexChain(currentBlock)
+    }
+  }
+}
+
+async function reindexChain(currentBlock: number): Promise<void> {
```
paulo-ocean, line 172: i still find this block of code/function a little bit strange, and perhaps too much complicated for what it should too..
For instance, we are logging too much again... 
the "`deleteAllAssetsFromChain`" functions already do some logging that we eventually repeat
anyway, most important imo is this part:
```
const block = await updateLastIndexedBlockNumber(REINDEX_BLOCK)
if (block !== -1) {
....
} else {
    INDEXER_LOGGER.error(`Block could not be reset. Continue indexing normally...`)
    REINDEX_BLOCK = null
  }
```
So, if we failed to update the block to REINDEX_BLOCK, we set it to null again, which means we already lost the context of the reindexCommand, IMO, we should only set REINDEX_BLOCK to null if everything was ok.. otherwise we need to repeat the command again (but we have no idea if the command failed or not, because the handler only calls a static method and sends a 200 back)

Also, it seems to me that we should try to delete the assets first, and only then update the block (just once) , either with current block or the other one...considering the results of the deletion... we are doing more stuff than needed and reverting, if we delete the assets successfully, we definitely need to reset the block to the REINDEX_BLOCK, cause the assets are gone..

thanks

paulo-ocean: as discussed we're gonna address cleaning the logs and keep most of the current logic. We will have separate issues, for instance, for sending a feedback/response to the issuer of the command (right now we don't know if commands succeed or not, we only know if they get delivered)

paulo-ocean (approved):
lgtm, we can always improve on separate issues anyway

PR #336, opened 2024-03-12
Reindex chain handler.

### LINKED ISSUE #297: BE: Allow admin to reindex a chain
The admin needs to be able to force restart indexing a chain. The assets on that chain will be deleted and indexing will start from the beginning. The other chains are unaffected

Let's expose that endpoint, which is only allowed for admins

Params: chain id

mariacarmina:
Fixes #297  .

Changes proposed in this PR:

- added tests for admin dashboard handlers
- implemented reindex chain handler
- fix signature for auth function


jamiehewitt15:
Are you testing this with postman? It would be helpful to have some test coverage for this new handler

src/@types/commands.ts
```diff
@@ -113,6 +113,12 @@ export interface ReindexTxCommand extends Command {
   txId: string
 }
 
+export interface ReindexChainCommand extends Command {
```
paulo-ocean, line 116: Same comments as for the other Admin commands... they should extend a common higher interface, but not the base Command, this one  is pretty much the same as ReindexTxCommand (which should extend from this one)
Start with some AdminCommand or similar, that extends Command (with expiryTimestamp and signature) than extend and build them up from there, not repeating existing fields
tkx

src/components/core/adminOperations.ts
```diff
@@ -121,3 +129,77 @@ export class ReindexTxHandler extends Handler {
...
+export class ReindexChainHandler extends Handler {
+  validate(command: ReindexChainCommand): ValidateParams {
+    const commandValidation = validateCommandParameters(command, [
+      'expiryTimestamp',
+      'signature',
+      'chainId'
+    ])
+    if (!commandValidation.valid) {
+      const errorMsg = `Command validation failed: ${JSON.stringify(commandValidation)}`
+      CORE_LOGGER.logMessage(errorMsg)
```
paulo-ocean, line 142: Same thing as other PR, copy/paste is usually not the best idea :-)

src/components/core/adminOperations.ts
```diff
@@ -121,3 +129,77 @@ export class ReindexTxHandler extends Handler {
...
+      return buildInvalidRequestMessage(errorMsg)
+    }
+    return commandValidation
+  }
+
+  async handle(task: ReindexTxCommand): Promise<P2PCommandResponse> {
+    const validation = this.validate(task)
+    if (!validation.valid) {
+      return new Promise<P2PCommandResponse>((resolve, reject) => {
```
paulo-ocean, line 156: same thing

src/components/core/adminOperations.ts
```diff
@@ -121,3 +129,77 @@ export class ReindexTxHandler extends Handler {
...
+        resolve(buildInvalidParametersResponse(validation))
+      })
+    }
+    CORE_LOGGER.logMessage(`Reindexing chaincommand called`)
+    const config = await getConfiguration()
+    const blockchain = new Blockchain(
+      config.supportedNetworks[task.chainId.toString()].rpc,
```
paulo-ocean, line 163: same, no checks on chainId, that breaks it

src/components/core/adminOperations.ts
```diff
@@ -121,3 +129,77 @@ export class ReindexTxHandler extends Handler {
...
+      if (!ret) {
+        CORE_LOGGER.log(
+          LOG_LEVELS_STR.LEVEL_ERROR,
+          `Reindex chain failed on chain ${task.chainId}.`,
+          true
+        )
+        return
+      }
+
+      return new Promise<P2PCommandResponse>((resolve, reject) => {
```
paulo-ocean, line 191: same, no need for this Promise sintax

src/components/core/adminOperations.ts
mariacarmina: there is no copy/paste involved... if you look closely `reindex-chain-command` starts from `reindex-tx-command` -> all the changes from reindex-tx-command will be reflected here as well. I did that to avoid conflicts and those issues are sort of related to each other...

src/@types/commands.ts
mariacarmina: Ok, now they extend `AdminCommand`. I'll also add the admin prefix to thse admin commands

src/components/core/adminOperations.ts
mariacarmina: I'll keep the promise syntax only for stop node handler, because there is a bigger workaround to modify the `handle` return type to each handler. Tried here and lint failed for all handlers https://github.com/oceanprotocol/ocean-node/pull/336/commits/3c2584ea5708af0cb03fd64ef8698e5d32bff63a.
The rest of admin handlers do not have promise syntax.

paulo-ocean: yes, the Stop node is different because of the process.exit(), but for all the others we can keep using "async handle" and remove all the Promise stuff, as its implicit

mariacarmina:
These modifications will be merged into `reindex-tx-command`, not `develop` branch.

src/components/core/adminOperations.ts
```diff
@@ -104,14 +107,83 @@ export class ReindexTxHandler extends Handler {
...
+        status: { httpStatus: 200 },
+        stream: new ReadableString('REINDEX TX OK')
+      }
     } catch (error) {
       CORE_LOGGER.log(LOG_LEVELS_STR.LEVEL_ERROR, `REINDEX tx: ${error.message} `, true)
     }
   }
 }
+
+export class ReindexChainHandler extends Handler {
```
jamiehewitt15, line 120: I suggest a refactoring of this file, it's going to get too big if we add every command that requires validation into this same file. Also, the logic for validating the signature can be moved to an abstract class that these handlers inherit from. My suggestions are: 

1. create a `src/components/core/admin` folder
2. create `src/components/core/admin/adminHandler.ts` which contains an abstract class which handles the validation. Every admin component that requires validation can inherit from this one. 
3. move each admin handler into their own file within the `src/components/core/admin` folder. That's going to make much smaller files which are easier to work with and also easier to find the relevant handler. 

I have implemented each of these already on my [PR](https://github.com/oceanprotocol/ocean-node/pull/368/files#diff-38ce78541024a0ba03b457d21e831f47ae5681418f80491439e8fb33083470ef) related to logs and stop node. Can you have a look at what I did there and consider the same for this PR.

mariacarmina: Ok, I understand your point, maybe it's better to merge it into reindex-tx-command branch and I'll keep it in mind to do that refactor on `reindex-tx-command` branch. How does it sound? @jamiehewitt15 @paulo-ocean

PR #337, opened 2024-03-12
fix write after response end

### LINKED ISSUE #335: Node crashes when response status !== 200
This one was under the wood for a long time :-)
Its on the sink function that reads and streams the commands responses. 
The node crashes when the status response is !== 200 (mostly on specific errors, without a stream response, cause that´s where we usually use other status codes). 
If its 200 then its fine
It basically tries to write more bytes after closing the stream
ERR_STREAM_WRITE_AFTER_END

paulo-ocean:
Fixes #335 .

Changes proposed in this PR:

- remove unnecessary write() call . (we already use pipe() to handle the reponses)


src/components/httpRoutes/commands.ts
```diff
@@ -60,11 +60,6 @@ directCommandRoute.post(
                 isBinaryContent = true
               }
             }
-            if (decoded.httpStatus !== 200) {
```
bogdanfazakas, line 63: is there a probability we can receive multiple chunks for a request with status different from 200 (maybe set by mistake to another status code by a different node) ?
do we still need to have the check and just break the for loop

paulo-ocean: all the responses are already handled either on
```
async handleDirectProtocolCommand(
    message: string,
    sink: any
  ): Promise<P2PCommandResponse> {
```
OR
`export async function handleProtocolCommands(connection: any) {`

So, basically here, we were just writing the bytes again to a stream that was already closed, the fact that it was a check on status 200 is not the important part... it could be any other status... it just happens it was that one (we were doing something already done).. 

Here (on the sink function) we just need to set proper status headers (if any) or check if the response content is string or binary, so the chunks are read correctly. We don't need any other checks.

Regarding your question, no, there is not.. each command returns only a P2PCommandResponse object, with a status and a stream (or not)

bogdanfazakas (approved):

jamiehewitt15 (approved):
Tested locally and it solves the problem.

PR #327, opened 2024-03-08
dashboard authentication

jamiehewitt15:
Fixes # .

Changes proposed in this PR:

- dashboard authentication

PR #338, opened 2024-03-13
Making the dashboard optional

jamiehewitt15:
Fixes #327 

Changes proposed in this PR:

- Making the dashboard optional
- Dashboard still runs by default, the new env is optional

To test:
```
export DASHBOARD="false"
npm run build
npm run start 
```

Navigate to http://localhost:8000/dashboard/ and there should be no dashboard. 

Now run: 
```
export DASHBOARD="true"
npm run build
npm run start 
```

and the dashboard should be running at http://localhost:8000/dashboard/


paulo-ocean (approved):
lgtm, apart that unit test we know it fails atm

bogdanfazakas (approved):
lgtm, think we should merge first the PR with env-md docs and add the `Dashboard` there as well

Issue #65, opened 2023-10-26
Retrieve and display connected nodes


LoznianuAnamaria:
Display the list of the nodes connected and also display the public status of a connected node.

Note: Each node has an IP address. We can use it to identify the geographical localization of the node. We can display them all on a map

mihaisc:
currently we can get just a list of ids

`[
    "16Uiu2HAkwWe6BFQXZWg6zE9X7ExynvXEe9BRTR5Wn3udNs7JpUDx",
    "16Uiu2HAkyqdMojcJGsyUJpNLRbwVPT9ZWK63qmY2dy58zwi5obhH"
]`

jamiehewitt15:
Yeah for now this task is just: 

1. call the `/getOceanPeers` endpoint
2. Show a list of the node IDs that this node is connected to

No geographical map or anything like that.

paulo-ocean:
closed on [https://github.com/oceanprotocol/ocean-node/pull/339](url)

PR #339, opened 2024-03-14
Feat/65 retrieve and display connected nodes

tom1145:
Fixes [#65](https://github.com/oceanprotocol/ocean-node/issues/65).

Changes proposed in this PR:

- list all the connected nodes.


dashboard/src/components/Dashboard/index.tsx
```diff
@@ -139,6 +162,20 @@ export default function Dashboard() {
                 <div className={styles.title24}>Address</div>
                 {truncateString(data?.address, 12)}
               </div>
+              {nodePeers.length > 0 && (
```
paulo-ocean, line 165: I think it would be nice to always show the header/title, even if no peers are connected (or maybe also show a msg like "no connected peers" if none is available..
Also, i think it would be nice to have a copy/paste icon, on all fields that currently are being shortened (node id, address, node peers, etc) , or at least it should copy to clipboard on click
Apart from those details, looks good
thanks

jamiehewitt15: Yeah, those are good points

jamiehewitt15:
Can you move all of this into a new component in a separate file and then import it, `dashboard/src/components/Dashboard/index.tsx` is already too big. We should also move `ConnectionDetails`, `Indexer` and `AdminAccounts` into separate files but you don't need to do that in this PR.

dashboard/src/components/Dashboard/index.tsx
tom1145: Thanks for the feedback, I'll work on it! I will let you know when it's done.

jamiehewitt15 (approved):

paulo-ocean (approved):
lgtm. maybe in the future we could listen for events when another peer joins, so we don't need to reload the page, but not for now of course

paulo-ocean:
i think we can merge this.. its been approved last week already, i'm merging it

PR #344, opened 2024-03-15
Fix/ Running local integration tests

### LINKED ISSUE #280: Integration tests don't reliably work when running locally
Integration tests don't reliably work when running locally, most of the time they fail even though they work fine on the CI. We need to figure out what's causing the issue and find a fix.

bogdanfazakas:
Fixes #280  .

Changes proposed in this PR:

- added missing env vars for each integration test
- most common error was generated by missing `ADDRESS_FILE`  required to create contract interactions on ganache network
- added missing constants 
- minor update to `buildEnvOverrideConfig()`


paulo-ocean:
what am i missing @bogdanfazakas  ?
they don't run locally on my end (only 3)

```
  3 passing (11s)
  1 failing

  1) Compute
       Initialize compute without transaction IDs:
     TypeError: Cannot read properties of undefined (reading 'id')
      at Context.<anonymous> (file:///home/paulo/workspace/OCEAN/ocean-node/dist/test/integration/compute.test.js:106:31)
      at callFn (/home/paulo/workspace/OCEAN/ocean-node/node_modules/mocha/lib/runnable.js:366:21)
```

bogdanfazakas:
> what am i missing @bogdanfazakas ? they don't run locally on my end (only 3)
> 
> ` npm run test:integration`
> 
> ```
>   3 passing (11s)
>   1 failing
> 
>   1) Compute
>        Initialize compute without transaction IDs:
>      TypeError: Cannot read properties of undefined (reading 'id')
>       at Context.<anonymous> (file:///home/paulo/workspace/OCEAN/ocean-node/dist/test/integration/compute.test.js:106:31)
>       at callFn (/home/paulo/workspace/OCEAN/ocean-node/node_modules/mocha/lib/runnable.js:366:21)
> ```

hmm let me double check compute i've touched that today

paulo-ocean:
yes, i do `./startOcean --with-c2d`

jamiehewitt15:
Yeah, it also didn't work for me. I'm also running `./startOcean --with-c2d`. @bogdanfazakas can you share the env values that you are using, in case that makes a difference?

jamiehewitt15:
Not sure what's going on in my local setup but I'm getting this error: 

```
  1 passing (86ms)
  1 failing

  1) Compute
       should publish compute datasets & algos:
     TypeError: Cannot read properties of undefined (reading 'ERC721Factory')
      at publishAsset (file:///home/jamie/Desktop/ocean/ocean-node/dist/test/utils/assets.js:20:57)
      at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
      at async Context.<anonymous> (file:///home/jamie/Desktop/ocean/ocean-node/dist/test/integration/compute.test.js:63:35)
```

Tried deleting node modules and doing a fresh `npm i` but that did not solve it

bogdanfazakas:
Screenshot 2024-03-22 at 09 37 43
tested on the node3, no env var set all tests passing

paulo-ocean:
> Screenshot 2024-03-22 at 09 37 43
> tested on the node3, no env var set all tests passing

well, i still have the same exact issue i mentioned above. Keeps failing there.. have unset all variables, use 'localhost', etc
still nothing :-(

paulo-ocean:
> Something weird is going on on my side. I'm using the latest version of barge:
> 
> ```
> {
>   component: 'DATABASE',
>   moduleName: 'DATABASE',
>   level: 'error',
>   message: '[DATABASE] => ❌ [DATABASE] => DDO entry with ID did:op:8bc47e948ec44d20e0ff13410ab8386a13f71c39c93d4be09e6d844d2619b3e0 not found in any schema.',
>   timestamp: '2024-03-25 15:55:54'
> }
> {
>   component: 'DATABASE',
>   moduleName: 'DATABASE',
>   level: 'error',
>   message: '[DATABASE] => ❌ [DATABASE] => DDO entry with ID did:op:8bc47e948ec44d20e0ff13410ab8386a13f71c39c93d4be09e6d844d2619b3e0 not found in any schema.',
>   timestamp: '2024-03-25 15:55:55'
> }
>     1) should publish compute datasets & algos
> {
>   component: 'CONFIG',
>   moduleName: 'CONFIG',
>   level: 'warn',
>   message: '[CONFIG] => Detected Ocean Node Configuration change... This might have unintended effects',
>   timestamp: '2024-03-25 15:55:55'
> }
> 
> 
>   1 passing (20s)
>   1 failing
> 
>   1) Compute
>        should publish compute datasets & algos:
>      Error: Timeout of 20000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/jamie/Desktop/ocean/ocean-node/dist/test/integration/compute.test.js)
>       at createTimeoutError (/home/jamie/Desktop/ocean/ocean-node/node_modules/mocha/lib/errors.js:498:15)
>       at Runnable._timeoutError (/home/jamie/Desktop/ocean/ocean-node/node_modules/mocha/lib/runnable.js:431:10)
>       at Timeout.<anonymous> (/home/jamie/Desktop/ocean/ocean-node/node_modules/mocha/lib/runnable.js:246:24)
>       at listOnTimeout (node:internal/timers:569:17)
>       at process.processTimers (node:internal/timers:512:7)
> ```

Well, it doesn't work for me either...

src/components/httpRoutes/index.ts
```diff
@@ -16,7 +16,8 @@ export * from './getOceanPeers.js'
 export const httpRoutes = express.Router()
 
 // P2P routes related
-export const hasP2PInterface = await (await getConfiguration()).hasP2P
+export const hasP2PInterface = (await (await getConfiguration())?.hasP2P) || false
```
paulo-ocean, line 19: why do we need this? 
its also might eventually change the default 
(the default is have both interfaces; p2p and http, so as long as it is not clearly otherwise, both should be true)
thanks

paulo-ocean: Ok, I see why you did the change, but it should not have the dependency cycle.. i still think this const must be moved somewhere else, for now, lets keep it

paulo-ocean: Or just change on config.ts:
export const hasP2PInterface = getNodeInterfaces().includes('P2P')
and remove the const with the same name on httpRoutes/index.ts
and update all the imports

paulo-ocean: well, this is still not OK... ( did a `console.log() `after the const definitions)
With my change, the tests do not pass because the value of `hasP2PInterface` is '**true**', which is the expected... The node HAS P2P
With your change, all the tests pass indeed, but the value of `hasP2PInterface` is '**false**', which is not the expected... 
So, the problem is in reality somewhere else

paulo-ocean (approved):
lgtm, i was able to run them..
I think we still have some issues that are not very clear to understand they happen, 
but eventually for another PR

alexcos20 (approved):

Issue #364, opened 2024-03-20
Default `/logs` endpoint only sends the last 24hrs


jamiehewitt15:
The default `/logs` endpoint only sends the last 24hrs which can result in little or no logs in many cases. It would be more useful if the default endpoint sends the last 100 logs, regardless of when they were generated. Then users can still use the query parameters to get a more specific time-period.

Issue #357, opened 2024-03-18
Admin dashboard: Logs should be protected


alexcos20:
Status quo:  anyone can requests the logs from ocean-node, but they might contain sensible information.

To do:
 - backend endpoint should be protected by auth  (do it in the handler)
 - dashboard:  should sent auth information before calling the endpoint

Issue #326, opened 2024-03-08
Replace custom wallet connection with rainbowKit or connectkit


jamiehewitt15:
- [ ] Replace custom wallet connection with rainbowKit or connectkit

PR #368, opened 2024-03-21
Dashboard authentication for stop node & logs

### LINKED ISSUE #293: Add Stop node interface
Using endpoint exported by https://github.com/oceanprotocol/ocean-node/issues/292, allow admin to stop (restart) the node

- [ ] auth admin by asking to sign a message
- [ ] expose stop button in UI

Will use endpoint exposed by #292

jamiehewitt15:
Fixes #293 
Fixes: #357
Fixes: #296 
Fixes: #364
Fixes: #326

Changes proposed in this PR:

- Stop node command requires the signing of the message
- Logs require the signing of a message
- Both of the `/logs` and `/logs:id` endpoints are now post requests where the body of the request includes `expiryTimestamp` and `signature`
- Signature is stored in local storage
- `expiryTimestamp` is set to 12hrs from curent time. 
- Using rainbow kit for wallet connection
- Using MUI for easier styling of components
- Logs now send up to 90 days of logs (with a maximum of 100 logs) by default. 
- Refactoring of admin handlers

To Test:

1. make sure that you export `ALLOWED_ADMINS` and include your wallet address in the array.
2. `npm run build && npm run start` from the root of the directory 
3. navigate to `http://localhost:8000/dashboard`
4. connect wallet
5. Click "unlock" to sign a mesage which will authorise you to call the admin endpoints for 12hrs. 
6. Attempting to do the same with an account that isn't listed in the `ALLOWED_ADMINS`  array should not work. You should see a message saying "Your account does not have admin access"


paulo-ocean:
Hi @jamiehewitt15 
I think something weird is going on with the wallet connection.
when i try to connect the wallet, it always wants to switch the network to Ethereum.
Tried setting up barge/ganache and also Sepolia on the RPCS, but same result...
I can see the correct node network on the dashboard, but the "connect wallet" always want to switch to Ethereum mainnet
switch

jamiehewitt15:
> Hi @jamiehewitt15 I think something weird is going on with the wallet connection. when i try to connect the wallet, it always wants to switch the network to Ethereum. Tried setting up barge/ganache and also Sepolia on the RPCS, but same result... I can see the correct node network on the dashboard, but the "connect wallet" always want to switch to Ethereum mainnet !

I've added ganache now, it was just mainnet and polygon before. It actually doesn't matter what network you're on as it's just signing a message. There are no transaction fees

jamiehewitt15:
Also added sepolia

paulo-ocean (approved):
lgtm, 
would suggest maybe add the message  "Your account does not have admin access" in red color or something (becomes more visible)

bogdanfazakas:
Looks good over all, but there are small things that think we can improve:

- an action would be nice after stopping the node, maybe adding a layer or hiding all the content, inactivating the page or smth like a transparent background with a reload idk 😄. Mainly because after the node is stopped you have no clue that it not running anymore and you can click on various buttons links that do no respond.
- we also need to fix the following scenario:
          1. user logs in with wallet a that is admin, and has acces to admin methods
          2. user select disconnect
          3. user connects with different wallet address, but still has acces to admin methods and the call works
          4. if you hit refresh at one stage everything is

jamiehewitt15:
Hey @bogdanfazakas the issue with switching accounts should be fixed now, can you test it again? 

Regarding some visual feedback to show that the node is down, I've implemented a very simple update: there is an alert (saying the node is down and the dashboard won't be displayed anymore) and then the dashboard is gone. Hopefully, that gets the message across. Let me know what you think.

PR #370, opened 2024-03-21
Fix/ Check proper order tx

### LINKED ISSUE #351: Security: Order transactions
Ocean-node hardcodes startOrder events in a tx, and it only takes the 1st one

(see https://github.com/oceanprotocol/ocean-node/blob/develop/src/components/core/utils/validateOrders.ts#L98)

scenario: you can create a smart contracts with 2 orders and 2 order started events in the same transaction, and then you can consume the 2nd asset while paying for the 1st one, which of course, is much cheaper or even free, created by attacker

Let's check contract address, payer, consumer, then retrieve the right eventIndex and extract the correct event

bogdanfazakas:
Fixes #351  .

Changes proposed in this PR:

- 
- 
-


paulo-ocean:
hey @bogdanfazakas this is a bit confusing with the stuff from the other PR :-) ...
what are the changes related with this PR only?
thanks

src/components/core/utils/validateOrders.ts
```diff
@@ -74,22 +81,33 @@ export async function validateOrderTransaction(
...
   )
-  if (
-    userAddress.toLowerCase() !== OrderStartedEvent[0].args[0].toLowerCase() &&
-    userAddress.toLowerCase() !== OrderStartedEvent[0].args[1].toLowerCase()
-  ) {
+  let orderEvent
+  for (const event of OrderStartedEvent) {
+    if (
+      (userAddress.toLowerCase() !== event.args[0].toLowerCase() &&
+        userAddress.toLowerCase() !== event.args[1].toLowerCase()) ||
```
paulo-ocean, line 93: this means  `userAddress` can be present at either on `event.args[0]` or `event.args[1] `? , or is this a typo?
by the way, imo, the loop would probably make more sense the other way around, eg;
If you want to break when you find the orderEvent i would do instead:
```
for (const event of OrderStartedEvent) {
    if (
      (userAddress.toLowerCase() === event.args[0].toLowerCase() ||
        userAddress.toLowerCase() === event.args[1].toLowerCase() ) &&
      erc20Address.toLowerCase()  === datatokenAddress.toLowerCase() &&
      erc721Address.toLowerCase()  === dataNftAddress.toLowerCase()
    ) {
      orderEvent = event
      break
    }
  }
```
I always find the usage of both "continue" and "break" together a bit strange, 
i mean, the normal behaviour of a loop is to "continue" to next iteration, so when we need to break the loop, we would just look and do it on that condition, its always cleaner imo...
anyway, just an hint

bogdanfazakas: updated the loop, makes more sense indeed, regarding the user address check it is not a type it can be either the payer or the consumer and would validate the flow.

paulo-ocean (approved):
lgtm

package.json
```diff
@@ -29,7 +29,7 @@
     "test": "npm run lint && npm run test:unit:cover && npm run test:integration:cover",
     "test:unit": "npm run build-tests && npm run mocha \"./dist/test/unit/**/*.test.js\"",
     "test:integration": "npm run build-tests && npm run mocha \"./dist/test/integration/**/*.test.js\"",
-    "test:integration:compute": "npm run build-tests && npm run mocha ./dist/test/integration/compute.test.js",
+    "test:integration:download": "npm run build-tests && npm run mocha ./dist/test/integration/download.test.js",
```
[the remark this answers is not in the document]
bogdanfazakas: actually removing it since not needed, if someone needs to test only one test can add this locally or run the proper cmd line comand, but should not be in the original package.json file

mariacarmina (approved):
lgtm

PR #373, opened 2024-03-25
Typsense pagination

### LINKED ISSUE #330: Introduce pagination for database queries
Typesense has a [limit of 250 results per page](https://github.com/typesense/typesense/issues/780), which may not be enough for some use cases. The solution here would be to introduce pagination on our search queries.

jamiehewitt15:
Fixes #330 

Changes proposed in this PR:

- Adding pagination to each of our search methods within `src/components/database/index.ts`
- Updating the logs HTTP route so that it can be used for paginated requests.
- Adding additional tests on the logs to test the pagination.


src/components/database/index.ts
```diff
@@ -224,19 +252,43 @@ export class DdoDatabase {
     }
   }
 
-  async search(query: Record<string, any>) {
+  async search(
```
paulo-ocean, line 255: Hi @jamiehewitt15 
Shouldn't we extend also the `QueryCommand`? The `search()` method is also used by the `QueryHandler`
Right now the command interface is:
```
export interface QueryCommand extends Command {
  query: Record<string, any>
}
```
I think we should also add the other 2 fields as well:
```
export interface QueryCommand extends Command {
  query: Record<string, any>,
  maxResultsPerPage?: number,
  pageNumber?: number
}
```
thanks

jamiehewitt15: Yeah, good point. Thanks for the review

paulo-ocean (approved):
lgtm

PR #374, opened 2024-03-25
Tests for malformed storage content

### LINKED ISSUE #358: Test service files validation for URL type
Test service files validation to have URL format if the storage type is URL.

mariacarmina:
Fixes #358  .

Changes proposed in this PR:

- add validations for arweave and ipfs to ensure that the format is the correct one
- add tests for malformed storage content

### DIFF
--- a/src/components/storage/index.ts
+++ b/src/components/storage/index.ts
@@ -26,6 +26,7 @@ export abstract class Storage {
   abstract getDownloadUrl(): string
   abstract fetchSpecificFileMetadata(fileObject: any): Promise<FileInfoResponse>
   abstract encryptContent(encryptionType: 'AES' | 'ECIES'): Promise<Buffer>
+  abstract isFilePath(): boolean
 
   getFile(): any {
     return this.file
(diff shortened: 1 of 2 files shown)


src/components/storage/index.ts
```diff
@@ -187,12 +192,13 @@ export class UrlStorage extends Storage {
   }
 
   isFilePath(): boolean {
-    const regex: RegExp = /^(.+)\/([^/]+)$/ // The URL should not represent a path
+    const regex: RegExp = /^(.+)\/([^/]*)$/ // The URL should not represent a path
     const { url } = this.getFile()
+    CORE_LOGGER.logMessage(`url: ${url}`)
```
jamiehewitt15, line 197: This log seems unnecessary, we should avoid creating too much noise in the logs. Also, until https://github.com/oceanprotocol/ocean-node/pull/368 is merged, logging out the URL leaks it out for anyone to see because the logs are still available on a public API endpoint. Even when access to the logs are restricted, we should avoid logging the URL / IPFS has / arweave txid in production code.

mariacarmina: I missed that when debugging, will remove it.

paulo-ocean (approved):
lgtm

alexcos20 (approved):

PR #375, opened 2024-03-26
Fix /state endpoint logic

### LINKED ISSUE #356: Fix 'state' endpoint
'State' command is implemented in a wrong manner, doing something else than it should. 
Let's fix it.

Purpose:
 - have a history for each did, identifying if the asset was successful indexed or we had some errors

Examples:
 - did with errors: https://v4.aquarius.oceanprotocol.com/api/aquarius/state/ddo?did=did:op:38fa9341749d1628a86d782013d1b6bb2fbd7aafe07be97cf96985d9829f8e59
 - successful did:  https://v4.aquarius.oceanprotocol.com/api/aquarius/state/ddo?did=did:op:fa0e8fa9550e8eb13392d6eeb9ba9f8111801b332c8d2345b350b3bc66b379d5

Definition of output (some changes compared to Aquarius)
  - chainId  - chain id of the asset
  - did  = id of the asset
  - nft = nft address
  - txId = last tx that created/updated metadata
  - valid = if schema was valid  (we might have cases when ddo is published on chain, but it's validation against the schema fails
  - error = text describing the error (maybe validation, maybe decrypt, maybe remote provider error, etc)

bogdanfazakas:
Fixes #356 .

Changes proposed in this PR:

- add ddo state database logic
- 
-


src/@types/commands.ts
```diff
@@ -64,6 +64,10 @@ export interface QueryCommand extends Command {
   query: Record<string, any>
 }
 
+export interface QueryDdoStateCommand extends Command {
```
paulo-ocean, line 67: isn't this the same? why do we need additional `QueryDdoStateCommand`  with same definition ?
```
export interface QueryCommand extends Command {
  query: Record<string, any>
}

export interface QueryDdoStateCommand extends Command {
  query: Record<string, any>
}
```
On a different subject, the node is crashing at startup @bogdanfazakas 
thanks

bogdanfazakas: yep good point initially made it a bit different but realised we need it to be more felxible and ended up like the one above, will remove it.
merging dev and stuff and checking the node start as well 
ty

src/components/core/queryHandler.ts
```diff
@@ -35,3 +35,28 @@ export class QueryHandler extends Handler {
     }
   }
 }
+
+export class QueryDdoStateHandler extends Handler {
+  validate(command: QueryCommand): ValidateParams {
```
mariacarmina, line 40: Is QueryCommand valid for ddo state handler? If yes, we can extend QueryHandler instead of Handler. 
I think `query` field from QueryCommand was used for typesense query.

mariacarmina (commented):
Can we also have some tests for QueryDdoStateHandler, to see how you form the query for ddo state and what contains?

Thanks.

src/components/core/queryHandler.ts
paulo-ocean: its the same interface @mariacarmina , that's why we don't need both, just QueryCommand is enough

bogdanfazakas: yep good point, updated

mariacarmina (approved):
lgtm

jamiehewitt15 (approved):
Looks good to me

PR #376, opened 2024-03-26
Avoid DDOS attacks, rate limit API calls

### LINKED ISSUE #355: DDOS:  Rate limiter
Let's have a rate limiter in core handlers.
All handlers should check the rate limiter first, to decide if they are going to serve the request, or deny it.

How to identify the other party?
 - for P2P, take remote peerID
 - for HTTP, take X-Forwarded-For header and remote ip

Limiters:
  - allow max env MAX_REQ_PER_SECOND requests from a remote party
  - have blacklist (defined as env)

paulo-ocean:
Fixes #355 

Changes proposed in this PR:

- Add rate limit (number of requests per second). Configurable.. we can limit by IP or peer ID
- Add blacklist. Configurable. List of blocked IPs or peer IDs
-


src/test/unit/networking.test.ts
```diff
@@ -91,3 +99,105 @@ describe('Test available network interfaces', () => {
...
+  })
+
+  it('Test rate limit per IP, on handler, different IPs', async () => {
+    // need to set it here, on a running node is done at request/middleware level
+    // none will be blocked, since its always another caller
+    const ips = ['127.0.0.2', '127.0.0.3', '127.0.0.4', '127.0.0.5']
+
+    const rateLimitResponses = []
+
+    for (let i = 0; i < 4; i++) {
```
bogdanfazakas, line 187: would be bettter ips.lenght in case we add one address more we may miss updating this hardcoding value

src/components/httpRoutes/requestValidator.ts
```diff
@@ -0,0 +1,47 @@
...
+import { getConfiguration } from '../../utils'
+import { HTTP_LOGGER } from '../../utils/logging/common'
+
+// TODO we should group common stuff,
+// we have multiple similar validation interfaces
+export interface CommonValidation {
+  valid: boolean
+  error?: string
+}
+const configuration = await getConfiguration()
```
bogdanfazakas, line 11: having this here may pose a problem but only for tests i would say, in case we are going to import this in one of our tests before having an '`PRIVATE_KEY`' ENV variable already configured this will be null.
there can be issues further on with the `configuration.blackList?` for eg trying to acces a property for an undefined null object.

bogdanfazakas (approved):
lgtm, besides some small tweaks think we can go further with it
great job  👍

src/components/httpRoutes/requestValidator.ts
paulo-ocean: yeah, i think i can move it inside checkIP anyway

src/@types/OceanNode.ts
```diff
@@ -8,6 +8,12 @@ export interface OceanNodeDBConfig {
   url: string | null
 }
 
+// blacklisted peer ids and ips (for rate limiting purposes)
+export interface BlackList {
```
jamiehewitt15, line 12: We've always used the terms allow list / deny list for all access control. I suggest doing the same here:

```suggestion
export interface DenyList {
```

paulo-ocean: really?  ok :-)

src/@types/OceanNode.ts
```diff
@@ -50,6 +56,8 @@ export interface OceanNodeConfig {
   assetPurgatoryUrl: string
   allowedAdmins?: string[]
   codeHash?: string
+  rateLimit?: number
+  blackList?: DenyList
```
jamiehewitt15, line 60: It still says blacklist here and in other places. @trentmc was very keen on avoiding using this term back when we first started working on access control

paulo-ocean: ok, sure, will refactor

jamiehewitt15:
Can you pull the latest changes from develop, the test should pass now

paulo-ocean:
> Can you pull the latest changes from develop, the test should pass now

yep, doing that

jamiehewitt15 (approved):
Looks good to me

PR #377, opened 2024-03-26
Provider fees fix

### LINKED ISSUE #350: Security:   Provider fees hardcoded
Can I consume a dataset with lower fees than the user should have? 

Yes, you can create a smart contracts with 2 orders and 2 provider fees events in the same transaction -> ocean node will take the first event of provider fees, which is the cheaper one because https://github.com/oceanprotocol/ocean-node/blob/develop/src/components/core/utils/feesHandler.ts#L169 hardcodes providerFee and it always takes the 1st one.

Let's fix this by iterating all providerFees events in the tx and check contract address, payer, consumer, etc

jamiehewitt15:
Fixes #350 

Changes proposed in this PR:

- Looping through events to make sure that we use the correct provider fee.


src/components/core/utils/feesHandler.ts
```diff
@@ -150,23 +151,48 @@ export async function verifyProviderFees(
...
+      ProviderFeesEvent = event
+      break // Exit the loop if a valid event is found
+    }
+  }
+
   // check provider address
-  if (
-    ProviderFeesEvent[0].args[0].toLowerCase() !== providerWallet.address.toLowerCase()
-  ) {
+  if (ProviderFeesEvent.args[0].toLowerCase() !== providerWallet.address.toLowerCase()) {
```
paulo-ocean, line 187: Hi @jamiehewitt15 
i'm having some doubts trying to understand what we have here on this function.
We are looping through the events, but at the end seems we only handle one
If we find something here:
```
CORE_LOGGER.logMessage('ProviderFee event found')
      ProviderFeesEvent = event
      break // Exit the loop if a valid event is found
```
Shouldn't we be validating them all? and only reply with an 'OK' if everything is valid?
If we have 4 events, with 3 invalid and 1 valid... we are only counting with the valid one...
It seems to me that as soon as we find one that is good, we pretty much take everything else as correct, per the response we send back

And if we don't have any valid this assignment: `ProviderFeesEvent = event` is never done which can break the node right here:
`if (ProviderFeesEvent.args[0].toLowerCase() !== providerWallet.address.toLowerCase()) {`


Also... i see that we set `errorMsg` in several places but on the return object (of type `ProviderFeeValidation`)
we always have '`isValid`' set to 'true' , so regardless of any validation errors i can always download the asset?
Something doesn't look good to me here..
thanks

src/components/core/utils/feesHandler.ts
```diff
@@ -138,90 +138,94 @@ export async function verifyProviderFees(
   computeEnv?: string,
   validUntil?: number // only for computeEnv
 ): Promise<ProviderFeeValidation> {
-  let errorMsg = null
-  if (!txId) errorMsg = 'Invalid txId'
+  if (!txId) {
+    CORE_LOGGER.logMessage('Invalid txId')
```
paulo-ocean, line 142: minor stuff, but we should not log error messages as normal ones..
we could use CORE_LOGGER.`error`() for instance..
If we use  CORE_LOGGER.logMessage() without any other arguments, it will use the default level INFO ... and error messages will not be distinguished from normal ones

paulo-ocean: hi @jamiehewitt15 , make sure you have updated your branch with the lastest from 'develop'.. 
not sure if is just this, or there is something else going on, but seems the node does not start from your branch:
```
npm run build
npm run start (fails)
```

thanks

paulo-ocean (approved):
lgtm

bogdanfazakas (approved):

PR #379, opened 2024-03-26
fix cron types, node breaks at startup

### LINKED ISSUE #378: Cron schedule is broken
missing node-cron types..
the call to `scheduleCronJobs` is breaking the node at startup

paulo-ocean:
Fixes #378 

Changes proposed in this PR:

- fix missing cron types
- 
-

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -79,6 +79,7 @@
         "@types/lzma-native": "^4.0.4",
         "@types/mocha": "^10.0.4",
         "@types/node": "^20.10.5",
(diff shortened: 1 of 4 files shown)


bogdanfazakas (approved):
lgtm, the failing unit test should not be related to this

jamiehewitt15 (approved):

paulo-ocean:
> lgtm, the failing unit test should not be related to this

no, its not not but i'm taking care of that annoying test once and for all :-D

PR #385, opened 2024-04-04
Added validations for processing event from chain MetadataCreated & MetadataUpdated

### LINKED ISSUE #347: Metadata Created & Updated:  validations
During processing of MetaData Created, let's validate:
 - ddo.id  = sha256(nftAddress + chainId)
 - ddo hash matches whatever is stored on chain event (as long as chain ddo hash ! =0)

mariacarmina:
Fixes #347  .

Changes proposed in this PR:

- added check for did in decrypt ddo handler
- added checks for did and ddo hash after decrypt
- test


src/components/Indexer/processor.ts
```diff
@@ -293,16 +305,33 @@ export class MetadataEventProcessor extends BaseEventProcessor {
...
         chainId,
         event.transactionHash,
-        decodedEventData.args[5],
-        decodedEventData.args[4]
+        decodedEventData.args[5], // document hash
+        metadata
       )
+      if (
+        ddo.id !==
+        'did:op:' +
```
bogdanfazakas, line 321: can you create a helper method for this check we have it already 3 times only in this file,  and one for the did creation will be usefull we have this exact hash creation in multiple places, maybe you can use this or move it to more generic place like a generic util method.

src/components/core/handler/ddoHandler.ts
```diff
@@ -306,6 +307,25 @@ export class DecryptDdoHandler extends Handler {
         }
       }
 
+      // did matches
+      const ddo = JSON.parse(decryptedDocument.toString())
+      if (
+        ddo.id !==
```
bogdanfazakas, line 313: same as above

mariacarmina:
Tested on op sepolia => when checking ddo hash after decryptDDO(), assure that the ddo is unencrypted (flag !== 2) too.

jamiehewitt15 (approved):
Looks good to me

src/components/httpRoutes/validateCommands.ts
```diff
@@ -38,13 +38,25 @@ export function validateCommandParameters(
...
   }
 
-  CORE_LOGGER.info(
-    `Checking received command data for Command "${commandStr}": ${JSON.stringify(
-      commandData,
-      null,
-      4
-    )}`
-  )
+  if (commandStr === 'encrypt') {
```
paulo-ocean, line 41: This could be much simpler..
both `CORE_LOGGER.info` lines are pretty much the same:
you could simply do something like:
```
const logCommandData = commandData
  if (commandStr === PROTOCOL_COMMANDS.ENCRYPT) {
    logCommandData.files = [] // hide files data for logging
  }
  CORE_LOGGER.info(
    `Checking received command data for Command "${commandStr}": ${JSON.stringify(
      logCommandData,
      null,
      4
    )}`
  )
```

paulo-ocean (approved):
lgtm

PR #360, opened 2024-03-19
Fix/ Indexer chunk size reduce

bogdanfazakas:
Fixes #354 #366 #388 .

Changes proposed in this PR:

- update reduce chunk size logic
- split retrieve and process logic for chunk size reduction
- added lock to loop so we wait util processing certain blocks before starting with new ones(that may start with the same block) 
- refactored the way we used last indexed block inside the indexer thread

PR #389, opened 2024-04-08
Fix/ integration test regarding ddo state 

### LINKED ISSUE #388: There is a failing integration test regarding ddo state
After the changes done for "state ddo" database, we have a test that  fails
"`should store the ddo state in the db with no errors and retrieve it using did`"
there is something wrong there, maybe a timeout or something, might worth a look

bogdanfazakas:
Fixes #388  .

Changes proposed in this PR:

- 
- 
-


bogdanfazakas:
think we can close this @paulo-ocean , was something off with barge as it seems, cause all was green locally and same in the branch i ve merge , also now green on main 🤯

paulo-ocean:
> think we can close this @paulo-ocean , was something off with barge as it seems, cause all was green locally and same in the branch i ve merge , also now green on develop after a re-run 🤯

Are you sure @bogdanfazakas ? cause it keeps failing on other branches even after a re-run.. its a bit strange that only one and the same test would be affected by barge... 
i think it might be some random thing instead

paulo-ocean:
this is definitely no ok, there is a bug somewhere, the same test keeps failing randomly , in multiple branches

bogdanfazakas:
> this is definitely no ok, there is a bug somewhere, the same test keeps failing randomly , in multiple branches

yes there is an issue with an asset beeing  already active from time to time when just published, investigating that

paulo-ocean:
> > this is definitely no ok, there is a bug somewhere, the same test keeps failing randomly , in multiple branches
> 
> yes there is an issue with an asset being already active from time to time when just published, think something related to things that got shared beetween test files investigating that

yeah, i suspect something similar, we are using the sample example DDOs in many places

bogdanfazakas:
> > > this is definitely no ok, there is a bug somewhere, the same test keeps failing randomly , in multiple branches
> > 
> > 
> > yes there is an issue with an asset being already active from time to time when just published, think something related to things that got shared beetween test files investigating that
> 
> yeah, i suspect something similar, we are using the sample example DDOs in many places

think i found it, let me test it more, maybe related to indexer loop

bogdanfazakas:
closing this, fixed in #360

PR #390, opened 2024-04-09
small tweak, allow own node to decrypt ddo (if list is set and missing own address)

### LINKED ISSUE #367: Node cannot decrypt DDO when processing Metadata events if address is not on AUTHORIZED_DECRYPTERS
On `MetadataEventProcessor.processEvent()`
we try to decrypt the DDO if the flags are '0x02' which means encrypted DDO
and we proceed if the nodeId matches the decryptorURL
```
if (nodeId === decryptorURL) {
          const decryptDDOTask: DecryptDDOCommand = {
            command: PROTOCOL_COMMANDS.DECRYPT_DDO,
            transactionId: txId,
            **decrypterAddress: keys.ethAddress,**
            chainId,
            encryptedDocument: metadata,
            documentHash: metadataHash,
            dataNftAddress: contractAddress,
            signature,
            nonce
          }
```
And we specify the `decrypterAddress` as the current node address on the command
But then the handler for the `DecryptDDOCommand` might fail if we have not set the current node address on the `AUTHORIZED_DECRYPTERS` array (if this is set)
But on the other hand it succeeds if `AUTHORIZED_DECRYPTERS` is not set at all

This is the part where the valid address is excluded:
```
if (
        config.authorizedDecrypters.length > 0 &&
        !config.authorizedDecrypters.includes(decrypterAddress)
      ) {
        CORE_LOGGER.logMessage('Decrypt DDO: Decrypter not authorized', true)
        return {
          stream: null,
          status: {
            httpStatus: 403,
            error: 'Decrypt DDO: Decrypter not authorized'
          }
        }
      }
```

paulo-ocean:
Fixes #367 

Changes proposed in this PR:

- If `authorizedDecrypters` is not empty, but the own address is not part of the list, indexer cannot decrypt DDO
- on the other hand, if nothing is set, it can
- FIX: if own address is not on the list, but decrypter address is own node, allow it


bogdanfazakas (approved):
lgtm

jamiehewitt15 (approved):
Looks good to me

PR #403, opened 2024-04-24
fix integration test

### LINKED ISSUE #401: Bug in download integration test
**Describe the bug**
There is a bug in `/integration/download.test.ts`. [Here](https://github.com/oceanprotocol/ocean-node/blob/develop/src/test/integration/download.test.ts#L288) it is missing await. That whole `it` block is not doing what it should.

paulo-ocean:
Fixes #401 .

Changes proposed in this PR:

- fix integration test (publish asset with credentials & deny download from another consumer)


jamiehewitt15 (approved):
Looks good to me

bogdanfazakas (approved):

PR #411, opened 2024-05-02
Dashboard: display index queue

### LINKED ISSUE #290: Dashboard:  Display the current indexing queue
Will use endpoint created by https://github.com/oceanprotocol/ocean-node/issues/289

jamiehewitt15:
Fixes #290 

Changes proposed in this PR:

- The dashboard now shows the indexing queue: 
image

### To Test:

1. `npm run build && npm run start`

2. Add a transaction to the indexing queue. send a POST request to the direct command endpoint: `http://127.0.0.1:8000/directCommand`

Example body:

```
{
    "command": "reIndex",
    "chainId": 137,
    "txId": "0xc53ba79bdcdba60c466c401631edc97383c230e1fdd1a734737cdab033446923"
}
```

3. You should see the transaction appear in the indexing queue on the dashboard.


paulo-ocean (approved):
LGTM!
not really sure if we should poll the queue every 2 seconds, might be too much

dashboard/src/components/IndexQueue.tsx
```diff
@@ -0,0 +1,77 @@
...
+          }))
+          setQueue(transformedQueue)
+        })
+        .catch((error) => {
+          console.error('Error fetching queue:', error)
+        })
+    }
+
+    fetchQueue() // Initial fetch
+    const intervalId = setInterval(fetchQueue, 2000) // Poll API every 2000 milliseconds (2 seconds)
```
bogdanfazakas, line 37: Maybe worth using the interval time that we have for indexer processing blocks here, we have an env var (process.env.INDEXER_INTERVAL) for that as well.

bogdanfazakas (approved):
lgt,  just one small suggestion but not mandatory

dashboard/src/components/IndexQueue.tsx
jamiehewitt15: Yeah, seems like a good idea. I've updated it

PR #412, opened 2024-05-03
add optional start block on rpcs

### LINKED ISSUE #400: Store startBlock in RPCS env var
Store startBlock in RPCS env var to let the user define from where the indexing should start.

paulo-ocean:
Fixes #400 .

Changes proposed in this PR:

- add optional `startBlock` on RPCS (for start crawling and reindex purposes)
- If startBlock is present and valid consider it instead of deployBlock
Other fixes:
- Some fixes on the DB functions to delete all assets, minor tweaks on the crawler thread
- remove static methods on Indexer (make sure we use the unique Indexer instance from the ocean node singleton)
- Refactor tests and indexer instances where needed (we had multiple). On tearDown() stop all worker threads (avoid multiple indexers and threads that kept running between different test suites). Make sure the node singleton is correctly updated with any new instance created


src/components/database/index.ts
```diff
@@ -567,18 +567,22 @@ export class DdoDatabase {
...
-          return response
-        }
+
+        DATABASE_LOGGER.debug(
+          `Number of deleted ddos on schema ${schema} : ${response.num_deleted}`
+        )
+
+        numDeleted += response.num_deleted
       } catch (error) {
+        console.log('got error', error)
```
mariacarmina, line 585: I think we can remove this console.log and keep only since it logs each error:
```javascript
DATABASE_LOGGER.logMessageWithEmoji(
            `Error when deleting DDOs from schema ${schema.name}: ` + error.message,
            true,
            GENERIC_EMOJIS.EMOJI_CROSS_MARK,
            LOG_LEVELS_STR.LEVEL_ERROR
          )
```

src/utils/constants.ts
```diff
@@ -75,6 +75,12 @@ export const EVENTS = {
   DISPENSER_CREATED: 'DispenserCreated'
 }
 
+export const INDEXER_CRAWLING_EVENTS = {
+  CRAWLING_STARTED: 'crawlingStarted',
+  REINDEX_QUEUE_POP: 'popFromQueue'
+  // TODO REINDEX_CHAIN
```
mariacarmina, line 81: This is for issue #414, right? When worker thread communicates back with parent port for admin handler notification?

mariacarmina (commented):
nice work!

src/utils/constants.ts
paulo-ocean: yes, exactly

mariacarmina (approved):
lgtm!

bogdanfazakas (approved):

Issue #314, opened 2024-03-04
Dashboard: Allow admin to force reindex some tx


alexcos20:
The admin needs to be able to force node to reindex a tx.

Input Params:
 - chain id
 - txId
 

Will use endpoint exposed by #313

PR #416, opened 2024-05-08
Dashboard: re-indexing chain & transaction

### LINKED ISSUE #298: Clean and restart indexing chain
The admin needs to be able to force restart indexing a chain. The assets on that chain will be deleted and indexing will start from the beginning. The other chains are unaffected
Params: chain id

Will use endpoint exposed by #297

jamiehewitt15:
Fixes #298 
Fixes: #314 

Changes proposed in this PR:

- Allowing users to re-index the chain and individual transactions from the dashboard


dashboard/next.config.js
```diff
@@ -1,5 +1,10 @@
 /** @type {import('next').NextConfig} */
+require('dotenv').config()
+
 const nextConfig = {
+  env: {
+    RPCS: process.env.RPCS || '{}'
```
paulo-ocean, line 6: hi @jamiehewitt15 
shouldn't we get this from the node status instead? this is already parsed on the node config.
besides, there are a few issues if the env var is not correct JSON. Check this issue here for reference:
[https://github.com/oceanprotocol/ocean-node/issues/419](url)
(it doesn't even build the dashboard if the process.env.RPCS is set with a wrong JSON).. i will address that issue with the config but i think we should get this info from directCommand "status" as well.

There is also another issue, if we have multiple RPCS/networks the selection boxes on the UI do not display anything (attaching screenshots)
thanks
Screenshot from 2024-05-10 09-47-03
Screenshot from 2024-05-10 09-45-54

jamiehewitt15: Thanks for the feedback. I've updated it so that we use the networks provided by the status endpoint. It's displaying multiple networks ok for me, maybe that was something caused by the env... Can you have another look and see if it's displaying correctly? I'll also have a look into the indexing queue issue. 

Screenshot from 2024-05-13 11-43-51

paulo-ocean: The display of the networks is OK now...  :+1: 
The other thing is just UI perception, when queue is empty...the way the message is displayed looks like the indexing queue is only associated with the first network ("Indexing queue is empty") is right bellow the first network.. maybe an empty line to add more space does the visual trick
Screenshot from 2024-05-13 11-45-09

paulo-ocean (approved):
lgtm, by the way
i actually handled the rpcs on indexQueue, while working on this one:
[https://github.com/oceanprotocol/ocean-node/pull/420](https://github.com/oceanprotocol/ocean-node/pull/420) 
not exactly sure if we should remove it completely, or use the status as i did there

bogdanfazakas (approved):
lgtm, tested locally all working as expected

mariacarmina (approved):
lgtm

PR #417, opened 2024-05-09
Feature/ Replace getDdo

### LINKED ISSUE #348: replace getDDO with findDDO
Right now, there are a lot of getDDO calls all over the code.
But if that asset is not indexed, or node has no index capabilities, a lot of requests will fail (imagine a download call).

So, let's refactor:
 - make sure findDDO searches local database first, then remote
 - if findDDO finds a newer version of the ddo, it should update our database
 - add a flag force to findDDO function (if that is set to TRUE, it will force a remote search, no matter what is stored in our node)
 - add force flag to p2p command and http endpoint
 - see if we need to refactor code so we can always pass did (string) or DDOs as arguments for all functions dealing with ddos, in order to reduce number of calls for findDDO

bogdanfazakas:
Fixes #348  .

Changes proposed in this PR:

- replaced some getDDO or ddo.retrieve with findDDO
- added force flag


src/components/core/handler/ddoHandler.ts
```diff
@@ -664,22 +665,25 @@ export class FindDdoHandler extends Handler {
...
-    try {
-      const ddo = await node.getDatabase().ddo.retrieve(ddoId)
-      return ddo as DDO
-    } catch (error) {
-      CORE_LOGGER.logMessage(
-        `Unable to find DDO locally. Proceeding to call findDDO`,
-        true
-      )
+    // First try to find the DDO Locally if findDDO is not enforced
+    if (!force) {
```
paulo-ocean, line 671: i think we should return `null` in case of error (inside the catch) so it matches the function signature
also, when we call this function we always look for a return value. Same on line 721. As is, this is a bit inconsistent...
here, we return nothing.. on the outer try/catch we re-throw the error again.
IMO, we should be consistent with the declared return types... and if we try/catch an error doesn't make much sense to throw another error that would force another try / catch somewhere up in the stack..
we would usually just want to use "throw new Error" when execution should not continue, not to pass the hot potato  to an upper component :-)

src/components/core/handler/ddoHandler.ts
```diff
@@ -714,8 +718,8 @@ export class FindDdoHandler extends Handler {
 
       return null
     } catch (error) {
-      CORE_LOGGER.logMessage(`Error getting DDO: ${error}`, true)
```
paulo-ocean, line 717: i think we should return null in case of error (inside the catch) so it matches the function signature, on line 721. As is, this is a bit inconsistent... (or move the return null to end of function)
here, we return nothing.. on the outer try/catch we re-throw the error again.
IMO, we should be consistent with the declared return types... and if we try/catch an error doesn't make much sense to throw another error that would force another try / catch somewhere up in the stack..
we would usually just want to use "throw new Error" when execution should not continue, not to pass the hot potato to an upper component :-)

bogdanfazakas: the idea of throwing an error here instead of returning null was brought by the fact the this method is now used for eg in the http handlers, (aquarius -> getDdo & getMetadata ) and in case of an error think it would be nice to return 500 and the proper error back to the client.
@paulo-ocean  based on your opinion after you read the above message i can make it back return null and return a generic error to the client or keep the error and make sure we are always handling that not just throwing it.

paulo-ocean: well, IMO its not consistent with the function signature` Promise<DDO | null>` so it either returns a DDO or null.
In this case the function itself is not consistent within "himself" ... sometimes it returns a DDO, other times it returns null, other times it throws an error :-)
I think all usages of `findAndFormatDdo` are checking for "something" on the result.. and if there is nothing (null), they are already handling the error and setting a proper message... some of them are even doing another "throw new Error" , so doing it again (and for nothing)...
Check this example for instance (on `componentts/c2d/index.ts`):  (there are others)
try {
    const algoDDO = await new FindDdoHandler(oceanNode).findAndFormatDdo(algoDID)
    if (!algoDDO) {
      throw new Error('Algorithm DDO not found')
    }
.....
} catch (error) {
    CORE_LOGGER.error(error.message)
    return checksums
  }
If you throw an error instead, the part:
```
if (!algoDDO) {
      throw new Error('Algorithm DDO not found')
    }
```
is not even executed
I'm also not sure about the 500 (but that's another topic :-)  )... usually a 500 is something really bad and not expected, IMO the most lame error code that exists... i would opt for a 404 when looking for some resource... but that's out of the scope here :-)

paulo-ocean (approved):
lgtm
btw, we just needed 1 "return null", after the catch block, not 2 (1 before + 1 after) , but OK :-)

alexcos20 (approved):

PR #421, opened 2024-05-13
Fix deletion of DDO's by filter & add delete in batches

### LINKED ISSUE #406: Use the batch_size parameter to control the number of documents.
Use the batch_size parameter to control the number of documents.
A larger value will speed up deletions, but will impact performance.
see when deleting all ddos from db with filter_by on chainId https://github.com/oceanprotocol/ocean-node/pull/332/files/28745f334c4a8cf4b6a5ef45998706cd7ee4d823#diff-a5b385b3226d31be17cd0f7df9a6a311b96ff62dc505120225fbcbd57d8a86ddR74

bogdanfazakas:
Fixes #406  .

Changes proposed in this PR:

- Fixed delete DDO by chainId function
- Add batch size param for delete DDO by chainId  

To test this:
1. I've published 1000 assets using ocean-cli

Screenshot 2024-05-30 at 10 12 22

2. Then triggered from the dashboard the reindex chain that is supposed to delete all assets and reindex all chain events

Screenshot 2024-05-30 at 10 09 32


paulo-ocean (approved):
lgtm

jamiehewitt15 (approved):
Looks good to me

bogdanfazakas:
Just tested with both having the field manually added with int32 type and also removing that field we manually added in the schemas and the delete query seems to work fine in both conditions if you send it the proper value and type, so decided to just remove that and and keep the schemas as we originaly had them.
Here a screens shot from today for reindexing a chain with around 1k ddo's.

Screenshot 2024-06-05 at 08 45 04

mariacarmina (approved):
great, let's merge it!

PR #445, opened 2024-05-15
Dashboard: get ocean peers polling

### LINKED ISSUE #440: Dashboard: Add polling for /getOceanPeers API request
Currently, the list of ocean node peers only updates when the page is refreshed. We should add polling so that it updates periodically by itself.

jamiehewitt15:
Fixes #440

Changes proposed in this PR:

- Polling the `/getOceanPeers` endpoint so that the UI updates the connected peer list every 2 minutes.


paulo-ocean (approved):
lgtm, 2 m should be ok

PR #449, opened 2024-05-15
fix p2p peers

alexcos20:
Changes proposed in this PR:

- fix p2p peers
- 
-


src/components/P2P/index.ts
```diff
@@ -428,22 +492,21 @@ export class OceanP2P extends EventEmitter {
...
+      // console.log('multiaddrs: ', multiAddrs)
+      await this._libp2p.contentRouting.provide(cid, multiAddrs)
+      // } else {
+      //  P2P_LOGGER.warn(
+      //    'Could not find any Ocean peers. Nobody is listening at the moment, skipping...'
+      //  )
+      //  // save it for retry later
+      //  // https://github.com/libp2p/js-libp2p-kad-dht/issues/98
+      //  if (!this._pendingAdvertise.includes(did)) {
+      //   this._pendingAdvertise.push(did)
```
paulo-ocean, line 508: i think we broke a feature here... If we publish a DDO and there are no peers connected at that moment, these will get lost for at least another 4 hours (when the REPUBLISH_INTERVAL_HOURS fires again)
This alone;  `await this._libp2p.contentRouting.provide(cid, multiAddrs)` will have no effect
this is the issue: [https://github.com/libp2p/js-libp2p-kad-dht/issues/98](https://github.com/libp2p/js-libp2p-kad-dht/issues/98)
also, there are some lint issues :-)
thanks

alexcos20: DHT should propagate itself, but it seems that there is a bug in dht implementation.
Anyway, I have reverted (and updated) the code, it will not hurt.

alexcos20:
Code wise is fine, I need to update docs and write a network tutorial

alexcos20:
ready for review

paulo-ocean (approved):
lgtm

docs/environment-variables.md
```diff
@@ -3,29 +3,40 @@
...
 - `P2P_dhtMaxOutboundStreams` : Maximum no of DHT outbound streams. Defaults to `500`
-- `P2P_mDNSInterval` : Interval (in ms) for discovery using mDNS.  Defaults to `20000` (20 seconds)
+- `P2P_mDNSInterval` : Interval (in ms) for discovery using mDNS. Defaults to `20000` (20 seconds)
 - `P2P_connectionsMaxParallelDials` : Maximum no of parallel dials. Defaults to `150`
-- `P2P_connectionsDialTimeout`: Timeout for dial commands.    Defaults to `10000` (10 seconds)
+- `P2P_connectionsDialTimeout`: Timeout for dial commands. Defaults to `10000` (10 seconds)
+- `P2P_ENABLE_UPNP`: Enable UPNP gateway discovery. Default: True
+- `P2P_ENABLE_AUTONAT`: Enable AutoNAT discovery. Default: True
+- `P2P_ENABLE_CIRCUIT_RELAY_SERVER`: Enable Circuit Relay Server. It will help network, but increase your bandwith usage. Should be disabled for edge nodes. Default: True
+- `P2P_ENABLE_CIRCUIT_RELAY_CLIENT`: Enable conections through relay servers. Default: True
```
jamiehewitt15, line 38: I'm not sure how we ended up with two files documenting the envs. These should be in `env.md` with the other environmental variables

README.md
```diff
@@ -318,3 +318,70 @@ npm run start
...
 
 The dashboard will be made available at: `http://localhost:8000/dashboard/`
+
+## Networking in cloud environments or DMZ
+
+In order for your node to join the network, the others nodes needs to be able to connect to it.
+All options can be controlled using [environment
+variables](docs/environment-variables.md#p2p)
+
+Here are the proper steps:
```
jamiehewitt15, line 328: Can we say something here about if it's ok just to keep the default values? For example: 

```suggestion
To quickly start your node, you can keep all of the default values. If you want a customised approach, here are the full steps:
```

docs/environment-variables.md
paulo-ocean: actually this file is the original one :-) you added the other much later

alexcos20: I will create a new issue for this

PR #454, opened 2024-05-16
Fix consume flow

### LINKED ISSUE #441: Make ProviderFees backwards compatible
Make ProviderFees backwards compatible.
cc @bogdanfazakas

bogdanfazakas:
Fixes #441 .

Changes proposed in this PR:

- Use proper blockchain provider & signer 
- fix initialize endpoint , make it backwards compatible 
-


src/utils/blockchain.ts
```diff
@@ -40,11 +40,24 @@ export class Blockchain {
     this.signer = new ethers.Wallet(process.env.PRIVATE_KEY, this.provider)
   }
 
-  public getSigner(): Signer {
+  public async getSigner(): Promise<Signer> {
+    if (
+      (await this.signer.getAddress()) === '0x0000000000000000000000000000000000000000'
```
paulo-ocean, line 45: hi @bogdanfazakas  
I know this is still a draft but  i think you´re going into a wrong direction
First, the node does NOT start without a private key,  and if you provide a wrong one you also get issues setting peer id on lib p2p... The check for address ZERO after creating the wallet, is useless IMO (node admin sets ZERO address?)

src/utils/blockchain.ts
```diff
@@ -40,11 +40,24 @@ export class Blockchain {
...
+      console.log('Signer not initialized')
+      this.signer = new ethers.Wallet(process.env.PRIVATE_KEY, this.provider)
+      return this.signer
+    }
     return this.signer
   }
 
   public getProvider(): JsonRpcApiProvider {
+    console.log('provider ready or not == ', this.provider.ready)
+    if (!this.provider.ready) {
```
paulo-ocean, line 56: Same thing here, IMO, this is pointless, if you want to check (some other place) if the provider/network is ready you should use the `isNetworkReady()` method..
but here, on the getter function it doesn´t make much sense. So we create a provider on the constructor, then.. if  when we call `getProvider()`  it is not ready, we're creating new instances with the same rpc and network? Why? What guarantees you that it will be ready after this? Nothing.. :-)
The problem you're trying to solve is not solved by changing anything on these getters :-)
tkx

src/components/c2d/index.ts
```diff
@@ -58,8 +58,13 @@ export async function getAlgoChecksums(
     if (!algorithmService) {
       throw new Error('Algorithm service not found')
     }
+    const sanitizedServiceFiles =
```
paulo-ocean, line 61: hi @bogdanfazakas 
we have the same repeated code in multiple places. How about an utils function that sanitizes  this kind of string?
probably better than doing the same thing in many places :-)
thanks

src/components/core/compute/startCompute.ts
```diff
@@ -116,7 +121,15 @@ export class ComputeStartHandler extends Handler {
               }
             }
           }
-          const provider = await getJsonRpcProvider(ddo.chainId)
+          const config = await getConfiguration()
+          const { rpc, network, chainId, fallbackRPCs } =
+            config.supportedNetworks[ddo.chainId]
+          const blockchain = new Blockchain(rpc, network, chainId, fallbackRPCs)
+          const delay = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms))
+          while (!(await blockchain.isNetworkReady()).ready) {
```
paulo-ocean, line 129: hi @bogdanfazakas 
what happens if we can't get a connection ready? Do we block here forever? IMO, this is not ok, infinite loops rarely are :-) ...
If we can't have a network/RPC connection in 2, 3, 4 seconds, maybe we should log the error and return an appropriate response
tkx

src/components/core/handler/ddoHandler.ts
```diff
@@ -159,6 +159,10 @@ export class DecryptDdoHandler extends Handler {
         supportedNetwork.chainId,
         supportedNetwork.fallbackRPCs
       )
+      const delay = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms))
```
paulo-ocean, line 162: same thing here... maybe you just need to check if` isNetworkReady()` ? 
Honestly i think we might be overcomplicating this a bit... we do have a valid provider and signer using the blockchain class, if the operation "requested" fails because there is an RPC or network error (because its not ready)  we probably should just log an return a proper error response, its not exactly the same thing as trying to initially do the  setup for the indexer thread

mariacarmina:
For compute the decrypt step succesed using this branch.

src/components/c2d/index.ts
bogdanfazakas: makes sense ty

src/utils/blockchain.ts
```diff
@@ -41,10 +41,20 @@ export class Blockchain {
...
+    if (!this.signer || typeof this.signer.signMessage !== 'function') {
+      CORE_LOGGER.warn('Signer is not initialized or not a valid ethers Signer')
+      this.signer = new ethers.Wallet(process.env.PRIVATE_KEY, this.provider)
+      return this.signer
+    }
     return this.signer
   }
 
   public getProvider(): JsonRpcApiProvider {
+    if (!this.provider || typeof this.provider.getNetwork !== 'function') {
```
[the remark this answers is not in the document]
bogdanfazakas: yep makes sense, fixed 
thanks

src/utils/blockchain.ts
```diff
@@ -41,10 +41,16 @@ export class Blockchain {
   }
 
   public getSigner(): Signer {
+    if (!this.signer || typeof this.signer.signMessage !== 'function') {
```
paulo-ocean, line 44: hey @bogdanfazakas you can safely remove this.
I think you still don't trust the change :-)

Its impossible to not have a signer and not have a provider instance (we're creating the corresponding instances on the constructor), so they will never be undefined or null.. So the IF block is not necessary... Also you're looking for a specific function(s) to be defined on each one... I don't  understand the purpose of this, unless you want to filter specific ethers versions  where these are present/absent (but i don't think these are a concern)
thanks

bogdanfazakas: nope think we can remove them if we always check that network ready before trying to use them.
Thanks

paulo-ocean: the fact that the network is ready or not has `nothing` to do with having the instances, or checking a function existence, this is just wrong. The IF is pointless as the code will never be executed :-)

paulo-ocean (approved):
lgtm

jamiehewitt15 (approved):
Looks good to me

PR #455, opened 2024-05-16
set option for log level on process.env

### LINKED ISSUE #451: As a node admin i should be able to set "default" LOG level
When i start a node i can set my own "default" log level
otherwise it picks some level by default

paulo-ocean:
Fixes #451.

Changes proposed in this PR:

- allow configure LOG_LEVEL on .env variable . Use that or default


jamiehewitt15 (approved):
Looks good to me

PR #457, opened 2024-05-20
Issue 397 warn env db logs

### LINKED ISSUE #397: get Logs API (specially from admin dashboard) should consider NODE_ENV setting
By default the `NODE_ENV`, if not existing or otherwise specified,  is considered '`development`'.
While running on 'dev' mode logs are only written to console or files.
In order to save logs on Database we have to remember to set NODE_ENV to '`production`'
if we forget about this,... the logs endpoints will not return anything (just empty stuff)
So we probably need to at least warn (or eventually write something on the response) to make sure the node admin is alerted to this possibility

paulo-ocean:
Fixes #397 

Changes proposed in this PR:

- Log/WARN that NODE_ENV should NOT be "development" for DB transport (logs saved to DB instead of console + file only)
- works mainly as a reminder for node admin
-


jamiehewitt15 (approved):
Looks good to me

mariacarmina (approved):
lgtm

PR #459, opened 2024-05-20
Remove hardcoded code for root endpoint.

### LINKED ISSUE #435: Remove hardcoded supported chain IDs from '/'
We should remove the hardcoded supported chains from '/' https://github.com/oceanprotocol/ocean-node/blob/develop/src/components/httpRoutes/rootEndpoint.ts. Check status command for it or just show the endpoints.

mariacarmina:
Fixes #435  .

Changes proposed in this PR:

- complete the endpoints for `/`
- fix chain ids list
- tested on remote VM


mariacarmina:
Response from root endpoint now:

Screenshot 2024-05-21 at 15 41 03

src/OceanNode.ts
```diff
@@ -23,7 +23,8 @@ export class OceanNode {
     private db?: Database,
     private node?: OceanP2P,
     private provider?: OceanProvider,
-    private indexer?: OceanIndexer
+    private indexer?: OceanIndexer,
+    private config?: OceanNodeConfig
```
paulo-ocean, line 27: we don't need config here at all, we already removed it in the past. Lets not do it again
tkx

paulo-ocean: this is enough:
```
rootEndpointRoutes.get('/', async (req, res) => {
```
no need to complicate

mariacarmina: Ok, I'll make it async then, thanks

jamiehewitt15:
This is good, but it would be even better if the route endpoints are also dynamically generated, rather than hardcoded in a json. As you've seen that list tends to be ignored and it will probably become out of date again soon. It would be nice if we don't have to maintain it.

paulo-ocean:
> This is good, but it would be even better if the route endpoints are also dynamically generated, rather than hardcoded in a json. As you've seen that list tends to be ignored and it will probably become out of date again soon. It would be nice if we don't have to maintain it.

I agree, this will get outdated again probably, but probably better do it on another PR

paulo-ocean (approved):
lgtm

jamiehewitt15 (approved):
Looks good to me, the endpoints will be handled in another PR

PR #464, opened 2024-05-22
First draft of job status response

### LINKED ISSUE #461: Communicate admin command results to dashboard (pooling)
This is a follow up of other PRs like #418  &  #431 
Since the dasboard is another app we can't listen to the events directly on the dashboard..
Also, as we're only serving/using static pages (not another node.js server) we don't have APIs or routes on the dashboard (so we can't call any API back either)

Implementing a more advanced messaging mechanism (queues, web sockets)  is pretty much impossible  task atm, unless we convert the dashboard to a fully node.js app (But currently the dashboard does not even build using other option than "output:export" on nextConfig, so that is another limitation.. and we probably don't need another full node.js app anyway

What we can do atm is have some sort of pooling mechanism to get the status of the latest admin commands...
When we issue the admin commands, we get a "job id" as part of the 200 response, so we can ask about his execution a few seconds later.. (right now we just send a random msg saying the command was delivered without follow up) 

When we catch/listen the events, we update the job status on the list of "pending execution"

Once the job is cleared/finished and dashboard has received that info at least once, the job is cleared permanently from the list

paulo-ocean:
Fixes #461 

Changes proposed in this PR:

- Implement a Job status mechanism (jobs queue) for long running commands (async commands with no direct/response, like currently the admin indexer / "reindex" ones are)
- Handler sends job info on the response (instead of just a "dummy" message)
- When actions complete (either with success or failure emit events &
- Modify received job status accordingly (delivered, pending execution, failed, success, etc), when they execute/complete/fail
- Created Get Job status API/endpoint
- Pool job status from dashboard and update UI/provide visual feedback accordingly

Screenshot from 2024-05-27 17-01-02
Screenshot from 2024-05-27 17-01-07
Screenshot from 2024-05-27 17-01-26


jamiehewitt15:
This looks good, I like the pooling solution. The Alert gets the message across in the UI nice, although I'm wondering if the Pending message would be better with an info status rather than a warning? The warning makes it look like something might have gone wrong.

paulo-ocean:
> This looks good, I like the pooling solution. The Alert gets the message across in the UI nicely, although I'm wondering if the Pending message would be better with an info status rather than a warning? The warning makes it look like something might have gone wrong.

Well, i see... but "pending" its not good or bad, it can still go both ways :-) 
I wanted to have different colors for different status, and considered there's only 4 by default, it looked appropriated, but i didn't wanted to go into much details on the UI side..
Any color suggestions? Maybe i can just easily change the default color for a given "severity"
tkx

jamiehewitt15:
> Any color suggestions?

Maybe yellow, rather than orange?

paulo-ocean:
> > Any color suggestions?
> 
> Maybe yellow, rather than orange?

I've update the color scheme.. i didn't went for the `yellow` because its hard to read with `white` text, and didn't wanted to override also different text colors for each status, seemed a bit too much...
So, used some `light grey ` for the "Pending" because its actually not doing anything... and updated the others to slight different shades of 'red', 'blue' and 'green'
tkx

jamiehewitt15 (approved):
Tested locally and it works well. The feedback to the user is a good addition

bogdanfazakas (approved):
lgtm

PR #465, opened 2024-05-23
Pretty console logs

alexcos20:
Our console logs are pretty much unreadable, displayed as objects.
image
image

Now, it looks more appealing:

image

### DIFF
--- a/src/utils/logging/Logger.ts
+++ b/src/utils/logging/Logger.ts
@@ -173,7 +173,7 @@ const format: winston.Logform.Format = winston.format.combine(
   winston.format.prettyPrint()
 )
 const alignedWithColorsAndTime: winston.Logform.Format = winston.format.combine(
-  winston.format.colorize(),
+  winston.format.colorize({ all: true }),
   winston.format.timestamp(),
   winston.format.align(),
   winston.format.printf(


alexcos20:
Now it looks even better:

image

alexcos20:
Tests are looking way better as well:

image

src/utils/logging/Logger.ts
```diff
@@ -461,8 +469,8 @@ export class CustomNodeLogger {
   ) {
     if (!level) level = this.getLoggerLevel() || getDefaultLevel()
 
-    let msg = includeModuleName ? this.buildMessage(message) : message
-
+    // let msg = includeModuleName ? this.buildMessage(message) : message
```
jamiehewitt15, line 472:
```suggestion
```

src/utils/logging/Logger.ts
```diff
@@ -476,9 +484,9 @@ export class CustomNodeLogger {
   buildMessage(message: string) {
     const cpName = this.getModuleName()
     if (cpName) {
-      message = '[' + cpName.toUpperCase() + '] => ' + message
+      // message = '[' + cpName.toUpperCase() + '] => ' + message
```
jamiehewitt15, line 487:
```suggestion
```

jamiehewitt15 (approved):
They're looking a lot better! I suggest removing the comments to avoid cluttering the code, otherwise it's all good

paulo-ocean (approved):
less crowded indeed  :+1: 
i would personally prefer to keep the color scheme on the text as well, as i think it can drive more attention than just white text, but that's just my own preference anyway :-)

alexcos20:
> less crowded indeed 👍 i would personally prefer to keep the color scheme on the text as well, as i think it can drive more attention than just white text, but that's just my own preference anyway :-)

done, see above commit

PR #467, opened 2024-05-23
Updating Env docs

### LINKED ISSUE #462: Document env vars
Somehow, we ended up with two files documenting the envs.

Let's keep only one

jamiehewitt15:
Fixes #462

Changes proposed in this PR:

- Updating Env docs
- Removing second env file


bogdanfazakas:
Based on our channel discussion think we need to update the env docs using the generally preferred style.

bogdanfazakas:
@jamiehewitt15 since you also removed the 2'nd env file from docs, please update the reference from this topic as well https://github.com/oceanprotocol/ocean-node/tree/env-docs?tab=readme-ov-file#environment-variables
thanks

bogdanfazakas (approved):
lgtm

PR #470, opened 2024-05-24
Reduce build time

### LINKED ISSUE #468: Reduce build time
To reduce the build time, we don't have to build the dashboard every time. We could first check if the dashboard has had any changes, if not then there is no need to build it again.

jamiehewitt15:
Fixes #468 

Changes proposed in this PR:

- Static dashboard files are added to the git repo, so that the don't have to be built by the user
- Script has been added to check for changes in the dashboard. It is only rebuilt if there are changes.
- `npm run clean` script has been modified so that it doesn't delete the dashboard build files. 

Discussion point: We could also do the same thing for the node i.e. run a script that checks for changes, only build if something has changed.


paulo-ocean:
i have a few doubts/questions on this PR
why can't we just have "build:dashboard" optional on the package.json
IMO, we could have simply 2 entries (for example):
```
"build": "npm run clean && npm run build:tsc",
"build-all": "npm run clean && npm run build:tsc && npm run build:dashboard",
```
we use the 1st as we usually do, also on CI ... and when we want the dashboard as well we use the other other one. Seems much simpler than committing tons of chunk files to github, than having scripts to detect changes. For the node itself we probably always want to build on CI, only exception would probably be when committing only documentation or code comments (that we don't do much anyway :-) )...
tkx

jamiehewitt15:
> we could have simply 2 entries

We already have two commands `build` & `build:no-dashboard`. Your suggestion is to change the names and switch them around so that the default case is that the dashboard isn't built by default. 

That doesn't solve the problem for people who want to use the dashboard. The problem is that the dashboard takes a long time to build. The solution proposed here is: we don't have to build it every time, only build it if there are changes, and if there are no changes then it doesn't need to be built the static files can be run straight out of the repo.

paulo-ocean:
> > we could have simply 2 entries
> 
> We already have two commands `build` & `build:no-dashboard`. Your suggestion is to change the names and switch them around so that the default case is that the dashboard isn't built by default.
> 
> That doesn't solve the problem for people who want to use the dashboard. The problem is that the dashboard takes a long time to build. The solution proposed here is: we don't have to build it every time, only build it if there are changes, and if there are no changes then it doesn't need to be built - the static files can be run straight out of the repo.


"That doesn't solve the problem for people who want to use the dashboard. "  
Why do you think  that? 
If i want to use the dashboard i build it with "built-all" (or whatever name we choose), but that should be the exception not the rule... so IMO, the default build would be; without it

In any case, in my opinion it will never solve the problem, cause as long as we change something on the dashboard (regardless if i want to use it or not) it will always take some time to build, so we're not really reducing the dashboard build time... we're just either building or skipping the build... but with a very complex approach instead :-)

anyway, this is my opinion

jamiehewitt15:
> Why do you think that?

If you consider some people want to run a node, with a dashboard, and want it to all start together quickly. The simplest solution is providing them with the static dashboard files, so they don't need to build it.

> If i want to use the dashboard i build it with "build-all" (or whatever name we choose), but that should be the exception not the rule

Your solution amounts to telling people that if they want their node to start quickly they shouldn't run the dashboard. We can do better than this, simply by providing them with the static files.

paulo-ocean (approved):
ok, lets go with this solution then

jamiehewitt15:
@bogdanfazakas @mariacarmina @alexcos20 looking for a second review on this

bogdanfazakas (approved):

PR #474, opened 2024-05-27
Fix DecryptDdoHandler

FilipMasar:
Fixes edit DDO

When DDO metadata was updated, it was not updated in ocean-node


paulo-ocean (approved):
LGTM, 
nevertheless i think we should try to provide a better description of the issue, in this case is not critical since the change is minimal, but it looks like the issue is that we cannot "decrypt" a DDO when the transaction id provided was generated by some metadata "updated" event... and we're only expecting a metadata "created" event...

The provided description:
"When DDO metadata was updated, it was not updated in ocean-node" is a bit vague IMO, seems like we did not processed the event on the indexer/db or something, instead
thanks

PR #475, opened 2024-05-27
Fix FileInfoHandler

FilipMasar:
Fixes decrypt function in fileInfoHandler.

The "0x" prefix is not recognized by the Buffer.from method with the 'hex' encoding.


FilipMasar:
Closing because it's part of another PR already https://github.com/oceanprotocol/ocean-node/pull/454/files#diff-956bfe26272e9e8e9df0d65c5f60e65894abb754e6c47937e02956b913501f22R41

PR #479, opened 2024-05-28
Proper handle of boolean envs

alexcos20:
Changes proposed in this PR:

- add & use getBoolEnvValue function in config
- 
-

### DIFF
--- a/src/utils/config.ts
+++ b/src/utils/config.ts
@@ -63,10 +63,10 @@ function getBoolEnvValue(envName: string, defaultValue: boolean): boolean {
     return defaultValue
   }
   if (
-    process.env[envName] ||
-    process.env[envName] === 'true' ||
-    process.env[envName] === '1' ||
-    process.env[envName].toLowerCase() === 'yes'
+    process.env[envName] &&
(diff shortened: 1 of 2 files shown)


mariacarmina (approved):

src/utils/config.ts
```diff
@@ -58,6 +58,21 @@ function getIntEnvValue(env: any, defaultValue: number) {
   return isNaN(num) ? defaultValue : num
 }
 
+function getBoolEnvValue(envName: string, defaultValue: boolean): boolean {
+  if (!(envName in process.env)) {
+    return defaultValue
+  }
+  if (
+    process.env[envName] ||
```
paulo-ocean, line 66: this first check doesn't look good :-)
If i do this:
`export P2P_ENABLE_IPV4=false`
it will return `true` cause process.env[P2P_ENABLE_IPV4] evaluates to string

alexcos20: see my last commit

paulo-ocean (approved):
lgtm

bogdanfazakas (approved):

PR #482, opened 2024-05-29
Hide file object and log the type instead.

### LINKED ISSUE #447: Check if c2d exposes transactionId from arweave in publish/result logs
Check if c2d exposes transactionId from arweave in publish/result logs.
Create a test with publishing an Arweave asset and algo as well -> start compute on these.

mariacarmina:
Fixes #447  .

Changes proposed in this PR:

- Hide file object and log the type instead.


paulo-ocean (approved):
lgtm, although i think we could probably remove the log completely. 
not sure if logging this, adds any value

jamiehewitt15 (approved):
Looks good to me. Like Paulo said, it would also be fine to remove it altogether

bogdanfazakas (approved):

PR #484, opened 2024-06-03
Issue 311 - performance/load tests with grafana k6

### LINKED ISSUE #311: Write performance/load tests
We will use k6 for performance/load tests, it will be hosted locally in our infrastructure.

1. Research what k6 can do and figure out what we could easily implement. We want to test things like cpu/ram usage, response times, network usage, how many requests can the enpoints handle etc.
2. Propose a list of test with a rough estimation for each
3. Write tests

paulo-ocean:
Fixes #311 .

Changes proposed in this PR:

- Create load/performance tests (scripts on package.json). Later we might have a dedicated environment to run these 
- There are 3 types of tests according to the number or requests and virtual users used (`smoke`, `load` and `stress` tests)

To use:
```
npm run test:smoke
npm run test:load
npm run test:stress
npm run test:request:rate (runs under the supplied rate limit, must be less than node's default, to avoid triggering it)
```
The goal is to flood the node with HTTP requests, for a long period of time, reaching all possible endpoints multiple times, not testing any specific API logic (those are on integration/unit)

- Optionally, we can define a `TARGET_URL` ( a running node URL) otherwise it will try to run against the node on local machine
- Optionally, we can also  define a `TARGET_USERS` ( number of concurrent virtual users ) default is 5 for smoke, 100 for load, and 200 for stress (if you supply other value all these will adjust). During tests, the load/users/requests change as well, VUs make multiple requests, providing "peaks" of usage, sudden "increases" or "drops".

- Optionally, we can define a `RATE_LIMIT` (if we want to run it without triggering the node rate limitations). This value needs to be the less than the value that we have configured on the node itself)
-  While tests are running a dashboard can be accessed at http://127.0.0.1:5665/ . This provides live UI insigths on the tests execution. Unfortunately K6 only maintains this page while tests are being executed and we need to open it manually.
- At the end of the tests there are some reports on the console and an HTML report ( `html-report.html` ) is also generated with UI insigths, pretty much like the dashboard above

What test requests do:
- The test suite starts by reaching the root endpoint, then for each API returned it will will also reach those.
- For the "special" `directCommand` path it will also try all the available commands

 
Screenshot from 2024-06-05 23-16-24
Screenshot from 2024-06-05 23-16-07
Screenshot from 2024-06-05 23-15-47
Screenshot from 2024-06-05 23-15-18


src/test/performance/perf_test.js
```diff
@@ -0,0 +1,137 @@
...
+// cloud: {
+//   // The ID of the project to which the test is assigned in the k6 Cloud UI.
+//   // By default tests are executed in default project.
+//   projectID: "",
+//   // The name of the test in the k6 Cloud UI.
+//   // Test runs with the same name will be grouped.
+//   name: "script.js"
+// },
+
+// Uncomment this section to enable the use of Browser API in your tests.
```
jamiehewitt15, line 77: Is the Browser API useful? If so, it might be nice to have implemented via an env rather than uncommenting the code

src/test/performance/perf_test.js
```diff
@@ -0,0 +1,137 @@
...
+} from './util.js'
+// -----------------------------------------------------------------
+// LIST OF TESTS TO EXECUTE
+// -----------------------------------------------------------------
+
+// - Call node root enpoint (get a list of all endpoints)
+// - Call all HTTP endpoints (with & without proper params)
+// - Execute requests with & without RATE limits on the node instance
+// - Call directCommand enpoint with all supported commands
+//
```
jamiehewitt15, line 19: All of the files are nicely commented, there is nothing about them in the main readme though. It would be worth adding a short section about these tests in the readme as we also mention all the other types of tests there.

package.json
```diff
@@ -32,7 +32,12 @@
     "test:integration:light": "npm run build-tests && npm run mocha-light \"./dist/test/integration/**/*.test.js\"",
     "test:unit:cover": "nyc --report-dir coverage/unit npm run test:unit",
     "test:integration:cover": "nyc --report-dir coverage/integration --no-clean npm run test:integration",
-    "logs": "./scripts/logs.sh"
+    "logs": "./scripts/logs.sh",
+    "test:k6-common": "npm run build-tests && K6_WEB_DASHBOARD=true K6_WEB_DASHBOARD_EXPORT=html-report.html",
+    "test:smoke": "npm run test:k6-common k6 run -e TEST_TYPE=smoke \"./dist/test/performance/perf_test.js\"",
+    "test:load": "npm run test:k6-common k6 run -e TEST_TYPE=load \"./dist/test/performance/perf_test.js\"",
+    "test:stress": "npm run test:k6-common k6 run -e TEST_TYPE=stress \"./dist/test/performance/perf_test.js\"",
+    "test:request:rate": "npm run test:k6-common && k6 run -e RATE=true \"./dist/test/performance/perf_test.js\""
```
jamiehewitt15, line 40: Would you consider including any of these in the CI flow?

src/test/performance/perf_test.js
paulo-ocean: no, its not useful, i can remove the commented part though

src/test/performance/perf_test.js
paulo-ocean: yeah i can add something on the readme

package.json
paulo-ocean: the idea is to have a separate environment for these, not running on CI

src/test/performance/perf_test.js
jamiehewitt15: ok fair enough, may as well remove it if it's not useful

src/test/performance/perf_test.js
paulo-ocean: done

jamiehewitt15 (approved):
Nice, works well when runinng it locally. Each test taste about 30-40 seconds which seems very reasonable

jamiehewitt15:
When I can these commands it created a report `html-report.html` in the route of the repository. Can you add that to gitignore?

PR #485, opened 2024-06-03
Fix start compute

mariacarmina:
Fixes # .

Changes proposed in this PR:

- fix compute http requests, replace req.query with req.body
- conversion for chain id
- update valid until in start compute in engine class
- fix fileinfo route
- compatible with change [from ocean.js](https://github.com/oceanprotocol/ocean.js/pull/1821)
- create function for retrieving inet IP for indexer to be reachable from barge -> needs refactoring by completing with an FQDN


jamiehewitt15:
Is this still a work in progress or is it replaced by the other PRs?

mihaisc:
@mariacarmina what is the status on this?

mariacarmina:
I'll close this, since it was a debug PR for c2d, now c2d is fixed.

PR #491, opened 2024-06-10
Collect fees handler

### LINKED ISSUE #294: BE: Endpoint to allow admin to transfer collected fees
Ocean-node will collect fees  for different services (download, compute, etc).
Create an endpoint that will allow the admin to transfer those fees.

Input params:
 - token address  (if 0x0 = native token for that chain)
 - chainId
 - token amount
 - destination address

mariacarmina:
Fixes #294  .

Changes proposed in this PR:

- create handler for collecting provider fees as admin
- added tests
- tweak getProviderWallet() function


src/components/core/utils/feesHandler.ts
```diff
@@ -495,7 +505,7 @@ export async function getProviderFeeToken(chainId: number): Promise<string> {
   )
   if (result.length === 0 && chainId === 8996) {
     const localOceanToken = getOceanArtifactsAdresses().development.Ocean
-    return localOceanToken || ethers.ZeroAddress
```
paulo-ocean, line 498: Hi @mariacarmina 
 `localOceanToken` can still be null or undefined here, so we should return `ZeroAddress` in that case, which is basically as it was before
tkx

src/components/core/utils/feesHandler.ts
```diff
@@ -470,11 +471,20 @@ export async function checkFee(
  * @param chainId the chain id (not used now)
  * @returns the wallet
  */
-export async function getProviderWallet(chainId?: string): Promise<ethers.Wallet> {
-  const wallet: ethers.Wallet = new ethers.Wallet(
+export async function getProviderWallet(chainId: string = null): Promise<ethers.Wallet> {
+  if (chainId) {
```
paulo-ocean, line 475: Hi @mariacarmina 
I think we should at least check if the chainId is valid and supported (or we can do it before.. check the other comments)

Afaik, he original idea of this function was just to get the node **wallet address**, which is always the same for EVM chains, so the chain id would not be needed for that.

If we are "extending" the function usage, and want to perform operations with the wallet, we probably could be using the `utils/blockchain` class instead.. which already handles the provider, signer, etc..
So, we could probably refactor the function to create an instance of blockchain and get the wallet from there

thanks

src/components/core/admin/collectFeesHandler.ts
```diff
@@ -0,0 +1,96 @@
...
+import { getConfiguration } from '../../../utils/index.js'
+import { getProviderFeeToken, getProviderWallet } from '../utils/feesHandler.js'
+import { parseUnits, Contract, ZeroAddress } from 'ethers'
+import ERC20Template from '@oceanprotocol/contracts/artifacts/contracts/templates/ERC20Template.sol/ERC20Template.json' assert { type: 'json' }
+import { CORE_LOGGER } from '../../../utils/logging/common.js'
+import { Readable } from 'stream'
+
+export class CollectFeesHandler extends AdminHandler {
+  validate(command: AdminCollectFeesCommand): ValidateParams {
+    if (
```
paulo-ocean, line 19: Hi, 
We should be validating all mandatory parameters like the `chainId` as well like:
`if (!validateCommandParameters(command, ['chainId'])) {`

Imo, we don't need our own regex test for checking addresses, we can just use `isAddress()` function from ethers (we are using it in many other places as well)

Additionally, we should also check if the chainId is supported (not just provided as param) like:
`await checkSupportedChainId(task.chainId)` 
(check the reindex commands for instance)
thanks

src/components/core/admin/collectFeesHandler.ts
```diff
@@ -0,0 +1,96 @@
...
+      )} on a different node.`
+      CORE_LOGGER.error(msg)
+      return buildErrorResponse(msg)
+    }
+    const providerWallet = await getProviderWallet(String(task.chainId))
+    try {
+      const providerFeeToken = await getProviderFeeToken(task.chainId)
+      if (
+        task.tokenAddress.toLowerCase() !== providerFeeToken.toLowerCase() ||
+        task.tokenAddress === ZeroAddress ||
```
paulo-ocean, line 48: Hi 
I think we might be missing something else here.
From the issue definition:
```
Input params:
token address (if 0x0 = native token for that chain)
```
So, if the token address is the ZeroAddress, we should use the native token of the chain, not reject it
as far as i understand it, if the providerFeeToken is not set (ZeroAddress) and we also send the ZeroAddress, we should work with the native token of the chain
thanks

paulo-ocean (approved):
LGTM, 
some minor tweaks could still be done

src/components/core/utils/feesHandler.ts
```diff
@@ -470,11 +471,25 @@ export async function checkFee(
...
-  const wallet: ethers.Wallet = new ethers.Wallet(
+export async function getProviderWallet(
+  chainIdProvided: string = null
+): Promise<ethers.Wallet> {
+  if (chainIdProvided) {
+    // to avoid breaking change for other tests
+    const config = await getConfiguration()
+    const { rpc, network, chainId, fallbackRPCs } =
+      config.supportedNetworks[chainIdProvided]
+    const blockchain = new Blockchain(rpc, network, chainId, fallbackRPCs)
```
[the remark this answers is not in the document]
mariacarmina: ok, I'll add this, sure. Thanks!

src/components/core/admin/collectFeesHandler.ts
```diff
@@ -0,0 +1,118 @@
...
+    if (!checkChainId.validation) {
+      return buildErrorResponse(
+        `Chain ID ${task.chainId} is not supported in the node's config`
+      )
+    }
+    const providerWallet = await getProviderWallet(String(task.chainId))
+    try {
+      let providerFeeToken = await getProviderFeeToken(task.chainId)
+      if (task.tokenAddress === ZeroAddress) {
+        // for the moment I put Ocean token from address.json
```
alexcos20, line 64: if tokenAddress is 0x00, then you need to transfer native tokens (ETH, Matic, etc)

alexcos20 (changes requested):
see my comments

src/components/core/admin/collectFeesHandler.ts
```diff
@@ -0,0 +1,153 @@
...
+      const { rpc, network, chainId, fallbackRPCs } =
+        config.supportedNetworks[task.chainId]
+      const blockchain = new Blockchain(rpc, network, chainId, fallbackRPCs)
+      const provider = blockchain.getProvider()
+      const providerWallet = await getProviderWallet(String(task.chainId))
+      const ammountInEther = parseUnits(task.tokenAmount.toString(), 'ether')
+      const providerWalletAddress = await providerWallet.getAddress()
+
+      let receipt
+      if (task.tokenAddress === ZeroAddress) {
```
jamiehewitt15, line 74: Can you make sure that all the address comparisons are done in lowercase? 

```suggestion
      if (task.tokenAddress?.toLowerCase() === ZeroAddress) {
```

src/components/core/admin/collectFeesHandler.ts
```diff
@@ -0,0 +1,153 @@
...
+        return {
+          stream: null,
+          status: {
+            httpStatus: 404,
+            error: msg
+          }
+        }
+      }
+      if (receipt.status !== 1) {
+        const msg: string = `Reverted transaction: ${receipt}`
```
bogdanfazakas, line 123: isn't receipe here an object ? wouldn't it be better to add one the receipt properties to this message ?

src/components/core/utils/feesHandler.ts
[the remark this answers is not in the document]
bogdanfazakas: yeah agreed, but do we really need this method or we should always use the signer and provider from the blockchain class ?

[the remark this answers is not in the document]
paulo-ocean: > yeah agreed, but do we really need this method or we should always use the signer and provider from the blockchain class ?

i think we could use it as a "shortcut", for when we want the wallet, otherwise we will be repeating "new Blockchain()" part in a lot of places

src/components/core/admin/collectFeesHandler.ts
mariacarmina: tokenAddress is a mandatory parameter and validated that it's not missing, I'll add `toLowerCase()`, thanks.

src/components/core/admin/collectFeesHandler.ts
mariacarmina: put logs property and stringified

src/@types/commands.ts
```diff
@@ -116,6 +121,13 @@ export interface AdminReindexTxCommand extends AdminCommand {
   txId: string
 }
 
+export interface AdminCollectFeesCommand extends AdminCommand {
+  tokenAddress: string
+  chainId: number
+  tokenAmount: number
```
jamiehewitt15, line 127: Is it possible to make `tokenAmount` optional and make the default value the maximum available amount?

mariacarmina: I transfer the whole balance if token amount is not provided.

jamiehewitt15: Ok nice, thank you

jamiehewitt15 (approved):
Looks good to me

alexcos20 (approved):

PR #492, opened 2024-06-11
Issue 469 dynamic routes

### LINKED ISSUE #469: root endpoint should have dynamic "routes" instead of hardcoded ones
Right now we hardcode the routes that "we provide", which means everytime we change/add/remove a route, this gets outdated. So we need to make it dynamic... The routes should be pushed on some kind of list and we need to make sure the changes are always reflected on the response..
related with: 
[https://github.com/oceanprotocol/ocean-node/pull/459](https://github.com/oceanprotocol/ocean-node/pull/459)

paulo-ocean:
Fixes #469 

Changes proposed in this PR:

- root endpoint "`serviceEndpoints`" is now dynamic (custom names for existing routes maintained) . No need to update the json anymore
- we can still provide a "custom/normalized" name for a route (express does not support it out of box), but if not present, a default one is used

Just an example to clarify. Lets take all the` /compute `routes for instance (current response):
`computeStart`: ['POST', `${SERVICES_API_BASE_PATH}/compute`],
`computeStatus`: ['GET', `${SERVICES_API_BASE_PATH}/compute`],
`computeDelete`: ['DELETE', `${SERVICES_API_BASE_PATH}/compute`],
`computeStop`: ['PUT', `${SERVICES_API_BASE_PATH}/compute`],

They all point to` /api/services/compute`. For express the path is always the same, only thing different is the "method", if we don't supply a custom/pretty name (the `routesNames` mapping), we still get the dynamic routes, but we end with something like this on the response:
`/api/services/compute`: ['POST', `/api/services/compute`],
`/api/services/compute`: ['GET', `/api/services/compute`],
`/api/services/compute`: ['DELETE', `/api/services/compute`],
`/api/services/compute`: ['PUT', `/api/services/compute`],
Which is not very explanatory. We could also append the method name or something else to the "name" ( This "name" is the json field name) ..., but still not very pretty or explanatory.  But again, express does not have the concepts of route "names" or "descriptions", it just recognizes paths and methods... so, if we want, we have to do it manually, therefore the optional mapping... 

When doing a mapping from a "pretty/normalized" name to a route (**Not Mandatory**) we can have something more meaningful like:

`computeStart`: ['POST', `/api/services/compute`],
`computeStatus`: ['GET', `/api/services/compute`],
`computeDelete`: ['DELETE', `/api/services/compute`],
`computeStop`: ['PUT', `/api/services/compute`],

To summarize, when we `add/edit/delete` a route we DO NOT need to do anything else anymore.. `getAllServiceEndpoints()` will handle all automatically and always return everything available.. 
UNLESS we want a pretty name for the service endpoint, in that case we need to provide a value on the `routesNames`, otherwise it just gets the default (same name as path).


mariacarmina (approved):
LGTM!

src/components/httpRoutes/rootEndpoint.ts
```diff
@@ -14,50 +15,12 @@ rootEndpointRoutes.get('/', async (req, res) => {
...
-      // validateContainer: ['POST', `${SERVICES_API_BASE_PATH}/validateContainer`],
-      // create_auth_token: ['GET', `${SERVICES_API_BASE_PATH}/createAuthToken`],
-      // delete_auth_token: ['DELETE', `${SERVICES_API_BASE_PATH}/deleteAuthToken`],
-    },
+    serviceEndpoints: getAllServiceEndpoints(),
+    // Not yet implemented
+    // validateContainer: ['POST', `${SERVICES_API_BASE_PATH}/validateContainer`],
+    // create_auth_token: ['GET', `${SERVICES_API_BASE_PATH}/createAuthToken`],
+    // delete_auth_token: ['DELETE', `${SERVICES_API_BASE_PATH}/deleteAuthToken`],
+    // },
```
jamiehewitt15, line 23: We can also delete these, we will never uncomment them now that we have the `getAllServiceEndpoints` function

src/components/httpRoutes/rootEndpoint.ts
```diff
@@ -1,8 +1,9 @@
 import express from 'express'
-import { SERVICES_API_BASE_PATH } from '../../utils/constants.js'
-import { AQUARIUS_API_BASE_PATH } from './aquarius.js'
+// import { SERVICES_API_BASE_PATH } from '../../utils/constants.js'
+// import { AQUARIUS_API_BASE_PATH } from './aquarius.js'
```
jamiehewitt15, line 3: May as well delete these

jamiehewitt15 (approved):
Looks good to me

PR #496, opened 2024-06-12
Address comparison

### LINKED ISSUE #488: Convert and compare addresses in lowercase

jamiehewitt15:
Fixes #488

Changes proposed in this PR:

- Making sure that all address are in lower case before any comparisons


paulo-ocean (approved):
LGTM
altough you went to far on some :-)
for instance:
`ZeroAddress?.toLowerCase()`
is not needed, ZeroAddress is a constant, always defined and always lowercase, but anyway, doesn't hurt

alexcos20 (approved):

PR #502, opened 2024-06-19
Issue 501 - C2D getJobStatus - did vs documentId vs agreementId

### LINKED ISSUE #501: getComputeStatus "did" param mismatch "documentId" from SDK
On the node `getComputeStatus`  (either legacy route or directCommand) we expect a "did",
but if we actually supply one on the request we get empty array response.
The only way to actually get a result of the job status is if we DO NOT provide the "did"..
This is because on the node we expect a field named `did` but the SDK, that is used by the CLI for instance is actually calling the field `documentId`

HOWEVER, neither `did` or `documentId` should be provided, because no logic is done along the components regarding the DID. At the end the Operator Service is expecting an `agreementId` ( a transaction hash )

The SDK code (computeStatus):
```
 let url = `?consumerAddress=${consumerAddress}`
    url += (did && `&documentId=${this.noZeroX(did)}`) || ''
    url += (jobId && `&jobId=${jobId}`) || '
```
To note that the call to `this.noZeroX(did)` is actually meant to work on something else than a `did:op:` type of  string
This is sent to the provider, so on the provider, we need to read either `documentId` or `did`, which then calls Oparetor Service url

HOWEVER, on the Operator Service enpoint: 
```
def get_compute_job_status():
    """
    Get status for an specific or multiple jobs.
    ---
    tags:
      - operation
    consumes:
      - application/json
    parameters:
      - name: agreementId
        in: query
        description: agreementId
        type: string
```
we only expect an `agreementId` (the transaction id), so neither a `did` or `documentId`, cause its not an asset identifier but a transaction identifier. We expect One of; `jobId` , `agreementId` or `owner`

NEED ALSO FIXES:
- Cli [https://github.com/oceanprotocol/ocean-cli/issues/52](https://github.com/oceanprotocol/ocean-cli/issues/52)
- SDK [https://github.com/oceanprotocol/ocean.js/issues/1820](https://github.com/oceanprotocol/ocean.js/issues/1820)
- Provider [https://github.com/oceanprotocol/provider/issues/680](https://github.com/oceanprotocol/provider/issues/680)
- Operator service (API docs) [https://github.com/oceanprotocol/operator-service/issues/93](https://github.com/oceanprotocol/operator-service/issues/93)

They all read `documentId` but they do not use the field at all. They just forward it between them
so on the Final call from provider -> operator service  it is sent, 
BUT operator service only interprets something called; `agreementId`
So, all the flow involving the asset did is **useless**
All the 3 components above should be dealing with an  `agreementId`  instead , not a `documentId`

paulo-ocean:
Fixes #501 

Changes proposed in this PR:

- support for agreementId field
- 
-


src/components/core/compute/getStatus.ts
```diff
@@ -50,11 +50,12 @@ export class ComputeGetStatusHandler extends Handler {
         allC2dClusters = allC2dClusters.filter((arr) => arr.hash === hash)
         jobId = task.jobId.slice(index + 1)
       }
+      // TODO check
```
jamiehewitt15, line 53: What exactly is this TODO about?

paulo-ocean: just a self reminder, can remove it

jamiehewitt15 (approved):
Looks good to me

src/components/httpRoutes/compute.ts
```diff
@@ -150,8 +150,8 @@ computeRoutes.get(`${SERVICES_API_BASE_PATH}/compute`, async (req, res) => {
       command: PROTOCOL_COMMANDS.COMPUTE_GET_STATUS,
       node: (req.query.node as string) || null,
       consumerAddress: (req.query.consumerAddress as string) || null,
-      did: (req.query.did as string) || null,
```
mariacarmina, line 153: I think we can keep also did, but the correct name key is `documentId`, that's what is also used in production provider https://github.com/oceanprotocol/provider/blob/main/ocean_provider/utils/compute.py#L31. We can keep also agreementId and also documentId

mariacarmina (commented):
small comment on my side. otherwise it's fine.

src/components/httpRoutes/compute.ts
paulo-ocean: why is that ?  can you give me 1 good reason why we should keep a field that is not used anywhere? except perpetuating  a bug?

provider does nothing with that field (it could be null / not present as well), neither does operator service
[https://github.com/oceanprotocol/operator-service/blob/baa1095d8fc666dc9b73fbc954d8235210940d97/operator_service/routes.py#L353](https://github.com/oceanprotocol/operator-service/blob/baa1095d8fc666dc9b73fbc954d8235210940d97/operator_service/routes.py#L353)
there are 3 fields used to identify a job(s), owner, jobId, and agreementId
either `did` or `documentId` are useless

jamiehewitt15: If we're not using it let's just remove it. These things will likely change anyway when we work on C2D 2.0 and then we can revisit what parameters we need.

alexcos20 (approved):

PR #508, opened 2024-06-26
Issue 506 agreementid stop

### LINKED ISSUE #506: stop compute command, not dealing with optional `agreementId`
This is pretty much the same of:
[https://github.com/oceanprotocol/ocean-node/issues/501](https://github.com/oceanprotocol/ocean-node/issues/501)

From the **CLI** we pass a `did` as first argument, and a `jobId` as second
The `did`  is only used at CLI level (to check if the dataset exists)

Then we call **SDK**:
```
/** Instruct the provider to Stop the execution of a to stop a compute job.
   * @param {string} did the asset did
   * @param {string} consumerAddress The consumer address.
   * @param {string} jobId the compute job id
   * @param {string} providerUri The provider URI.
   * @param {Signer} signer The consumer signer object.
   * @param {AbortSignal} signal abort signal
   * @return {Promise<ComputeJob | ComputeJob[]>}
   */
  public async computeStop(
    did: string,
    consumerAddress: string,
    jobId: string,
    providerUri: string,
    signer: Signer,
    signal?: AbortSignal
...
```
  
we use the `did` on the message signature and we also set the field `documentId` from it as:
```
...
let signatureMessage = consumerAddress
    signatureMessage += jobId || ''
    signatureMessage += (did && `${this.noZeroX(did)}`) || ''
    signatureMessage += nonce
    const signature = await this.signProviderRequest(signer, signatureMessage)
    const payload = Object()
    payload.signature = signature
...
```
We send the request to Provider, 
On the **Ocean Node**, we do not check any `did` or `documentId` field(s):
```
const stopComputeTask: ComputeStopCommand = {
      command: PROTOCOL_COMMANDS.COMPUTE_STOP,
      node: (req.body.node as string) || null,
      consumerAddress: (req.body.consumerAddress as string) || null,
      signature: (req.body.signature as string) || null,
      nonce: (req.body.nonce as string) || null,
      jobId: (req.body.jobId as string) || null
    }
```
To Note that on existing Provider (python)  we also try to read a `documentId` instead
```
...
did = data.get("documentId")
    owner = data.get("consumerAddress")
    job_id = data.get("jobId")
...
```

When we call Operator Service, we do not use any of these fields [did, documentId] and we miss instead an `agreementId`:
```
const payload: OPFK8ComputeStop = {
      owner,
      providerSignature,
      providerAddress: config.keys.ethAddress,
      nonce,
      jobId
    }
```
On the operator service, we expect an optional `agreementId`, no `did` or `documentId`.
```
...
agreement_id = data.get("agreementId", None)
        owner = data.get("owner", None)
        job_id = data.get("jobId", None)
...
```
we are also sending an extra `providerAddress` (from Ocean Node, but not from python Provider) but we're not dealing with it on Operator Service (but that is secondary here)

paulo-ocean:
Fixes #506 .

Changes proposed in this PR:

- Add support for agreementId
- 
-


jamiehewitt15 (approved):
Looks good to me

PR #511, opened 2024-06-27
fix signature check for compute results, refactor fn

### LINKED ISSUE #509: Ocean Node computeResult signature/message mismatches the request from the SDK
On SDK we sign the following message:
```
let signatureMessage = await consumer.getAddress()
    signatureMessage += jobId
    signatureMessage += index.toString()
    signatureMessage += nonce
    const signature = await this.signProviderRequest(consumer, signatureMessage)
```
On Ocean Node we only check message like:
`String(ddoId + nonce)`

So, when we're gonna verify the address from the signature there is no match

paulo-ocean:
Fixes #509 

Changes proposed in this PR:

- Signature mismatches (check issue for details [#509 ](#509 )
- 
-


jamiehewitt15 (approved):
Looks good to me

alexcos20 (approved):

PR #512, opened 2024-06-27
Fix/ fileinfo service name

bogdanfazakas:
Fixes # .

Changes proposed in this PR:

- 
- 
-

### DIFF
--- a/src/components/httpRoutes/routeUtils.ts
+++ b/src/components/httpRoutes/routeUtils.ts
@@ -97,7 +97,7 @@ routesNames.set('broadcastCommand', {
   method: 'post'
 })
 // fileinfo
-routesNames.set('fileInfo', {
+routesNames.set('fileinfo', {
   path: '/api/fileinfo',
   method: 'post'
 })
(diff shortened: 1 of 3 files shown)


jamiehewitt15 (approved):
Looks good to me

src/components/httpRoutes/routeUtils.ts
```diff
@@ -97,7 +97,7 @@ routesNames.set('broadcastCommand', {
   method: 'post'
 })
 // fileinfo
-routesNames.set('fileInfo', {
+routesNames.set('fileinfo', {
   path: '/api/fileinfo',
```
mariacarmina, line 101: change path please

mariacarmina (commented):
after let's merge

src/components/httpRoutes/routeUtils.ts
bogdanfazakas: yep, good catch, need to update the http method itself since it was not backwards compatible

mariacarmina: it is updated in https://github.com/oceanprotocol/ocean-node/pull/485/commits/2242aca9569e48a87c4a5faad74eb2df7e5f027d

mariacarmina (approved):

paulo-ocean:
we need to have some kind of standard convention for this... all the other routes are Camel case, and this is the only one working differently :-)

jamiehewitt15:
@paulo-ocean Yeah, I agree

PR #516, opened 2024-06-28
fix operator service call to getResult (not computeResult) + signature message mismatches

### LINKED ISSUE #514: GET Compute Results calling wrong Operator Service path (computeResult vs getResult)
Ocean Node is calling Operator Service URL with wrong path:
example:
`GET https://stagev4.c2d.oceanprotocol.com/api/v1/operator/computeResult`  (gives a 404)
but should be:
`GET https://stagev4.c2d.oceanprotocol.com/api/v1/operator/getResult`

so correct PATH is `getResult` and NOT `computeResult`

Service definition on Operator Service:
`@services.route("/getResult", methods=["GET"])`

Also the signatures do not match, we're signing as:

```
public override async getComputeJobResult(
....
if (jobId) message = String(nonce + consumerAddress + jobId)
    else message = String(nonce + consumerAddress + jobId)
```
(nonce + consumerAddress + jobId)

Signature check on Operator Service (owner + jobId):
```
# verify provider's signature
        msg, status, provider_address = process_provider_signature_validation(
            data.get("providerSignature"), f"{owner}{job_id}", nonce
        )
```

paulo-ocean:
Fixes #514 

Changes proposed in this PR:

- The url is `getResult` not `computeResult` and the nonce is not part of the signature message
- index fix
-

### DIFF
--- a/src/components/c2d/compute_engines.ts
+++ b/src/components/c2d/compute_engines.ts
@@ -352,9 +352,11 @@ export class C2DEngineOPFK8 extends C2DEngine {
   ): Promise<Readable> {
     const nonce: number = new Date().getTime()
     const config = await getConfiguration()
-    let message: string
-    if (jobId) message = String(nonce + consumerAddress + jobId)
-    else message = String(nonce + consumerAddress + jobId)
+    // signature check on operator service is only owner + jobId
+    // nonce is not part of signature message
+    const message: string = jobId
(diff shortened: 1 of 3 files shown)


jamiehewitt15 (approved):
Looks good to me

mariacarmina (approved):
lgtm

PR #522, opened 2024-07-04
pass the headers back to client request

### LINKED ISSUE #521: getResult (get compute job result) is not passing the 'content-disposition' headers
The 'content-disposition' header is used to get the filename 
On the CLI there was already some changes to always have a default (it was mostly relying on this headers), but in any case
the headers should be passed back to the client to properly download the file(s)

paulo-ocean:
Fixes #521 

Changes proposed in this PR:

- pass the headers from the getResults (useful for download the proper attached file, with proper name)
-


jamiehewitt15 (approved):
Looks good to me

PR #537, opened 2024-07-08
Add default output stage.

### LINKED ISSUE #536: Create default output stage
Create default output stage

mariacarmina:
Fixes #536 .

Changes proposed in this PR:

- Add default output stage.


src/components/c2d/compute_engines.ts
```diff
@@ -212,6 +214,17 @@ export class C2DEngineOPFK8 extends C2DEngine {
         })
       index++
     }
+    // output stage
+    let getOuput = {}
```
paulo-ocean, line 218: Not a big deal, but this could be simpler,  (there are also typos btw :-) )
for example:
```
// output stage
    let getOuput = {}
    if (output) {
      getOuput = output
    } else if (config.hasHttp) {
      getOuput = {
        metadataUri: `http://${await publicIp()}:${config.httpPort}`
      }
    } 
```

src/components/c2d/compute_engines.ts
```diff
@@ -212,6 +214,15 @@ export class C2DEngineOPFK8 extends C2DEngine {
...
+    if (output) {
+      getOuput = output
+    } else if (config.hasHttp) {
+      getOuput = {
+        metadataUri: `http://${await publicIp()}:${config.httpPort}`
```
jamiehewitt15, line 223: It would be a good idea to check that we have a valid response from `await publicIp()`. There could be some edge cases where it fails or returns undefined.

mariacarmina: Sure good catch.

src/components/c2d/compute_engines.ts
```diff
@@ -22,6 +22,7 @@ import { getConfiguration } from '../../utils/config.js'
 import { ZeroAddress } from 'ethers'
 import { getProviderFeeToken } from '../../components/core/utils/feesHandler.js'
 import { URLUtils } from '../../utils/url.js'
+import { publicIp } from 'public-ip'
```
paulo-ocean, line 25: we have some IP utils functions for this under `src/utils/ip.ts`, we can use those. I think there's no need to install a package for such a simple function. Its the same kind of logic so it doesn't add.
We can probably improve to always have a default (even if it is just 'localhost') if something fails
tkx

alexcos20 (changes requested):
This is not how the overall architecture works!

metadataUri is used by pod-configuration to know where to fetch DDOs from (https://github.com/oceanprotocol/pod-configuration/blob/4ba2172b7ca2e1804f74b13d9a6126983b67a363/src/index.js#L95)

Making that public IP covers only one small edge case

Example
 - you have a network with private IPS.  192.168.0.2 is used by node , and your kubernetes cluster is using class 10.0.0.1/24 for pods.   And you have no port forwarding.      Setting metadataUri to some external IP will not make pod-configuration working by magic..
 
 Remove this PR please and replace it with:
  - add an env variable, called C2D_NODE_URI .   And use that. It's up to the network administrator to set this properly

mariacarmina:
> This is not how the overall architecture works!
> 
> metadataUri is used by pod-configuration to know where to fetch DDOs from (https://github.com/oceanprotocol/pod-configuration/blob/4ba2172b7ca2e1804f74b13d9a6126983b67a363/src/index.js#L95)
> 
> Making that public IP covers only one small edge case
> 
> Example
> 
> * you have a network with private IPS.  192.168.0.2 is used by node , and your kubernetes cluster is using class 10.0.0.1/24 for pods.   And you have no port forwarding.      Setting metadataUri to some external IP will not make pod-configuration working by magic..
> 
> Remove this PR please and replace it with:
> 
> * add an env variable, called C2D_NODE_URI .   And use that. It's up to the network administrator to set this properly

Ok sure then I'll let up to the user to set correctly, not trying to fallback to anything.

mariacarmina:
closing this and will refactor the logic in another PR

PR #539, opened 2024-07-08
Add algorithm metadata if it does not exist.

### LINKED ISSUE #538: Add algorithm metadata for compute jobs
Add algorithm metadata for compute jobs

mariacarmina:
Fixes #538 .

Changes proposed in this PR:

- Add algorithm metadata if it does not exist.


jamiehewitt15 (approved):
Looks good to me

src/@types/DDO/Metadata.ts
```diff
@@ -18,6 +18,11 @@ export interface MetadataAlgorithm {
    * @type {string}
    */
   rawcode?: string
+  /**
+   * Format of the algorithm
+   * @type {string}
+   */
+  format: string
```
paulo-ocean, line 25: Where does this "format" field comes from? What's its usage/values? Can't find any references anywhere 
tkx

src/components/core/compute/startCompute.ts
```diff
@@ -68,6 +68,7 @@ export class ComputeStartHandler extends Handler {
       const node = this.getOceanNode()
       const assets: ComputeAsset[] = [task.dataset]
       if (task.additionalDatasets) assets.push(...task.additionalDatasets)
+      const algorithm = structuredClone(task.algorithm)
```
paulo-ocean, line 71: there is no need for a deep copy here. In fact we might want exactly the opposite, as we want to add missing fields

src/components/core/compute/startCompute.ts
```diff
@@ -233,6 +234,19 @@ export class ComputeStartHandler extends Handler {
               validUntil: validFee.validUntil
             }
           }
+          if (ddo.metadata.type === 'algorithm' && !('meta' in algorithm)) {
+            algorithm.meta = {
```
paulo-ocean, line 238: This piece of code is not doing whats supposed to do..  we're adding some extra fields to an "`algorithm`" variable that we're not using anywhere. We want to update the` task.algorithm` cause that's what we are passing to `engine.startComputeJob()`
or we could use the constant above, but without the deep cloning
Also it looks like we could do some Object destructuring (for the `meta` and the `container` fields for instance... but checking first if the fields are present... otherwise this can break as is)
tkx

src/@types/DDO/Metadata.ts
paulo-ocean: Any hints? This field is not mentioned anywhere, neither its part of any ddo schema
thanks

mariacarmina: Hello, I found some occurrences in provider tests: https://github.com/oceanprotocol/provider/blob/4cd244bc18565c0d7bf631f97b584cbd199f0d03/tests/test_compute.py#L67. Format should indicate algorithm format if it's an docker image, but I checked the docs and also format field is not mentioned, in case this field is provided I can make it optional here.

paulo-ocean: ok, lets make sure we do not perpetuate bugs... most of current production code is buggy and it does not seem to be used. This field is another example... I think we should build right this time :-)

paulo-ocean (approved):
lgtm

PR #541, opened 2024-07-08
Check payment for compute.

mariacarmina:
Fixes # .

Changes proposed in this PR:

- 
- 
-

### DIFF
--- a/src/components/core/utils/feesHandler.ts
+++ b/src/components/core/utils/feesHandler.ts
@@ -76,7 +76,7 @@ export async function createProviderFee(
     providerFeeToken = computeEnv.feeToken
   } else {
     // it's download, take it from config
-    providerFeeToken = await getProviderFeeToken(asset.chainId)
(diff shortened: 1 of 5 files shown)


src/components/core/utils/feesHandler.ts
```diff
@@ -206,14 +206,12 @@ export async function verifyProviderFees(
...
   let isComputeValid = true
-  if (computeEnv) {
-    if (computeEnv) {
-      if (providerData.environment !== computeEnv && !isValidPayment) {
-        isComputeValid = false
-      }
-      if (validUntil > 0 && providerData.timestamp < validUntil) {
-        isComputeValid = false
-      }
+  if (computeEnv !== null) {
```
jamiehewitt15, line 209: Why not check `if (computeEnv)`? Now if `computeEnv` is undefined (rather than null) this equates to true

mihaisc:
@mariacarmina can you fix the tests please and the rest can we get some reviews

PR #542, opened 2024-07-08
Check if ddo state is active before executing node's commands.

### LINKED ISSUE #540: Check if ddo state is active before executing node's commands.
Check if ddo state is active before executing node's commands (download, validate ddo, initialize, initializeCompute, startCompute, fileinfo).

mariacarmina:
Fixes #540 .

Changes proposed in this PR:

- Check ddo state before executing core actions


jamiehewitt15 (changes requested):
This PR doesn't follow the state specification. You are treating every non-zero state the same when they are actually meant to be different. For example: ordering the asset is allowed for state 5. 

You need to refer to the state specification in the docs: https://docs.oceanprotocol.com/developers/ddo-specification#state

src/components/core/compute/initialize.ts
```diff
@@ -81,6 +82,16 @@ export class ComputeInitializeHandler extends Handler {
               }
             }
           }
+          if (ddo && ddo.nft.state && isOrderingAllowedForAsset(ddo)) {
```
paulo-ocean, line 85: Hi Maria
It seems you have the logic inverted on all the IF blocks :-)
we have:
```
if (ddo && ddo.nft.state && isOrderingAllowedForAsset(ddo)) {
...
// return some error
} 
```
We actually want to return an error IF  =>  ` ! isOrderingAllowedForAsset(ddo) ` (if this fn returns false)

One other thing, is that we're also not considering that `ddo.nft` might be `undefined` , so every time we have like  `ddo.nft.state`  it can potentially break..

IMO, it would be a lot more easier if all the IFs were simply re-written  like:
```
if (! isOrderingAllowedForAsset(ddo)) {
...
// return some error
}
```

Then we can check all the necessary conditions inside` isOrdedingAllowedForAsset() `function, without repeating same lines everywhere... like for example:
```
export function isOrderingAllowedForAsset(asset: DDO): boolean {
  if (!asset) {
    return false
  } else if (
    asset.nft &&
    !(asset.nft.state in [MetadataStates.ACTIVE, MetadataStates.UNLISTED])
  ) {
    return false
  }

  return true
}
```
thanks

paulo-ocean (changes requested):

src/components/httpRoutes/aquarius.ts
```diff
@@ -146,6 +147,10 @@ aquariusRoutes.post(`${AQUARIUS_API_BASE_PATH}/assets/ddo/validate`, async (req,
     }
 
     const node = req.oceanNode
+    if (ddo && ddo.nft.state && isOrderingAllowedForAsset(ddo)) {
+      HTTP_LOGGER.log(LOG_LEVELS_STR.LEVEL_ERROR, `Error: DDO NOT ALLOWED TO ORDER`)
+      res.status(404).send('Error: DDO NOT ALLOWED TO ORDER')
+    }
```
jamiehewitt15, line 153: This still doesn't seem right. For assets with state `1` and `4` they should be discoverable, which requires them to have a valid DDO. So why is the `/validate` endpoint blocked for these assets?

src/components/core/compute/initialize.ts
mariacarmina: sure, I'll modify accordingly.

src/components/httpRoutes/aquarius.ts
mariacarmina: Yes, that's right, for validate shouldn't be applicable.

src/components/core/compute/initialize.ts
mariacarmina: I'll also add to the isOrderingAllowedForAsset, the reason for possible errors in order to be logged.

```javascript

export function isOrderingAllowedForAsset(asset: DDO): OrdableAssetResponse {
  if (!asset) {
    return { 
       isOrdable: false,
       reason: `Asset provided is either null, either undefined ${asset}`
    }
  } else if (
    asset.nft &&
    !(asset.nft.state in [MetadataStates.ACTIVE, MetadataStates.UNLISTED])
  ) {
    return { 
       isOrdable: false,
       reason: "Nft not present in the asset or the state is different than ACTIVE or UNLISTED."
    }
  }

  return {
    isOrdable: true,
    reason: ''
  }
}
```

paulo-ocean (approved):
lgtm

jamiehewitt15 (approved):
Looks good to me

PR #546, opened 2024-07-09
use node v18.20.4

alexcos20:
Fixes # .

Changes proposed in this PR:

- use nodejs v18.20.4
- 
-


paulo-ocean (approved):
lgtm

bogdanfazakas (approved):
lgtm, worth trying 2+ hours on my end same memory usage as it was when it was turned on

PR #548, opened 2024-07-10
use static rpc provider

alexcos20:
Fixes # .

Changes proposed in this PR:

- use static RPC provider to decrease rpc calls

See:
 - https://github.com/ethers-io/ethers.js/discussions/3994
 - https://docs.ethers.org/v6/api/providers/jsonrpc/#JsonRpcApiProviderOptions


jamiehewitt15:
Can you fix the failing unit test?

paulo-ocean:
> Can you fix the failing unit test?

The problem is not the test. IMO, we need to revert this change (or simply forget about this PR)
This proposed change breaks network/availability functionality, we cannot anymore detect a proper RPC connection.. This change will accept any RPC url, even a completely wrong one...
Example:
If we set the RPC on the test (the 1st, not the fallbacks) to` http://fake-rcp.com:8545`
It will wrongly pass on the `detectNetwork()` function
Debug example:
```
RPC URL IS:  http://fake-rcp.com:8545
2024-07-24T09:14:05.037Z debug: DATABASE:       Request /collections/nonce: Request to Node localhost was made. Response Code was 200.
2024-07-24T09:14:05.039Z debug: DATABASE:       Request /collections/indexer
2024-07-24T09:14:05.039Z debug: DATABASE:       Request /collections/indexer: Attempting GET request Try #1 to Node localhost
    ✔ should get known rpcs
status: { ready: true }
    1) should get network not ready (wrong RPC setting)
```
Because we will have a valid `Network`, regardless of the address/ip/url being valid or not.. 

While with the previous/current  version we rightly detect the lower level network issue:
example:
```

error: CORE:   Unable to detect provider network: getaddrinfo ENOTFOUND fake-rcp.com
status: { ready: false, error: 'getaddrinfo ENOTFOUND fake-rcp.com' }
    ✔ should get network not ready (wrong RPC setting) (162ms)

```

paulo-ocean:
By the way, i juts figured out another way to do proper network detection, even with a static provider..
So, we can ignore my previous comments... Life found a way :-)

paulo-ocean (approved):
lgtm, fixed issue caused by change

jamiehewitt15 (approved):
Nice, looks good to me. Thanks for getting it working @paulo-ocean

PR #551, opened 2024-07-11
Fix/ Restart worker thread after unwanted exit

### LINKED ISSUE #544: Indexer worker exists, there is no retry
At some point in time, the indexer worker exited.
(I was running with one network only).

image

We should restart that worker if possible (most likely RPC connection failure)

bogdanfazakas:
Fixes #544  .

Changes proposed in this PR:

- 
- 
-


src/components/Indexer/index.ts
```diff
@@ -101,6 +101,114 @@ export class OceanIndexer {
...
+        }
+        INDEXER_LOGGER.logMessage(message, true)
+        // Restart the worker if it exits with a specific code (e.g., code 1)
+        if (code !== 0) {
+          INDEXER_LOGGER.log(
+            LOG_LEVELS_STR.LEVEL_INFO,
+            `Attempting to restart worker for network ${rpcDetails.network} with chainID ${chainID} ...`,
+            true
+          )
+          this.startThread(chainID)
```
paulo-ocean, line 208: hi @bogdanfazakas 
Have some doubts here.. i'm not sure if moving all this code from  `startThreads() ` into `startThread`() is the best approach. Also this recursive call `this.startThread()` feels a bit strange..
If we did this before:
`this.workers[chainID] = null`
aren't we also setting the `worker`  reference to `null` ?
we do this at the beginning:
`let worker = this.workers[chainID]` so its the same reference

And when we do these lines, the recursive call:
```
this.startThread(chainID)
...
this.workers[chainID] = worker
```
isn't the new worker null?
thanks

bogdanfazakas: Hi @paulo-ocean 
Let me try to reply to all the issues raised:
1. first with the idea of moving the event handles that we do on the worker thread to the `startThread()` method, were due to the following scenario, the thread for a certain network exited so it is dead, you want to restart the crawling process for that chain so you go and call startThread() again, but the events handlers on the new worker created are not attach to the worker object, cause we do that only in the main startThreads() method not in here (https://github.com/oceanprotocol/ocean-node/blob/develop/src/components/Indexer/index.ts#L101) when creating a new worker 
2. setting that worker reference to null is kind of neccesary since i belive, since that node exited due to an error and we want to create a new worker thread for the same chain so this reference needs to be null in the `startThread()` method
3. regarding the recursive call that can lead to filling the call stack, i agree maybe worth trying a certain amount of times (3 tries) for a specific chain, and just remove the refence and log a message

paulo-ocean: Hi @bogdanfazakas 
I think you're missing the point here :-) It just does not work AS IS
Just do this simple change and you'll see:

// create a timeout to terminate the worker thread  (this will trigger exit code 1)
```
...
if (!worker) {
      worker = new Worker('./dist/components/Indexer/crawlerThread.js', {
        workerData
      })

setTimeout(() => {
        this.killThread(worker)
      }, 20000)

  worker.on('message', (event: any) => {
...
```
create the following function to do that:

```
public killThread(worker: Worker) {
    console.log(`Killing worker thread...`)
    worker.terminate()
  }

Than checkout what happens

```

bogdanfazakas: good catch paolo thanks, the issues 2, 3 should be addressed and fixed now

paulo-ocean:
still not very good news @bogdanfazakas ,  i still get the code dump.. its true that it worked a couple of times, but still happens, i think we're basically missing something, needs a bit more testing IMO

bogdanfazakas:
> if (this.supportedChains.length > 0) {
>       setTimeout(() => this.killThread(Number(this.supportedChains[0])), 20000)
>     }
>   }

Hi @paulo-ocean i'm not able to reproduce the core dump issue any more with latest changes, also tried to test in a similar approach to yours multi times, never seen that core dump again always restarting until the 3'rd attempt them it's locked
attached the changes that are not merged for test regards and the usual  behaviour .

Screenshot 2024-08-05 at 12 21 44
Screenshot 2024-08-05 at 12 21 34
Screenshot 2024-08-05 at 12 19 17

paulo-ocean:
could you try with sepolia network?
Its weird, cause i get the crash 10/10 times

bogdanfazakas:
> could you try with sepolia network? Is the thread really indexing anything on local network?
> 
> Its weird, cause i get the crash 10/10 times

Yep sure on it, would be nice to catch it if it fails

paulo-ocean:
> > could you try with sepolia network? Is the thread really indexing anything on local network?
> > Its weird, cause i get the crash 10/10 times
> 
> Yep sure on it, would be nice to catch it if it fails

When the thread gets killed the `ThreadData` on crawlerThread dies with it,
the line here brings the problems:
`const { rpcDetails } = workerData as ThreadData`

bogdanfazakas:
> > > could you try with sepolia network? Is the thread really indexing anything on local network?
> > > Its weird, cause i get the crash 10/10 times
> > 
> > 
> > Yep sure on it, would be nice to catch it if it fails
> 
> When the thread gets killed the `ThreadData` on crawlerThread dies with it, (we delete all this) the line here brings the problems: `const { rpcDetails } = workerData as ThreadData`

So far so good tried 10 times till now means 30 successful restarts out 30 no issue. will do more testing to see if i can break it.
Regarding the question about indexing stuff yep, it continues to index from where it stopped see screen shot.
Regarding the thread data it should not be a problem because we in startThread we allways pass workerData when creating a new thread for that certain network.
Screenshot 2024-08-05 at 12 50 44

paulo-ocean:
> Regarding the thread data it should not be a problem because we in startThread we always pass workerData when creating a new thread for that certain network.

That's not the best assumption, IMO :-)...
If we're deleting this data, and we're trying to access it (we have that infinite loop that we abruptly kill), it will problematic, unless we're very lucky. I can approve the PR, but again it NEVER works on my side

bogdanfazakas:
> > Regarding the thread data it should not be a problem because we in startThread we always pass workerData when creating a new thread for that certain network.
> 
> That's not the best assumption, IMO :-)... If we're deleting this data, and we're trying to access it (we have that infinite loop that we abruptly kill), it will eventually be problematic, unless we're very lucky. I can approve the PR, but again, for the record, it NEVER works on my side thanks

yep as mentioned above i'm more trying to reproduce that failure again on my end rather than getting the PR approved 😄

paulo-ocean (approved):
seems we're getting different results.
On my end it always fails, but anyway, guess we can come back to this if others reproduce it

bogdanfazakas:
> seems we're getting different results. On my end it always fails, but anyway, guess we can come back to this if others reproduce it

I'll try now to test on the remote node, different OS and cpu than what i've been testing localy

bogdanfazakas:
> setTimeout(() => this.killThread(Number(this.supportedChains[0])), 20000)

@paulo-ocean indeed it reproduced on the remote vm with linux and not arm cpu..
Screenshot 2024-08-06 at 12 41 28

jamiehewitt15:
Yeah, it also doesn't work for me when running it locally (on Ubuntu) getting the same error: 

```
2024-08-06T11:18:16.080Z info: P2P:	Trying to punch a hole using UPNP
Segmentation fault (core dumped)
```

btw, if you include the `killThread` method, we could test this in the CI by calling the `killThread` and testing if it restarts.

paulo-ocean:
yep, happens every single time
i quickly tried a few approaches but couldn't reach to a fix..
It seems related with the data on the loop and/or the messages channel / listeners..
I tried to remove all the listeners of the "old worker" to avoid receiving stuff afterwards  - No effect
At one point i was getting some uncaught exceptions (i though i was close).. on the data we pass as argument to the worker thread ( rpcdetails -> threadData) ... like it was getting destroyed and still being accessed on the while() ... something like "cannot desconstruct something that is undefined), like here
`const { rpcDetails } = workerData as ThreadData`
so i tried to pass that data on the "start-crawling" message, instead of the constructor, and get rid of that, but i still had issues..
i also suspected about the `delete this.workers[chainID]` so i avoided that, and just replaced the object (added an optional  boolean on startThread() to force create a new Worker) - Still no effect
I tried to post a message to the thread on the 'exit', to do some cleanup, but didn't received the msg either
Maybe we need to run this with a profiler/debugger ...

bogdanfazakas:
> yep, happens every single time for me i quickly tried a few approaches but couldn't reach to a fix.. It seems related with the data on the loop and/or the messages channel / listeners.. I tried to remove all the listeners of the "old worker" to avoid receiving stuff afterwards - No effect At one point i was getting some uncaught exceptions (i though i was close).. on the data we pass as argument to the worker thread ( rpcdetails -> threadData) ... like it was getting destroyed and still being accessed on the while() ... something like "cannot desconstruct something that is undefined), like here `const { rpcDetails } = workerData as ThreadData` so i tried to pass that data on the "start-crawling" message, instead of the constructor, and get rid of that, but i still had issues.. i also suspected about the `delete this.workers[chainID]` so i avoided that, and just replaced the object (added an optional boolean on startThread() to force create a new Worker) - Still no effect I tried to post a message to the thread on the 'exit', to do some cleanup, but didn't received the msg either Maybe we need to run this with a profiler/debugger ...

yep also trying bunch of stuff on my end some of the similar with what you've tried, still having issues on the linux vm.

paulo-ocean:
> Yeah, it also doesn't work for me when running it locally (on Ubuntu) getting the same error:
> 
> ```
> 2024-08-06T11:18:16.080Z info: P2P:	Trying to punch a hole using UPNP
> Segmentation fault (core dumped)
> ```
> 
> btw, if you include the `killThread` method, we could test this in the CI by calling the `killThread` and testing if it restarts.

i think we could have a unit/integration test instead... we do have methods to stop / start threads already, but "normal" stop/start .. this is more to caught unexpect stuff... don't think we want a public method for terminate

paulo-ocean:
its a segmentation fault, so we basically are reading or writing to some place that we cannot anymore or have no access... i can't spot any issues on the index.js so i would say its on the crawlerThread.js ... but its only if we "restart" it which is strange

mariacarmina:
> > Yeah, it also doesn't work for me when running it locally (on Ubuntu) getting the same error:
> > ```
> > 2024-08-06T11:18:16.080Z info: P2P:	Trying to punch a hole using UPNP
> > Segmentation fault (core dumped)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > btw, if you include the `killThread` method, we could test this in the CI by calling the `killThread` and testing if it restarts.
> 
> i think we could have a unit/integration test instead... we do have methods to stop / start threads already, but "normal" stop/start .. this is more to caught unexpect stuff... don't think we want a public method for terminate

@bogdanfazakas  can we have this type of test to monitor the behaviour?

mihaisc:
@bogdanfazakas what is the status on this?

bogdanfazakas:
> behaviour

Need to jump back on it & try to find a solution for the Segmentation fault (core dumped) issue on linux OS

bogdanfazakas:
> @bogdanfazakas what is the status on this?

Need to jump back on it & try to find a solution for the Segmentation fault (core dumped) issue on linux OS

paulo-ocean:
> > @bogdanfazakas what is the status on this?
> 
> Need to jump back on it & try to find a solution for the Segmentation fault (core dumped) issue on linux OS

Maybe this kind of approach can help:
[https://stackoverflow.com/questions/58365748/how-to-prevent-a-node-js-12-worker-thread-from-terminating-immediately](https://stackoverflow.com/questions/58365748/how-to-prevent-a-node-js-12-worker-thread-from-terminating-immediately)
We probably need to catch other `Signals` on the main thread (instead of just trying a restart on 'exit' message) ...
Once we get those signals we call `stopThread`. This will do the necessary cleanup (it already stops processing)
Once we get the 'exit' message on the main thread, it should be already safe to restart the threads by calling `startThreads` again ... that will probably be enough

bogdanfazakas:
> > > @bogdanfazakas what is the status on this?
> > 
> > 
> > Need to jump back on it & try to find a solution for the Segmentation fault (core dumped) issue on linux OS
> 
> Maybe this kind of approach can help: https://stackoverflow.com/questions/58365748/how-to-prevent-a-node-js-12-worker-thread-from-terminating-immediately We probably need to catch other `Signals` on the main thread (instead of just trying a restart on 'exit' message) ... We have to make sure the while loop is not executing anymore... Once we get those signals we call `stopThread`. This will do the necessary cleanup (it already stops processing) Once we get the 'exit' message on the main thread, it should be already safe to restart the threads by calling `startThreads` again ... that will probably be enough (in theory :- ) )

Yes will try some more things on it once we merge the elastic stuff and the changes you've done there and see what else we should do from that. Ty

paulo-ocean:
IMO, i think you're still overdoing it and not addressing the real issue :-) 
We already have everything in place to start and stop a thread. There's no need to introduce a different restart logic... 

The only thing we need to make sure (IF possible) is that when the thread exits we immediately post a message to it in order to stop any code execution going on and stop running in a cleaner way (again, IF possible).. Otherwise, nothing will work. If we can't do this... our best chance is to make sure we cover everything with error handling inside the worker thread, to avoid that situation.. 

otherwise we will always have a segmentation fault with this separate worker thread design we have, if it terminates abruptly and any piece of code is still called there (for instance on the while() loop, before we can "clean" it and prevent more logic).. it will always create a segmentation fault

PR #553, opened 2024-07-15
Changes on logging transports (.env var based locations) 

### LINKED ISSUE #523: Logging:  Allow node owner to decide where to log
We should allow node-owner to decide where to see/keep the logs.

Proposal:  Refactor logging module such:

- have env LOG_LEVEL decide what logs are activated (no matter if NODE_ENV is production or development).  If LOG_LEVEL is not defined, then if NODE_ENV='production' -> LOG_LEVEL=info, else LOG_LEVEL=DEBUG
 - if env LOG_CONSOLE is True, then all logs should be displayed on the console
 - if env LOG_FILES is True, then all logs should be save in log fines
 - if env LOG_DB is True and we have noSQL db, then all logs should be saved in noSQL db

If none of LOG_CONSOLE, LOG_FILES, LOG_DB are set the we default to console
 

Default for casual user will be set in the yaml file (will not be set in code) would be LOG_DB: true and LOG_LEVEL: ERROR

paulo-ocean:
Fixes #523 .

Changes proposed in this PR:

- use .env vars to decide where to log (file, console, db)
- fix tests and core log logic
-


jamiehewitt15 (approved):
Looks good to me. We could probably also benefit from having a section in the docs about the logs, I'll handle that in the documentation PR

paulo-ocean:
> Looks good to me. We could probably also benefit from having a section in the docs about the logs, I'll handle that in the documentation PR

thanks, btw, just did a small update on the `env.md`

bogdanfazakas (approved):
lgtm

PR #554, opened 2024-07-15
add naive detection of public ip address

alexcos20:
Changes proposed in this PR:

- detects addresses used by node
- for each ip family, try to detect public ip address
- if found, add it to transports as observedAddress, other nodes will have to confirm it


jamiehewitt15 (approved):
Looks good to me. It's definitely better than nothing

paulo-ocean (approved):
lgtm

PR #559, opened 2024-07-19
Fix downloading full content of the file.

### LINKED ISSUE #543: Download stream does not contain all algorithm file
Download stream does not contain all algorithm file. When starting a compute job in /data/transformations/algorithm, the algorithm is incomplete, maybe from writting incompletely the chunks from stream. The file has aprox 2.5 kb https://github.com/oceanprotocol/c2d-examples/blob/main/branin_and_gpr/gpr.py
Output:
```
/pvc/transformations # cat algorithm
#
# Copyright 2022 Ocean Protocol Foundation
# SPDX-License-Identifier: Apache-2.0
#
import json
import os
import pickle
import sys

import arff
import matplotlib
import numpy
from matplotlib import pyplot
from sklearn import gaussian_process

matplotlib.use("agg")


def branin_mesh(X0, X1):
    # b,c,t = 5.1/(4.*(pi)**2), 5./pi, 1./(8.*pi)
    b, c, t = 0.12918450914398066, 1.5915494309189535, 0.039788735772973836
    u = X1 - b * X0**2 + c * X0 - 6
    r = 10.0 * (1.0 - t) * numpy.cos(X0) + 10
    Z = u**2 + r

    return Z


def create_mesh(npoints):
    X0_vec = numpy.linspace(-5.0, 10.0, npoints)
    X1_vec = numpy.linspace(0.0, 15.0, npoints)
    X0, X1 = numpy.meshgrid(X0_vec, X1_vec)
    Z = branin_mesh(X0, X1)

    return X0, X1, Z


def get_input(local=False):
    if local:
        print("Reading local file branin.arff.")

        return "branin.arff"

    dids = os.getenv("DIDS", None)

    if not dids:
        print("No DIDs found in environment. Aborting.")
        return

    dids = json.loads(dids)

    for did in dids:
        filename = f"data/inputs/{did}/0"  # 0 for metadata service
        pri
```

mariacarmina:
Fixes #543  .

Changes proposed in this PR:

- pass the full content length of the file


src/components/core/handler/downloadHandler.ts
```diff
@@ -129,13 +130,33 @@ export async function handleDownloadUrlCommand(
...
     } else {
       // Download request is not using encryption!
-      return {
-        stream: inputStream.stream,
-        status: {
-          httpStatus: inputStream.httpStatus,
-          headers
-        }
-      }
+      const chunks: any = []
```
paulo-ocean, line 133: hi 
i don't think this is the correct fix. we don't need to read the stream on the node, just to get the content length. 
It should be the final client to read the stream content.
what is missing is to fetch the metadata like this:
`const metadata = await storage.fetchSpecificFileMetadata(task.fileObject, true)`
before:
`const inputStream = await storage.getReadableStream()`
the metadata object has all the info we need, including content type, correct content lenght and file name (no need for the `unknownfile` either
the metadata will look similar to this:
```
{
  valid: true,
  contentLength: '319520',
  contentType: 'text/plain; charset=utf-8',
  checksum: '1f7c17bed455f484f4d5ebc581cde6bc059977ef1e143b52a703f18b89c86a22',
  name: 'shs_dataset_test.txt',
  type: 'url',
  encryptedBy: undefined,
  encryptMethod: undefined
}
```
we should use the `contentLenght` to set the corresponding header and the `name` to set the file name
thanks

mariacarmina: Ok, sure, I can try using our methods, my approach was more rudimentary to process in chunks the file content and update the content-length with the buffer size :).

paulo-ocean (approved):
lgtm

PR #566, opened 2024-07-29
Issue 565 optimize get status

### LINKED ISSUE #565: Status endpoint caching (called often from dashboard for instance)
Trying to optimize some stuff here, due to the high CPU usage.
Regarding the status endpoint...
The status endpoint is called at least 1 time per minute if running the dashboard, but most of the info there is static, and will not change until the node restarts, so it would be nice to cache some info and avoid some I/O and other intensive calls, like DB calls

paulo-ocean:
Fixes #565 #567 

Changes proposed in this PR:

- cache some data on getStatus()... Only fetch new data that actually changes between requests
- fix small issue on P2P` handleProtocolCommands()` if task/command is `undefined` (can crash node)
-

### DIFF
--- a/dashboard/src/components/IndexQueue.tsx
+++ b/dashboard/src/components/IndexQueue.tsx
@@ -9,7 +9,6 @@ import {
 } from '@mui/material'
 import styles from './Dashboard/index.module.css'
 import { useAdminContext } from '@/context/AdminProvider'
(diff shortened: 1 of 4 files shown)


jamiehewitt15 (approved):
Looks good to me

bogdanfazakas (approved):

PR #572, opened 2024-07-31
fix possible crash, add try catch when cannot process task and send response

paulo-ocean:
Fixes # .

Changes proposed in this PR:

- add try/catch when trying to process invalid/missing data and send response
- 
-


alexcos20 (approved):
lgtm

bogdanfazakas (approved):

PR #575, opened 2024-08-01
fix root endpoint, without RPCS

paulo-ocean:
Fixes #754

Changes proposed in this PR:

- fix root endpoint when no RPCS is set (avoid crash)
- we were trying to send the json after closing the response with a 400
-


jamiehewitt15 (approved):
Looks good to me

bogdanfazakas (approved):

PR #578, opened 2024-08-01
Clean up unused deps

mariacarmina:
Fixes # .

Changes proposed in this PR:

- 
- 
-


mariacarmina:
> Did you use `v20.16.0` when you ran `npm i` ?

Nope, it was before merging the PR with node version update.

jamiehewitt15:
Can you do a pull from develop then `npm i` using node v20.16.0

paulo-ocean:
btw, it does not build with node v18 either... why did we removed the types `chai` deps? that's used on tests

PR #579, opened 2024-08-01
Check for timeout

### LINKED ISSUE #581: Ensure that we send a 408 if the express request has timed out
Ensure that we send a 408 if the express request has timed out

jamiehewitt15:
Fixes #581 

Changes proposed in this PR:

- Check for timeout in express request.
- Logging the timeout
- Sending `408`

### DIFF
--- a/env.md
+++ b/env.md
@@ -28,6 +28,7 @@ Environmental variables are also tracked in `ENVIRONMENT_VARIABLES` within `src/
 - `LOG_CONSOLE`: Write logs to the console. Default is `false`, but becomes `true` if neither `LOG_FILES` or `LOG_DB` are set.
 - `LOG_FILES`: Write logs to files. Default is `false`
 - `LOG_DB`: Write logs to noSQL database. Default is `false`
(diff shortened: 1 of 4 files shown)


paulo-ocean:
i think we might need a bigger timeout... when we're dealing with database or remote P2P nodes... depending on the request, this might not be enough.. For instance, just for the P2P connection (from node A to B) we have default timeout of 10 secs... i would double it

paulo-ocean (commented):
what happens if we are streaming a response, somewhere on our handlers, and we hit this timeout?
15 seconds is way too low.. the default is 2 minutes... we're reducing way too much 
I doubt we can fulfill  any big request, specially if it deals with DB and remote peers in this interval
That said, what's the point / advantage of having a smaller request timeout?
If we have (possibly) timeout issues with bigger intervals, doesn't this change make it worst?
thanks

jamiehewitt15:
Yeah, good to have it with a high default (2min or more) and then the option to decrease it with an env

mihaisc:
@jamiehewitt15  status on this?

PR #582, opened 2024-08-01
Check we have a response before accessing it

### LINKED ISSUE #580: Always check we have received a response before accessing it
For various reasons, we might not receive a response from a request (e.g. connection has timed out). We should always check we have received a response before accessing it so we don't cause the node to crash

jamiehewitt15:
Fixes #580 

Changes proposed in this PR:

- Check we have a response before accessing it
- There are many reasons why we might not receive a response (e.g. the request could have timed out / the node we're requesting from could have crashed etc).


jamiehewitt15:
> How can the response not be defined? :-)

Yeah true, I'll close this. I can potentially open some more targetted.

PR #583, opened 2024-08-01
Logging all errors

jamiehewitt15:
Fixes # .

Changes proposed in this PR:

- We have a few try/catch statements where we don't log the error. This is not very helpful for debugging. 
- Adding the additional logs gives a clearer picture of why the node might have failed in these situations.


paulo-ocean (approved):
lgtm

bogdanfazakas (approved):
lgtm

PR #586, opened 2024-08-01
more try catch

paulo-ocean:
Fixes # .

Changes proposed in this PR:

- outer try/catch on directCommand, just in case
- improve some logs
-


jamiehewitt15 (approved):
looks good to me

PR #594, opened 2024-08-05
make sure we always start indexing on ganache, even if no block info available (from block 0)…

### LINKED ISSUE #493: Check indexer start block on ganache, with ADDRESS_FILE provided
Some notes from meeting:
- If not provided it should start at block 0
- deployment block could be null
- consider that ADDRESS_FILE is mandatory for local network
TODO:
double check all this

paulo-ocean:
…, from block zero

Fixes #493 

Changes proposed in this PR:

- even if we can't get block info, we still start from block 0 (local network only) 
- helper utils fn to check for undefined/null same time


bogdanfazakas (approved):
lgtm

jamiehewitt15 (approved):
Looks good to me

PR #599, opened 2024-08-11
Feature/ add Elasticsearch database alternative for typesense

### LINKED ISSUE #590: Database:  Move from typesense to ElasticSearch
- Add a new DB Class for ElasticSearch
 - Change all functions to use new ElasticSearchDB class

bogdanfazakas:
Fixes #590  .

Changes proposed in this PR:

- create abstract base database classes 
- created factory to instantiate one of the available databases
- add elastic implementation for typesense databases
- add datatoken address, nftAddress,  did in order index or schemas


src/components/database/DatabaseFactory.ts
```diff
@@ -0,0 +1,91 @@
...
+import { IDdoStateQuery } from '../../@types/DDO/IDdoStateQuery.js'
+import { TypesenseDdoStateQuery } from './TypesenseDdoStateQuery.js'
+import { ElasticSearchDdoStateQuery } from './ElasticSearchDdoStateQuery.js'
+import { TypesenseMetadataQuery } from './TypesenseMetadataQuery.js'
+import { IMetadataQuery } from '../../@types/DDO/IMetadataQuery.js'
+import { ElasticSearchMetadataQuery } from './ElasticSearchMetadataQuery.js'
+
+export class DatabaseFactory {
+  static createNonceDatabase(config: OceanNodeDBConfig): AbstractNonceDatabase {
+    if (process.env.DB_TYPE === 'elasticsearch') {
```
paulo-ocean, line 37: I think we could simplify these. Its pretty much the same code repeated multiple times.
For instance, all these functions are called here:
`(src/components/database/index.ts)`
```
this.ddo = await DatabaseFactory.createDdoDatabase(this.config)
      this.nonce = await DatabaseFactory.createNonceDatabase(this.config)
      this.indexer = await DatabaseFactory.createIndexerDatabase(this.config)
      this.logs = await DatabaseFactory.createLogDatabase(this.config)
      this.order = await DatabaseFactory.createOrderDatabase(this.config)
      this.ddoState = await DatabaseFactory.createDdoStateDatabase(this.config)
```
we could easily just have one method to create them all..

src/components/Indexer/processor.ts
```diff
@@ -90,13 +90,15 @@ class BaseEventProcessor {
       const saveDDO = await ddoDatabase.update({ ...ddo })
       await ddoState.update(
         this.networkId,
-        saveDDO.id,
+        saveDDO._id ?? saveDDO.id,
```
paulo-ocean, line 93: what changes here? nothing :-)

src/test/integration/download.test.ts
```diff
@@ -174,6 +176,10 @@ describe('Should run a complete node flow.', () => {
   })
   it('should publish compute datasets & algos', async () => {
     publishedDataset = await publishAsset(downloadAsset, publisherAccount)
+    function sleep(ms: number) {
+      return new Promise((resolve) => setTimeout(resolve, ms))
+    }
+    await sleep(5000)
```
paulo-ocean, line 182: why the sleep() ? its unpredictable , it could be enough time or not..
we could just use the `waitToIndex` call bellow , with a bigger test timeout and/or force waiting for the event

src/test/integration/download.test.ts
```diff
@@ -73,16 +73,18 @@ describe('Should run a complete node flow.', () => {
...
           JSON.stringify(mockSupportedNetworks),
           JSON.stringify([8996]),
           '0xc594c6e5def4bab63ac29eed19a134c130388f74f019bc74b8f4389df2837a58',
-          'http://localhost:8108/?apiKey=xyz',
+          'http://localhost:9200',
           JSON.stringify(['0xe2DD09d719Da89e5a3D0F2549c7E24566e947260']),
           JSON.stringify(['0xe2DD09d719Da89e5a3D0F2549c7E24566e947260']),
-          `${homedir}/.ocean/ocean-contracts/artifacts/address.json`
+          `${homedir}/.ocean/ocean-contracts/artifacts/address.json`,
+          'elasticsearch'
```
paulo-ocean, line 87: I think it would be easier to maintain if we create a constants like for ex:
```
DB_TYPES = {
ELASTIC_SEARCH: 'elasticsearch',
TYPESENSE: 'typesense'
}
```
And we replace the string occurrences... if we use just the plain strings every time we need to distinguish we can easily get misspelling / typos

paulo-ocean:
How do we use elastic search on the nodes?
If i set this `export DB_TYPE=elasticsearch`
The node does not even starts ?
tkx

src/components/database/DatabaseFactory.ts
bogdanfazakas: yes makes sense, will update that

src/test/integration/download.test.ts
bogdanfazakas: hmm fair enough, i'll have to check it maybe i've missed this once i reviewed the changes i've merged into this big PR 😃

src/test/integration/download.test.ts
bogdanfazakas: will handle it makes sense

paulo-ocean:
> > How do we use elastic search on the nodes? If i set this `export DB_TYPE=elasticsearch` The node does not even starts ? tkx
> 
> hmm strange, are you running barge with elastic search ? do you have the db env var pointing to proper elastic instance?

Nope, that's why we need to update documentation as well :-)

jamiehewitt15 (commented):
We need some documentation on the different databases and how to use them. Probably best avoid cluttering the main readme with it. It would be good to have a separate file in the `/docs` directory which explains the options and how to start them.

jamiehewitt15:
Looks like there are some conflicts, can you pull from main and resolve them?

paulo-ocean:
i checked the conflicts. I think you can cherry-pick this commit for instance:
[https://github.com/oceanprotocol/ocean-node/pull/706/commits/dcab22bea84010f078241dc98491106edcd8eb7b](https://github.com/oceanprotocol/ocean-node/pull/706/commits/dcab22bea84010f078241dc98491106edcd8eb7b)
We lost some functionality there, not sure why you deleted the utils function to check DB configuration..
Also, we cannot prooced to configure the DB transport if the DB url is not valid

paulo-ocean:
> i checked & fixed the conflicts on another branch. I think you can cherry-pick this commit for instance: [dcab22b](https://github.com/oceanprotocol/ocean-node/commit/dcab22bea84010f078241dc98491106edcd8eb7b) We lost some functionality there, for instance; we should not proceed to configure the DB transport if the DB url is not valid. That should be the first check Also, not sure why you deleted the utils function to check DB configuration.. what's the point of that? I'm also not sure about the try/catch on Database constructor (although i've kept it)... If there is an error being `Throw` doesn't that break the server startup? We should be able to use the node without DB

We should probably return `null` if we catch an error (instead on throwing a new Error)
and update the check on index:
```
const dbconn: Database = await new Database(config.dbConfig)
if (!dbconn || !hasValidDBConfiguration(config.dbConfig)) {
...
config.hasIndexer = false
```

paulo-ocean:
@bogdanfazakas can you update from main? thanks

bogdanfazakas:
> @bogdanfazakas can you update from main? thanks

Yep sure

paulo-ocean (approved):

jamiehewitt15 (approved):
Looks good to me

PR #653, opened 2024-08-26
add total dashboard

mihaisc:
image


src/components/P2P/index.ts
```diff
@@ -119,9 +119,53 @@ export class OceanP2P extends EventEmitter {
     this._libp2p.addEventListener('peer:disconnect', (evt: any) => {
       this.handlePeerDisconnect(evt)
     })
-    this._libp2p.addEventListener('peer:discovery', (evt: any) => {
-      this.handlePeerDiscovery(evt)
+    this._libp2p.addEventListener('peer:discovery', (details: any) => {
```
paulo-ocean, line 122: i think we should either use the functions `handlePeerDiscovery` and  `handlePeerDiscovery2` or just remove them (they are not used at the moment)
The code on the 'peer:discovery' listener seems the same that we have on function `handlePeerDiscovery` (apart the `await`  keyword)
Also the code on event listener 'kad-dht:query:final-peer' seems pretty much the same (only difference is the event name)
There are also a bunch of `console.log` that we might want to remove?
I think it needs some cleaning/refactor
thanks

PR #662, opened 2024-08-29
Dashboard UI update: connected nodes

### LINKED ISSUE #661: Dashboard: reduce the number of connected nodes shown by default.
Dashboard: reduce the number of connected nodes shown by default. 

The current situation where we show a list of 500+ node addresses is quite a bad UI. 

image

jamiehewitt15:
Fixes #661

Changes proposed in this PR:

- Showing only the first 10 connected node IDs
- Users can click "Show All" if they want to see all of them.

image


paulo-ocean (approved):
LGTM, but we need to improve this later, probably we need a table to avoid endless scrolling with many nodes :-) ..

jamiehewitt15:
> probably we need a table to avoid endless scrolling with many nodes

Yeah good point, a table would be better. We can add that later

bogdanfazakas (approved):

PR #667, opened 2024-08-30
removing IP address & adjusting UI

### LINKED ISSUE #629: Node http up IP
Where do we get this IP from? It's not ok

jamiehewitt15:
Fixes #629 

Changes proposed in this PR:

- removing IP address
- adjusting UI to remove the empty column on the right.

image


paulo-ocean (approved):
LGTM
btw, not sure if we need to shorten the node ids, its public data and we have a lot of space on the page anyway

jamiehewitt15:
> LGTM btw, not sure if we need to shorten the node ids, its public data and we have a lot of space on the page anyway

Yeah good point, I will remove that. Much better to see the whole string given that we have the space.

alexcos20 (approved):

PR #674, opened 2024-09-04
Renaming dashboard as Control Panel

jamiehewitt15:
Fixes # .

Changes proposed in this PR:

- The control panel (formerly known as dashboard) is now available at `/controlpanel`
- I have kept the `/dashboard` endpoint so that this is not a breaking change and it does not confuse anyone who is used to access it there.
- Updated mentions of the dashboard and changed the to control panel in the readmes
- To make this more lightweight I haven't changed every use of the word dashboard in our code.


paulo-ocean (approved):
LGTM

bogdanfazakas (approved):
lgtm

PR #675, opened 2024-09-04
Downgrade indexer errors

### LINKED ISSUE #672: Mark RPC errors as warnings
Users get concerned when they start the node with the default settings and then they immediately see errors. These are typically RPC-related errors which arise due to the public RPCs that are provided as default. I suggest that we downgrade RPC related errors to warnings.

jamiehewitt15:
Fixes #672 

Changes proposed in this PR:

- Issue with getting events is now a warning rather than an error
- "INDEXER_NETWORKS is not defined" is not an error, it is now shown as `INFO`

### DIFF
--- a/src/components/Indexer/crawlerThread.ts
+++ b/src/components/Indexer/crawlerThread.ts
@@ -150,7 +150,7 @@ export async function processNetworkData(
         } catch (error) {
           INDEXER_LOGGER.log(
             LOG_LEVELS_STR.LEVEL_WARN,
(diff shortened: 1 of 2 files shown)


mariacarmina (approved):
LGTM

alexcos20 (approved):

paulo-ocean (approved):
LGTM
btw, we can also use the shortcut functions (less verbose)
like:
```
INDEXER_LOGGER.warn()
INDEXER_LOGGER.error()
```
etc

PR #679, opened 2024-09-11
Issue 668 oasis decryption

### LINKED ISSUE #668: Add sapphire decryption
When downloading an asset with dt4 and on confidential network (sapphire or test network) decrypt FileInfo using oasis SDK. After this follow standard flow

paulo-ocean:
Fixes #668 

Changes proposed in this PR:

- handles template 4 (confidential EVM) download/decryption
-


src/components/core/compute/initialize.ts
```diff
@@ -104,14 +110,58 @@ export class ComputeInitializeHandler extends Handler {
...
-            canDecrypt = true
+            if (!confidentialEVM) {
+              await decrypt(
+                Uint8Array.from(Buffer.from(sanitizeServiceFiles(service.files), 'hex')),
+                EncryptMethod.ECIES
+              )
+              canDecrypt = true
+            } else {
+              // TODO use oasis sdk to decrypt
+              CORE_LOGGER.warn('TODO confidential EVM initialize compute')
```
jamiehewitt15, line 144: Let's not put a TODO in the logs. Whenever we have a warning in the logs it leads to a load of questions on Discord. Can you make sure we are also tracking the TODOs in an issue if you don't implement them on this PR.

src/components/core/compute/startCompute.ts
```diff
@@ -126,14 +132,57 @@ export class ComputeStartHandler extends Handler {
...
-            canDecrypt = true
+            if (!confidentialEVM) {
+              await decrypt(
+                Uint8Array.from(Buffer.from(sanitizeServiceFiles(service.files), 'hex')),
+                EncryptMethod.ECIES
+              )
+              canDecrypt = true
+            } else {
+              // TODO
+              CORE_LOGGER.warn('TODO confidential EVM start compute')
```
jamiehewitt15, line 165: Same again here, I wouldn't put a TODO in the logs as a warning.

jamiehewitt15:
Is there a reason you didn't add any tests? It would make it easier to check that it's working, even if they are on the live network and not included in the CI (like Maria did in her PR)

src/components/core/compute/initialize.ts
paulo-ocean: i can throw an "not implemented" exception instead of logging. This is something that is not even defined ATM

paulo-ocean:
> Is there a reason you didn't add any tests? It would make it easier to check that it's working, even if they are on the live network and not included in the CI (like Maria did in her PR)

we already have tests for the download, imo, we don't need a specific just to do an extra smart contract function call, and this is a live network so  it will not be maintainable. I guess the same reason we don't have a test for all the EVMs we support :-)

src/components/core/compute/initialize.ts
paulo-ocean: created an issue to address the signatures for confidential evm c2d

paulo-ocean:
> > Is there a reason you didn't add any tests? It would make it easier to check that it's working, even if they are on the live network and not included in the CI (like Maria did in her PR)
> 
> we already have tests for the download, imo, we don't need a specific just to do an extra smart contract function call, and this is a live network so it will not be maintainable. I guess the same reason we don't have a dedicated test for all the EVMs we support :-) Also. If they are not part of the CI, what are they good for? we can just use the CLI for ex:

added a couple of unit tests for aux functions

bogdanfazakas (approved):

jamiehewitt15 (approved):
Looks good to me

PR #694, opened 2024-09-23
Policy Server specs

alexcos20:
- [x] Write Policy Server specs - https://github.com/oceanprotocol/ocean-node/blob/feature/policyserver/docs/PolicyServer.md
- [x] Update ocean-node to check policyServer for newDDO when indexing
- [x] Update ocean-node to check policyServer for updateDDO when indexing
- [x] Update ocean-node to check policyServer for intialize handler
- [x]  Update ocean-node to check policyServer for download handler
- [x]  Update ocean-node to check policyServer for encrypt handler
- [x]  Update ocean-node to check policyServer for decrypt handler
- [x] Add ocean-node policyServer  passthrough endpoint


docs/PolicyServer.md
```diff
@@ -0,0 +1,109 @@
+Sometimes, actions performed by Ocean Node has to be double checked on a higher level of authorization. This might include Oath tokens, SSI verifible credentials, Enterprise LDAP, etc..
```
jamiehewitt15, line 1: We should have an h1 title at the top of the page

```suggestion
# Policy Server

Sometimes, actions performed by Ocean Node have to be double-checked on a higher level of authorization. This might include Oath tokens, SSI verifiable credentials, Enterprise LDAP, etc...
```

docs/PolicyServer.md
```diff
@@ -0,0 +1,109 @@
...
+
+For every command, Ocean Node will query PolicyServer (if such env is defined) and wait for it to perform all needed checks.
+
+For 200 OK responses, Ocean Node will continue and perform the action. For everything else, will deny. If there is a body in response, we will forward that body to the caller. (so users can see the PolicyServer error messages and act accordingly)
+
+Every Ocean Node command will also accept a data field, called "policyServer" which will be added to the query (so we can pass data from user to PolicyServer)
+
+# PolicyServer API definition
+
+All querys will be performed by sending a POST request to PolicyServer Endpoint, with a json payload that looks like:
```
jamiehewitt15, line 13:
```suggestion
For this, we will adopt a simple, but flexible arhitecture:

For every command, Ocean Node will query PolicyServer (if such env is defined) and wait for it to perform all needed checks.

For 200 OK responses, Ocean Node will continue to perform the action. For everything else, it will deny. If there is a body in response, we will forward that body to the caller. (so users can see the PolicyServer error messages and act accordingly)

Every Ocean Node command will also accept a data field, called "policyServer" which will be added to the query (so we can pass data from the user to PolicyServer)

## PolicyServer API definition

All queries will be performed by sending a POST request to PolicyServer Endpoint, with a json payload that looks like this:
```

docs/PolicyServer.md
```diff
@@ -0,0 +1,109 @@
...
+
+```json
+{
+    "action":"newDDO",
+    ......
+}
+```
+
+Every command will have it's own set of data, in addition to "action" field.
+I will describe them below
```
jamiehewitt15, line 23:
```suggestion
Every command will have its own set of data, in addition to the "action" field.
I will describe them below:
```

src/components/policyServer/index.ts
```diff
@@ -0,0 +1,103 @@
...
+export class PolicyServer {
+  serverUrl: string
+
+  public constructor() {
+    this.serverUrl = process.env.POLICY_SERVER_URL
+  }
+
+  private async askServer(command: any): Promise<PolicyServerResult> {
+    if (!this.serverUrl) return { success: true, message: '', httpStatus: 0 }
+    const response = await fetch(this.serverUrl, {
```
paulo-ocean, line 14: i would at least put some try/catch here, in case something goes wrong with the `fetch` or maybe we have an invalid  `process.env.POLICY_SERVER_URL`
Apart from that details looks good

alexcos20: good catch. done

paulo-ocean (approved):
lgtm

jamiehewitt15 (approved):
Looks good to me

PR #697, opened 2024-09-24
Add max ram memory in node start cmd

bogdanfazakas:
Fixes # .

Changes proposed in this PR:

- allow ocean node to use up to 28gb of ram if available instead of the default  8gb of memory
- 
-

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "clean": "if [ -d ./dist ]; then find ./dist -mindepth 1 -not -path './dist/dashboard*' -delete; fi",
     "clean:all": "rm -rf ./dist/ ./doc/ ./.nyc_output",
     "build-tests": "rm -rf ./dist/test && npm run build-tests:tsc",
-    "start": "node --trace-warnings --experimental-specifier-resolution=node dist/index.js",
+    "start": "node --max-old-space-size=28784 --trace-warnings --experimental-specifier-resolution=node dist/index.js",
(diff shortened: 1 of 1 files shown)


jamiehewitt15 (approved):

paulo-ocean (approved):
lgtm
but i wonder if we should change the specs/system requirements section (on the docs)
Currently we have:
```
- 1vcpu
- 2 GB ram
- 4 GB storage
```
According to the node docs we should put a lower number than the total available RAM
[https://nodejs.org/api/cli.html#--max-old-space-sizesize-in-megabytes](https://nodejs.org/api/cli.html#--max-old-space-sizesize-in-megabytes)
In this case, for a 2GB RAM we are allowing the entire 2GB for garbage collection

bogdanfazakas:
> lgtm but i wonder if we should change the specs/system requirements section (on the docs) Currently we have:
> 
> ```
> - 1vcpu
> - 2 GB ram
> - 4 GB storage
> ```
> 
> According to the node docs we should put a lower number than the total available RAM https://nodejs.org/api/cli.html#--max-old-space-sizesize-in-megabytes In this case, for a 2GB RAM we are allowing the entire 2GB for garbage collection, but i guess we always have more than this anyway

i think the specs are fine, they are related to the vm rather the node app config, so in this case the process will use around 1.5 gb of ram max , like in the eg bellow for a vm with 2gb of ram avaialble
Screenshot 2024-09-25 at 11 16 14

PR #699, opened 2024-09-26
fix default filtered p2p ip list (non routable ips that should not be announced)

### LINKED ISSUE #691: correct the default value for P2P_FILTER_ANNOUNCED_ADDRESSES
The current default value for `P2P_FILTER_ANNOUNCED_ADDRESSES` is `172.15.0.0/24` (which does not belong to a private range)

https://github.com/oceanprotocol/ocean-node/blob/181a3e177521b32f099d4bb44b803043d04df7e8/src/utils/config.ts#L532-L534

However, the default range for Docker IPAM is larger and depends on the driver being used (for example, for the bridge, it is `172.17.0.0/16`).

In other cases (for Docker), the range used is `10.0.0.0/8`

**Proposal:** Set the default value for this variable to include all non-routable IP addresses that are not accessible on the public internet, serve specific purposes within private networks, or are reserved for specific uses such as localhost, DHCP, multicast, testing, documentation, and future use.

```shell
127.0.0.0/8
10.0.0.0/8
172.16.0.0/12
192.168.0.0/16
100.64.0.0/10
169.254.0.0/16
192.0.0.0/24
192.0.2.0/24
198.51.100.0/24
203.0.113.0/24
224.0.0.0/4
240.0.0.0/4
```

paulo-ocean:
Fixes #691 .

Changes proposed in this PR:

- set proper default non public ips on P2P_FILTER_ANNOUNCED_ADDRESSES
- 
-

### DIFF
--- a/src/test/unit/oceanP2P.test.ts
+++ b/src/test/unit/oceanP2P.test.ts
@@ -56,7 +56,7 @@ describe('OceanP2P Test', () => {
     config2.p2pConfig.bootstrapNodes = []
     // enable private IP
     config2.p2pConfig.announcePrivateIp = true
-    config1.p2pConfig.filterAnnouncedAddresses = ['172.15.0.0/24'] // allow nodes to see each other locally for tests
+    config2.p2pConfig.filterAnnouncedAddresses = ['172.15.0.0/24'] // allow nodes to see each other locally for tests
     node2 = new OceanP2P(config2, null)
     await node2.start()
     assert(node2, 'Failed to create P2P Node instance')
--- a/src/test/unit/oceanP2P.test.ts
+++ b/src/test/unit/oceanP2P.test.ts
@@ -42,6 +42,7 @@ describe('OceanP2P Test', () => {
     config1.p2pConfig.bootstrapNodes = []
     // enable private IP
     config1.p2pConfig.announcePrivateIp = true
+    config1.p2pConfig.filterAnnouncedAddresses = ['172.15.0.0/24']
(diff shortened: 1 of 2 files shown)


paulo-ocean:
need to fix tests, break unit tests with 2 local nodes

src/test/unit/oceanP2P.test.ts
```diff
@@ -55,6 +56,7 @@ describe('OceanP2P Test', () => {
     config2.p2pConfig.bootstrapNodes = []
     // enable private IP
     config2.p2pConfig.announcePrivateIp = true
+    config1.p2pConfig.filterAnnouncedAddresses = ['172.15.0.0/24'] // allow nodes to see each other locally for tests
```
alexcos20, line 59: shouldn't be config2 ?

```suggestion
    config2.p2pConfig.filterAnnouncedAddresses = ['172.15.0.0/24'] // allow nodes to see each other locally for tests
```

alexcos20 (approved):
Although I do not fully support this because of two reasons:
 1. users should learn networking, otherwise they will not understand what's happening
 2. this breaks complex setups, like private networks of nodes or  semi private (a couple of nodes inside an interna network that are using some relay servers to protect them selves from outside)
 
 I will approve it just because it makes life eaiser for noobies .
 
 One small suggestion and we are good to go

src/test/unit/oceanP2P.test.ts
paulo-ocean: yeah :-) .... damn copy/paste

PR #701, opened 2024-09-27
improve error message, transfer fees and tweak node response

### LINKED ISSUE #666: Transfer fees error messages should be more clear
If some error occurs while trying to transfer collected fees, we show a very generic error message.
We should use the error message from the `directCommand` response to provide a clear picture.
Also, so improvements must be made to the dashboard to avoid thinks like putting an unsupported chain id (should be select box of supported ones only)

Screenshot from 2024-08-30 11-31-37
Screenshot from 2024-08-30 11-30-19

paulo-ocean:
Fixes #666 .

Changes proposed in this PR:

- properly sets the node status message in case of errors (node)
- shows proper/clear error message on transfer fees issues (cpanel)
-


jamiehewitt15 (approved):
Looks good to me

paulo-ocean:
forgot to add an ex screenshot
Screenshot from 2024-09-27 11-01-23

PR #702, opened 2024-09-28
move c2d engines under OceanNode class

alexcos20:
Prepare for upcoming new C2D Engines


src/OceanNode.ts
```diff
@@ -10,25 +10,29 @@ import StreamConcat from 'stream-concat'
...
   // requester
   private remoteCaller: string | string[]
   // eslint-disable-next-line no-useless-constructor
   private constructor(
     private db?: Database,
     private node?: OceanP2P,
     private provider?: OceanProvider,
-    private indexer?: OceanIndexer
+    private indexer?: OceanIndexer,
+    private config?: OceanNodeConfig
```
paulo-ocean, line 29: Imo, i think it would be better if we keep the config out... we previously had the config being passed around between all components and that lead to some issues. We can access the config from anywhere. Why not pass the `C2DEngines` instance on the OceanNode constructor like the other components? e.g:
```
private db?: Database,
private node?: OceanP2P,
private indexer?: OceanIndexer,
private c2dEngines?: C2DEngines
```
On node startup ( index.ts ) we are reading the configuration and setting all the components given the configuration, i think we should keep the same approach for C2D engines

alexcos20: Done in https://github.com/oceanprotocol/ocean-node/pull/702/commits/452ab0af88fb5780eb94be4ff50ea66c02b3507c

(removed config and only using addEngines)

paulo-ocean (approved):

PR #707, opened 2024-10-03
Database tests 2db

paulo-ocean:
Fixes # .

Changes proposed in this PR:

- Refactor, clean and fix tests, on top of changes for Elastic Search support branch
- Removed DB specific config/env override from individual test suites (DB is only configured **ONCE** on the hooks)
- DB config is properly set & read from `Configuration` object (avoid constant direct access to `process.env` )
(sometimes we had different components running with different DB, on same tests, and some tests only passed because they "caught" a specific DB, but would not run on the other.. it was a bit random, with all the` .env `changes going on)
- All tests suites run on the same DB configuration picked at startup  (sometimes can be Elastic, others can be Typesense, but is DOES NOT change until the end)
- cleaned some old tests


paulo-ocean:
there is a segmentation fault on the unit tests, checking

paulo-ocean:
> there is a random segmentation fault on the unit tests, checking

apparently its was an issue related with the storage tests, seems fixed now

docs/env.md
```diff
@@ -75,6 +75,10 @@ Environmental variables are also tracked in `ENVIRONMENT_VARIABLES` within `src/
 - `P2P_AUTODIALINTERVAL`: Auto dial interval (miliseconds). Amount of time between close and open of new peer connection. Default: 5000
 - `P2P_ENABLE_NETWORK_STATS`: Enables 'getP2pNetworkStats' http endpoint. Since this contains private informations (like your ip addresses), this is disabled by default
 
+## Policy Server
+
+- `POLICY_SERVER_URL`: URI definition of PolicyServer, if any
```
jamiehewitt15, line 80:
```suggestion
- `POLICY_SERVER_URL`: URI definition of PolicyServer, if any. See [the policy server documentation for more details](docs/PolicyServer.md).
```

paulo-ocean: that's already on main branch and not related with this PR, i can update it, but this should be out of scope

docs/PolicyServer.md
```diff
@@ -0,0 +1,111 @@
...
+
+For every command, Ocean Node will query PolicyServer (if such env is defined) and wait for it to perform all needed checks.
+
+For 200 OK responses, Ocean Node will continue to perform the action. For everything else, it will deny. If there is a body in response, we will forward that body to the caller. (so users can see the PolicyServer error messages and act accordingly)
+
+Every Ocean Node command will also accept a data field, called "policyServer" which will be added to the query (so we can pass data from the user to PolicyServer)
+
+## PolicyServer API definition
+
+All queries will be performed by sending a POST request to PolicyServer Endpoint, with a json payload that looks like this:
```
jamiehewitt15, line 15:
```suggestion
# Policy Server

Sometimes, actions performed by Ocean Node have to be double-checked on a higher level of authorization. This might include Oath tokens, SSI verifiable credentials, Enterprise LDAP, etc...

For this, we will adopt a simple, but flexible architecture:

For every command, Ocean Node will query PolicyServer (if such env is defined) and wait for it to perform all needed checks.

For 200 OK responses, Ocean Node will continue to perform the action. For everything else, it will deny. If there is a body in response, we will forward that body to the caller. (so users can see the PolicyServer error messages and act accordingly)

Every Ocean Node command will also accept a data field, called "policyServer" which will be added to the query (so we can pass data from the user to PolicyServer)

## PolicyServer API definition

All queries will be performed by sending a POST request to PolicyServer Endpoint, with a json payload that looks like this:
```

docs/PolicyServer.md
```diff
@@ -0,0 +1,111 @@
+# Policy Server
+
+Sometimes, actions performed by Ocean Node have to be double-checked on a higher level of authorization. This might include Oath tokens, SSI verifiable credentials, Enterprise LDAP, etc...
+
+For this, we will adopt a simple, but flexible arhitecture:
```
jamiehewitt15, line 5:
```suggestion
For this, we will adopt a simple, but flexible architecture:
```

src/components/database/DatabaseFactory.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-unused-vars */
```
jamiehewitt15, line 1: Why did you add this? If we have unused vars, it would be better to remove them?

paulo-ocean: that's for further refactor regarding `enum DatabaseCollections`, because we don't have a list off all the tables we have, will do that, so we can easily update/check , but they should be used, will update that

paulo-ocean: on second though, will remove it instead.. not worth at this time i think

bogdanfazakas (approved):
lgtm, there are few comments that maybe we need to remove with old db setup from before and after hooks, and something is looking strange with the unit tests 🤔 
also i'll make the enterprise team aware of the change of the Indexer components

jamiehewitt15:
Are the tests failing due to the changes in this PR or is something else random going on? Other than that it all looks good.

src/components/Indexer/index.ts
```diff
@@ -85,15 +90,83 @@ export class OceanIndexer {
     return false
   }
 
+  // it does not start crawling until the network connectin is ready
+  async startCrawler(blockchain: Blockchain): Promise<boolean> {
```
bogdanfazakas, line 94: was there a reason why you moved this to index from the thread, or was just one of the things you tried. Asking from curiosity mainly 😄 ?

src/components/Indexer/crawlerThread.ts
```diff
@@ -285,48 +285,60 @@ export function checkNewlyIndexedAssets(events: BlocksEvents): void {
...
-      await sleep(retryInterval)
-      // recursively call the same func
-      return retryCrawlerWithDelay(blockchain, retryInterval)
-    }
-  } catch (err) {
-    INDEXER_LOGGER.error(`Error starting crawler: ${err.message}`)
-    return false
-  }
-}
+// async function retryCrawlerWithDelay(
```
bogdanfazakas, line 288: also if we decide to keep in the index lets remove it for real from the thread

src/test/unit/indexer/indexer.test.ts
```diff
@@ -1,64 +1,28 @@
...
-    // eslint-disable-next-line no-unused-expressions
-    expect(mockWorker.postMessage.calledOnceWith({ method: 'start-crawling' })).to.be
-      .false
-    // eslint-disable-next-line no-unused-expressions
-    expect(mockWorker.on.calledThrice).to.be.false
+  let mockDatabase: Database
+  let config: OceanNodeConfig
+  before(async () => {
+    config = await getConfiguration(true)
+    mockDatabase = await new Database(config.dbConfig)
```
bogdanfazakas, line 14: with this this test starts to look more like an integration test, not 100% sure if we should keep it or lets say keep it under unit tests 😃 , but nothing critical from my point of view

src/components/Indexer/index.ts
paulo-ocean: Both :-)
Remember this? [https://github.com/oceanprotocol/ocean-node/pull/551](https://github.com/oceanprotocol/ocean-node/pull/551)
When the thread exits we CANNOT process any more code inside, that's what originates the segmentation fault / core dumps
Everything works fine when we have proper DB and RPC connections, but when we don't we get a lot of errors, From DB errors to RPC errors.. Many of these cause the thread to exit, and we get that error code 1 on 'exit'..
If we exit the thread and are still trying to connect to a provider (wrong rpc for instance), that is one example.
Either by us retrying the connection (checking if is ready), or the JSONRPCProvider itself, like here:
[https://github.com/ethers-io/ethers.js/blob/9e7e7f3e2f2d51019aaa782e6290e079c38332fb/src.ts/providers/provider-jsonrpc.ts#L767](https://github.com/ethers-io/ethers.js/blob/9e7e7f3e2f2d51019aaa782e6290e079c38332fb/src.ts/providers/provider-jsonrpc.ts#L767)
when we create a` new JSONRPCProvider` it retries at least 4 or 5 times.. We can't have any of this code on the crawler thread.

src/test/unit/indexer/indexer.test.ts
paulo-ocean: not really :-) , its just creating an instance of the class and calling a few public functions, not much different than all the others on the unit test suite

mariacarmina (approved):
LGTM

PR #716, opened 2024-10-14
Bump version axios 1.6.0 -> 1.7.4.

mariacarmina:
Fixes # .

Changes proposed in this PR:

- 
- 
-

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -45,7 +45,7 @@
         "@types/rdfjs__to-ntriples": "^2.0.6",
         "@zazuko/env-node": "^2.1.3",
         "aegir": "^37.3.0",
-        "axios": "^1.6.0",
+        "axios": "^1.7.4",
         "base58-js": "^2.0.0",
         "cors": "^2.8.5",
         "delay": "^5.0.0",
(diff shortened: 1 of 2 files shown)


mariacarmina:
Unit tests failing on this PR dues to Segmentation fault error :(

paulo-ocean:
> Unit tests failing on this PR dues to Segmentation fault error :(

Same issue already mentioned on another PR
Check these log messages:
JsonRpcProvider failed to detect network and cannot start up; retry in 1s (perhaps the URL is wrong or the node is not `started)`
The main issue is that we have the crawler thread starting when we don't have a DB connection or a proper RPC connection (wrong behaviour)...
Inside the crawler thread we have a mechanism to retry the RPC/Blockchain connection... that can stay there forever (keep retrying)... Eventually it can kill the memory...
This was already addressed on the Elastic PR..
The crawler thread should not start without a DB connection and an RPC connection ... Only after... Eventually it should not keeping retrying RPC connection forever, but a few times  ( Max 10 times )

paulo-ocean (approved):

jamiehewitt15 (approved):
Looks good to me

PR #723, opened 2024-10-16
nonce db sql lite 

### LINKED ISSUE #721: Make sure Nonce Database is always SQL LITE
Right now, if we have Typesense configured but it fails, we default to SQL Lite
But if we Have Elastic Search configured, the same thing does not happen.
Following some internal discussions we decided that SQL Lite is more than enough for Nonce DB (just a number and an address basically) 
So, this particular database should always be SQL Lite

paulo-ocean:
Fixes #721 .

Changes proposed in this PR:

- nonce db always SQL Lite
- 
-


mariacarmina (approved):
LGTM

bogdanfazakas (approved):
Lgtm

PR #733, opened 2024-10-24
fix system tests. Running old version of node?

### LINKED ISSUE #732: Check system-test on main branch
There is something weird going on o main branch..

I don't see this log:
`Nonce Database initiated with SQLite provider`. It seems like the requests go to Typesense Nonce?
But on any other branch created from main, the system tests do not fail, the log message is present, and everything works as expected
Even on main branch, this only happens on system-tests, unit tests and integration tests are green and everything is OK.. Something going on with these tests only 

UPDATE:

Seems the system tests are running an older version of the node. Check screenshots bellow... the log message
`'INDEXER_NETWORKS is not defined, running Indexer with all supported networks defined in RPCS env variable ...'`
Shows as an ERROR on the failing tests, but on main branch this log message is a INFO (and we can see the Nonce DB log too)..... Its clearly not fetching the updated code of the node

Screenshot from 2024-10-24 10-44-08
Screenshot from 2024-10-24 10-43-01

paulo-ocean:
Fixes #732 

Changes proposed in this PR:

- system tests on main branch are running ocean node from "develop" branch, which is wrong and outdated
- 
-

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - main
-      - develop
     tags:
       - '**'
   pull_request:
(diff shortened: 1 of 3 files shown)


bogdanfazakas (approved):

mariacarmina (approved):
lgtm

Issue #740, opened 2024-10-29
Node's Indexer does not resume indexing after stoping the node


mariacarmina:
**Describe the bug**
Node's Indexer does not resume indexing after stoping the node (with Ctrl + C or Cmd + C). The expected behaviour is to resume the last indexed block, not starting from deployer block. The issue is fixed if the startBlock is specified within RPCS env var, but it is not user intuitive/friendly to keep updating startBlock.

**To Reproduce**
1. Run Ocean Node on local with npm run start
2. Wait the node to sync
3. Stop the node
4. Restart node

**Expected behavior**
The expected behaviour is to resume the last indexed block, not starting from deployer block.

**Logs**
[ocean-node-op-sepolia-indexer.txt](https://github.com/user-attachments/files/17556839/ocean-node-op-sepolia-indexer.txt)

md00ux:
- **ocean-node** container was started using docker compose 
- database type => typesense
```yaml
DB_TYPE: 'typesense'
DB_URL: "http://typesense:8108/?apiKey=xyz"
....
typesense:
    image: typesense/typesense:26.0
    container_name: typesense
    ports:
      - "8108:8108"
    networks:
      - backend
    volumes:
      - typesense-data:/data
    command: '--data-dir /data --api-key=xyz'
```

paulo-ocean:
It looks to me that we could not get the `lastIndexedBlock` from DB for some reason.. that's the only explanation i can see by looking at the code.. if for some reason we fail to get that info, it will start from either `rpcDetails.startBlock` or `contractDeploymentBlock` (whatever is more up to date), but that is expected, as we cannot guess the last indexed block..
I would say the problem here was at DB/connection level

mariacarmina:
> It looks to me that we could not get the `lastIndexedBlock` from DB for some reason.. that's the only explanation i can see by looking at the code.. and the logs... The expected behaviour is Ok.
> 
> But if for some reason we fail to get that info, it will start from either `rpcDetails.startBlock` or `contractDeploymentBlock` (whatever is more up to date), but that is expected, as we cannot guess the last indexed block.. The `crawlingStartBlock` variable is calculated from these
> 
> I would say the problem here was at DB/connection level. Maybe we had no data there on the restart? We lost it?
> 
> @md00ux can we have the full logs?
> 
> 
> 
> We can easily improve the logging, so we know exactly what we are reading:
> 
> ```
> 
[...]

PR #743, opened 2024-10-29
Add cleaner logging when idexing and getting/updating block info from db

### LINKED ISSUE #742: Improve logging when start indexing chain
When we start indexing we should also log the last indexed block retrieved from the database.
Additionally we should also log the start block and deployment block if available/supplied
This can help clarify issues with blocks

paulo-ocean:
Fixes #742 #740 

Changes proposed in this PR:

- imporve logging a bit on indexer thread (block related stuff) for clearity
- minor refactor on a couple of functions to log and avoid dealing with '`nulls`'
-


jamiehewitt15 (approved):
Looks good to me

PR #744, opened 2024-10-30
fix dashboard hash out of sync

### LINKED ISSUE #664: Dashboard  files marked as modified after a simple build
The tracking of the dashboard files is still not correct and needs improvement
Starting from a clean branch and doing a simple: `npm run build`
marks a bunch of dashboard files under `/dist/dashboard` as modified
But i'm not changing anything on the dashboard or main node sources, i'm just doing a local build
I should not have a bunch of git modified/deleted files (example bellow)
Screenshot from 2024-08-30 09-45-53

paulo-ocean:
Fixes #664 

Changes proposed in this PR:

- The dashboard hash was not updated the last time we changed something on the dashboard (+ details on issue)
- when we update anything (any file) inside `dashboard` directory, we need to re-build to re-calculate the hash and commit the changes as well, sometimes we forget that
-


jamiehewitt15 (approved):
Looks good to me

mariacarmina (approved):

PR #747, opened 2024-11-01
improve reindex command, optional block, also check if thread is running

### LINKED ISSUE #741: Add optional block number to Reindex Chain admin command
Right now, when we reindex the chain, it either starts from the `deployed` block or the `startBlock` (if defined on `RPCS`)
we could optionally have another parameter to indicate a specific "block number" to start the indexing from (of course we need to validate it against deployed block, network height, etc)

paulo-ocean:
Fixes #741 

Changes proposed in this PR:

- optional block number on reindex chain command
- check if thread is running (otherwise it does not get any messages)
- if not running, try to restart it first (needs a new reference)

### DIFF
--- a/src/@types/commands.ts
+++ b/src/@types/commands.ts
@@ -132,6 +132,7 @@ export interface AdminCollectFeesCommand extends AdminCommand {
 
 export interface AdminReindexChainCommand extends AdminCommand {
   chainId: number
+  block?: number
 }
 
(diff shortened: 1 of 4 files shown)


jamiehewitt15 (approved):
Looks good to me

mariacarmina (approved):
lgtm

PR #750, opened 2024-11-12
check if rpcs and indexer_networks are aligned, log

### LINKED ISSUE #728: If RPCS and INDEXER_NETWORKS are not aligned we should show a log message
`INDEXER_NETWORKS` takes precedence over `RPCS`...
if the INDEXER_NETWORKS contains a different chain than the ones defined on RPCS (or if the chains from RPCS are not present on INDEXER_NETWORKS)
It does not index anything, but there are no logs or anything warning about the .env mismatch..
at least a warning message would be nice, since the 2 .env vars if present, need to be aligned...
This can generate some confusion

paulo-ocean:
Fixes #728 

Changes proposed in this PR:

- make sure we at least do some logging when RPCS and INDEXER_NETWORKS have mismatches (making the node to run without indexer)
- 
-

### DIFF
--- a/src/utils/config.ts
+++ b/src/utils/config.ts
@@ -102,9 +102,12 @@ function getSupportedChains(): RPCS | null {
 
 function getIndexingNetworks(supportedNetworks: RPCS): RPCS | null {
   const indexerNetworksEnv = process.env.INDEXER_NETWORKS
+
+  const defaultErrorMsg =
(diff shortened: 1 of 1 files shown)


jamiehewitt15 (approved):
Looks good to me

mariacarmina (approved):
lgtm

PR #752, opened 2024-11-15
try debug and fix randomly logs failing test

### LINKED ISSUE #676: Random failing test in the CI
We're getting another failing integration test in the CI. It's an issue with the log tests again:

```
  111 passing (5m)
  1 failing

  1) LogDatabase retrieveMultipleLogs with pagination
       should retrieve logs for a specific page:

      AssertionError: expected '651' to not equal '651'
      + expected - actual


      at Context.<anonymous> (file:///home/runner/work/ocean-node/ocean-node/dist/test/integration/logs.test.js:284:44)
      at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```

paulo-ocean:
Fixes #676 

Changes proposed in this PR:

- The issue doesn't seem to be the test, but the feature itself. At least with typesense

### DIFF
--- a/src/test/integration/logs.test.ts
+++ b/src/test/integration/logs.test.ts
@@ -467,7 +467,7 @@ describe('LogDatabase retrieveMultipleLogs with pagination', () => {
       undefined,
       2 // Page 2
     )
-    console.log('LOGS_PAGE 2: ', logsPage1)
+    console.log('LOGS_PAGE 2: ', logsPage2)
     // make sure we have enough logs for 2 pages
     const logsCount = await database.logs.getLogsCount()
     console.log('LOGS_COUNT: ', logsCount)
--- a/src/test/integration/logs.test.ts
+++ b/src/test/integration/logs.test.ts
@@ -7,7 +7,7 @@ import {
   configureCustomDBTransport,
   getCustomLoggerForModule
 } from '../../utils/logging/Logger.js'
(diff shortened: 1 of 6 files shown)


mariacarmina (approved):
lgtm

src/test/integration/logs.test.ts
```diff
@@ -87,11 +87,11 @@ describe('LogDatabase CRUD', () => {
     if (logs.length > 0) {
       logs = logs.filter((log) => log.message === newLogEntry.message)
 
-      expect(logs?.length).to.equal(1)
-      expect(Number(logs?.[0].id)).to.greaterThan(Number(logId))
-      expect(logs?.[0].level).to.equal(newLogEntry.level)
-      expect(logs?.[0].message).to.equal(newLogEntry.message)
-      expect(logs?.[0].moduleName).to.equal('HTTP')
```
jamiehewitt15, line 94: What's the point of deleting these checks?

paulo-ocean: For having a cleaner API and kind of standard response... In this case we should always return an array.. either empty or with elements, instead of checking for null and also array length
Having ambiguous APIs where we can have different kind of results leads to unnecessary code branching and can hide some issues

jamiehewitt15 (approved):
Looks good to me

PR #761, opened 2024-11-26
Removing unused express dependency from dashboard

### LINKED ISSUE #760: Remove Express dependency from dashboard
Remove Express dependency from dashboard

jamiehewitt15:
Fixes #760 
Fixes: https://github.com/oceanprotocol/ocean-node/issues/765

Changes proposed in this PR:

- Removing unused express dependency from the dashboard
- Updating the static dashboard files


scripts/docker-compose.yml
```diff
@@ -0,0 +1,90 @@
+services:
```
paulo-ocean, line 1: Why are we including this file? Isn't this generated by the quick start script?
i think we should remove it and add it to gitignore

jamiehewitt15: The dashboard script (`scripts/dashboardChanges.js`) takes a hash of the dashboard directory and then compares it to this hash. If they are different it builds the dashboard again. If you don't have this file the dashboard will be built every time.

jamiehewitt15: yeah, good point. I'll add it to gitignore

paulo-ocean (approved):

mariacarmina (approved):
lgtm

PR #766, opened 2024-11-27
omit db compute job fields on get status and start compute

### LINKED ISSUE #763: getComputeStatus command exposes too much information (DBComputeJob instead of ComputeJob)
As far as I remember for user usage, the job id with hash should be used & exposed. In fact it can be a bit misleading for end user.
 
Screenshot 2024-11-26 at 12 34 58

paulo-ocean:
Fixes #763 

Changes proposed in this PR:

- Removes `DBComputeJob` specific fields from response (only `ComputeJob` fields)
- utility function for deleting keys
- unit test


src/test/unit/compute.test.ts
```diff
@@ -177,6 +180,31 @@ describe('Compute Jobs Database', () => {
     expect(convertStringToArray(str)).to.deep.equal(expectedArray)
   })
 
+  it('should convert DBComputeJob to ComputeJob and omit internal DB data', () => {
```
mariacarmina, line 183: Can we have a check for jobId as well? What value will take, I've seent that in deleteKeysFromObject jobId is not mentioned. What I want to address, it's that the issue that I've created targets especially the jobId and what I see in this PR targets the cleanup in other response fields of compute jobs.
Am I missing any info here?
Thank you!

mariacarmina (commented):
Hello @paulo-ocean, a question on my side regarding the fields modifications. Should we update the issue as well or the code in this PR?

src/test/unit/compute.test.ts
paulo-ocean: the issue you reported on the PR is not an issue.. we addressed that yesterday... the internal job id has nothing to do with the cluster hash... the value we return to the user is cluster hash + job id ... but that doesn't change the ComputeJob interface... the real issue is that we should always return a `ComputeJob` not a `DBComputeJob`, therefore we need to strip out those fields

paulo-ocean: but its better to update the text on the issue, for sure.. will do that. thanks :+1:

jamiehewitt15 (approved):
Looks good to me

PR #774, opened 2024-12-04
Fix DDO: Stats and Prices for exchanges/dispensers 

### LINKED ISSUE #525: DDO: Stats and Prices for exchanges/dispensers
In order to proper display assets in dApps, we need all info regarding assets, including prices.
Every datatoken can have one or more fixed rate exchanges or dispensers.
But we need to be careful when adding this to a DDO, cause this can quickly result to hash missmatches.

To Do:
 - add a root key at ddo level, called offchain.  This key will be excluded every time we hash the ddo (like validateDDO or when indexing and comparing)
 - in indexedMetadata, add a new key: stats.      Then, for each service, add an array that looks like:
 ```javascript
 {
   datatokenAddress: "0x123",
   name: "Branin dataset: DT1",
   symbol: "DT1",
   serviceId: "0",
   orders: 1,
   prices:[
        {
            type: "fixedrate", // or "dispenser"
            price: "123",
            contract: "contractAddress",
            token: "0x89777", // token accepted for payment
            exchangeId:  "xx" // only for fre, this is exhangeID
    }
   ]
}
```

So , a DDO can look like:
```javascript
{
  "@context": [
    "https://w3id.org/did/v1"
  ],
  "id": "did:op:37f710f82f73afcb86b66a32182c2d620d2edd9eb00abb5ba5a36e943bc7cd74",
  "version": "4.1.0",
  "chainId": 137,
  "nftAddress": "0x26C693AD67d45e339A5A50a5a64963892E14EFdD",
  "metadata": {
              ...
  },
  "services": [
    {
      "name": "Download last hour of data",
      "id": "0",
      "type": "access",
      "datatokenAddress": "0x456",
     .....
    },

    {
      "name": "Download last week of data",
      "id": "1",
      "type": "access",
      "datatokenAddress": "0x34e84f653Dcb291838aa8AF8Be1E1eF30e749ba0",
     .....
    },
    {
      "name": "Download last month of data",
      "id": "2",
      "type": "access",
      "datatokenAddress": "0x1234565",
     .....
    }

  ],
 indexedMetadata: {
    stats: {
        [
            {
                datatokenAddress: "0x456",
                name: "BDT1",
                symbol: "DT1",
                serviceId: "0",
                orders: 1,
                // this service has one dispenser available
                prices: [
                    {  type: "dispenser", price: "0", contract: "dispenserContractAddress"}]

            },
            {
                datatokenAddress: "0x34e84f653Dcb291838aa8AF8Be1E1eF30e749ba0",
                name: "BDT2",
                symbol: "DT2",
                serviceId: "1",
                orders: 5,
                // this service accepts OCEAN for payment, costs 1 token per access
                prices:[
                            {  type: "fixedrate", price: "1", token:"0x967da4048cD07aB37855c090aAF366e4ce1b9F48", contract: "freContractAddress",  exchangeId:  "23434" }
                        ]
            },
            {
                datatokenAddress: "0x1234565",
                name: "BDT3",
                symbol: "DT3",
                serviceId: "2",
                orders: 5,
                // this service accepts either 2 OCEAN or 1 FETCH for payment
                prices:[  
                   {  type: "fixedrate", price: "2", token:"0x967da4048cD07aB37855c090aAF366e4ce1b9F48", contract: "freContractAddress",  exchangeId:  "23434" },
                  {  type: "fixedrate", price: "1", token:"0xaea46A60368A7bD060eec7DF8CBa43b7EF41Ad85", contract: "freContractAddress",  exchangeId:  "4535" }
             ]
            }
        ]
    }
}            
}
```

To do:

-[ ] exclude "offchain" from every hash compute

-[ ] for MetaDataCreated:  

    - For each service:
        - get the list of dispensers (call getDispensers() - https://github.com/oceanprotocol/contracts/blob/main/contracts/templates/ERC20TemplateEnterprise.sol#L1107C14-L1107C27) from datatoken contract
            - for each dispenser, check if active. if yes, add it to the list
        - get the list of fixed rates (call getFixedRates())
            - for each fixed rate, check if active and grab the price, then add it to the list
    - Update stats key

-[ ] for MetaDataUpdated:  

    - For each service:
        - get the list of dispensers (call getDispensers()) from datatoken contract
            - for each dispenser, check if active. if yes, add it to the list
        - get the list of fixed rates (call getFixedRates())
            - for each fixed rate, check if active and grab the price, then add it to the list
    - Remove services from stats which are no longer present
    - Update stats key

-[ ] indexer: add monitor for DispenserActivated and DispenserDeactivated events (https://github.com/oceanprotocol/contracts/blob/main/contracts/pools/dispenser/Dispenser.sol#L39-L45).  If such an event is detected, get all ddos with a match and update it (remove dispenser/add dispenser)

-[ ] indexer: add monitor for ExchangeActivated and ExchangeDeactivated events (https://github.com/oceanprotocol/contracts/blob/main/contracts/pools/fixedRate/FixedRateExchange.sol#L103-L111).  If such an event is detected, get all ddos with a match and update it (remove dispenser/add fixed rate)

-[ ] indexer: add monitor for ExchangeRateChanged event(https://github.com/oceanprotocol/contracts/blob/main/contracts/pools/fixedRate/FixedRateExchange.sol#L90C11-L94).  If such an event is detected, get all ddos with a match and update them with the new rate

-[ ] indexer: add monitor for OrderStarted event(https://github.com/oceanprotocol/contracts/blob/main/contracts/templates/ERC20TemplateEnterprise.sol#L85-L93).  If such an event is detected, get all ddos with a match and update orders count

mariacarmina:
Fixes #525  .

Changes proposed in this PR:

- Create type for IndexedMetadata
- Delete key when validating DDO
-

### DIFF
--- a/src/@types/DDO/IndexedMetadata.ts
+++ b/src/@types/DDO/IndexedMetadata.ts
@@ -1,0 +1,19 @@
+export interface Price {
+  type: string
+  price: string
+  contract: string
+  token?: string
+  exchangeId?: string
+}
+
+export interface Stats {
+  datatokenAddress: string
+  name: string
+  serviceId: string
+  orders: number
+  prices: Price[]
+}
+
+export interface IndexedMetadata {
+  stats: Stats[]
+}
--- a/src/@types/DDO/DDO.ts
+++ b/src/@types/DDO/DDO.ts
@@ -2,6 +2,7 @@ import { Service } from './Service'
 import { Metadata } from './Metadata'
 import { Credentials } from './Credentials'
 import { Event } from './Event'
+import { IndexedMetadata } from './IndexedMetadata'
 import { Nft } from './Nft'
 
 /**
@@ -67,4 +68,6 @@ export interface DDO {
   event?: Event
 
   nft?: Nft
+
+  indexedMetadata?: IndexedMetadata
 }
--- a/src/components/core/utils/validateDdoHandler.ts
+++ b/src/components/core/utils/validateDdoHandler.ts
@@ -49,12 +49,24 @@ export function makeDid(nftAddress: string, chainId: string): string {
   )
 }
 
+export function deleteIndexedMetadataIfExists(
+  ddo: Record<string, any>
+): Record<string, any> {
+  let ddoCopy: Record<string, any> = ddo
+  if ('indexedMetadata' in ddoCopy) {
+    delete ddoCopy.indexedMetadata
+    return ddoCopy
+  }
+  return ddo
+}
+
 export async function validateObject(
   obj: Record<string, any>,
   chainId: number,
   nftAddress: string
 ): Promise<[boolean, Record<string, string[]>]> {
-  const ddoCopy = JSON.parse(JSON.stringify(obj))
+  const updatedDdo = deleteIndexedMetadataIfExists(obj)
+  const ddoCopy = JSON.parse(JSON.stringify(updatedDdo))
   ddoCopy['@type'] = 'DDO'
 
   const extraErrors: Record<string, string[]> = {}


mariacarmina:
For the moment only ocean-cli test regarding getDDO fails:
To be more precise at this check https://github.com/oceanprotocol/ocean-cli/blob/main/test/consumeFlow.test.ts#L144
```
Ocean CLI Publishing
    ✔ should publish a dataset using 'npm run cli publish' (3446ms)
    ✔ should publish a compute dataset using 'npm run cli publish' (3099ms)
    ✔ should publish a js Algorithm using 'npm run cli publishAlgo' (3025ms)
    ✔ should publish a python Algorithm using 'npm run cli publishAlgo' (3042ms)
    1) should get DDO using 'npm run cli getDDO' for download dataset


  4 passing (3m)
  1 failing

  1) Ocean CLI Publishing
       should get DDO using 'npm run cli getDDO' for download dataset:
     Uncaught AssertionError: expected '\n> ocean-cli@2.0.0 cli\n> TS_NODE_PR…' to include 'https://w3id.org/did/v1'
      at /home/runner/work/ocean-node/ocean-node/ocean-cli/test/consumeFlow.test.ts:144:31
      at ChildProcess.exithandler (node:child_process:414:7)
      at ChildProcess.emit (node:events:5[19](https://github.com/oceanprotocol/ocean-node/actions/runs/12207966655/job/34060344250#step:20:20):28)
      at ChildProcess.emit (node:domain:488:12)
      at maybeClose (node:internal/child_process:1105:16)
      at Process.ChildProcess._handle.onexit (node:internal/child_process:305:5)
```

I'll put this in review to see if for Metadata events and Order events, I implemented as expected on the specs.
Still FixedRateExchange and Dispenser events needs to be monitored (WIP) along with tests for fixed rate exchange & dispenser pricing schemas.

jamiehewitt15:
Have you looked into why the tests are failing?

mariacarmina:
> Have you looked into why the tests are failing?

Yes, of course I looked into the failure. They fail because on ExchangeCreate event decodedData of the event is null and I cannot retrieve the exchange id -> the event data is not parsed correctly or the abi is wrong, because the event and receipt exist on the chain.

mariacarmina:
> > Have you looked into why the tests are failing?
> 
> Yes, of course I looked into the failure. They fail because on ExchangeCreate event decodedData of the event is null and I cannot retrieve the exchange id -> the event name is wrong

I see the root cause is that in `getEventData`, we hardcode to retrieve the first log from tx receipt, when we call a bundled function (for e.g. deploying an nft, a erc20 and a fre), threre are multiple events within that tx and by default it retrives the first log which is not correct, I have added a workaround to check topics[0] from logs[i] to match the event hash that we are looking for.

mariacarmina:
The tests regarding download stream fails again with "stream not present"... will look into it

mariacarmina:
I'll bring back only the tests regarding DDO pricing just for highlighting that the functionality is working for this PR. The failing test regarding file stream will be tackled in a separate PR.

paulo-ocean (commented):
these values are NOT ok.. maybe they were already wrong before, but
the test timeout MUST be `bigger` or `equal` than the timeout passed to the `waitToIndex`. There is a least one other test with the same description that is also wrong)

ex:
`this.timeout(DEFAULT_TEST_TIMEOUT * 5)`  
(this means that the entire test will timeout in 100... 20 * 5 seconds)

AND 
```
 const { ddo, wasTimeout } = await waitToIndex(
      assetDID,
      EVENTS.METADATA_UPDATED,
      DEFAULT_TEST_TIMEOUT * 8
    )
```
This means we will wait for indexer UNTIL MAX 155 ... 20 * 8 - 5 seconds 

So, we would wait for 155 seconds... but the test itself expires in 100 seconds... This way we never catch a `expectedTimeoutFailure`
So, the timeout for the `waitToIndex` function should always be equal or less than the entire test timeout (preferably it should be `equal` , which already makes it 5 seconds shorter)

In any case, the timeouts here are very big... i doubt that is necessary

src/test/integration/pricing.test.ts
```diff
@@ -342,11 +342,11 @@ describe('Publish pricing scehmas and assert ddo stats - FRE & Dispenser', () =>
     assert(setMetaDataTxReceipt, 'set metada failed')
   })
   it('should store the ddo in the database and return it ', async function () {
-    this.timeout(DEFAULT_TEST_TIMEOUT * 2)
+    this.timeout(DEFAULT_TEST_TIMEOUT * 5)
```
paulo-ocean, line 345: these values are NOT ok.. maybe they were already wrong before, but
the test timeout MUST be `bigger` or `equal` than the timeout passed to the `waitToIndex`. There is a least one other test with the same description that is also wrong)

ex:
`this.timeout(DEFAULT_TEST_TIMEOUT * 5)`  
(this means that the entire test will timeout in 100... 20 * 5 seconds)

AND 
```
 const { ddo, wasTimeout } = await waitToIndex(
      assetDID,
      EVENTS.METADATA_UPDATED,
      DEFAULT_TEST_TIMEOUT * 8
    )
```
This means we will wait for indexer UNTIL MAX 155 ... 20 * 8 - 5 seconds 

So, we would wait for 155 seconds... but the test itself expires in 100 seconds... This way we never catch a `expectedTimeoutFailure`
So, the timeout for the `waitToIndex` function should always be equal or less than the entire test timeout (preferably it should be `equal` , which already makes it 5 seconds shorter)

In any case, the timeouts here are very big... i doubt that is necessary

src/@types/DDO/IndexedMetadata.ts
```diff
@@ -0,0 +1,19 @@
+export interface Price {
+  type: string
```
paulo-ocean, line 2: Afaik, the type can be either `fixedrate` or `dispenser`, but we're not setting that constraint , so nothing stops us from setting a price type to `'something else'`
we could have something like:
```
export type PriceType = 'fixedrate' | 'dispenser'
export interface Price {
  type: PriceType
...
```
This would avoid any eventual typos and  would be cleaner

src/@types/DDO/IndexedMetadata.ts
```diff
@@ -0,0 +1,19 @@
...
+  price: string
+  contract: string
+  token?: string
+  exchangeId?: string
+}
+
+export interface Stats {
+  datatokenAddress: string
+  name: string
+  serviceId: string
```
paulo-ocean, line 12: i think we're missing the `symbol` key

src/components/Indexer/processor.ts
```diff
@@ -421,10 +510,69 @@ export class MetadataEventProcessor extends BaseEventProcessor {
...
+              try {
+                const dispensers = await datatoken.getDispensers()
+                for (const dispenser of dispensers) {
+                  const dispenserContract = new ethers.Contract(
+                    dispenser,
+                    Dispenser.abi,
+                    signer
+                  )
+                  if (
+                    (await dispenserContract.status(await datatoken.getAddress()))[0] ===
```
paulo-ocean, line 541: It seems to me, we overcomplicated this part..
we do this a few lines above:
` const ddoWithPricing = await getPricingStatsForDddo(ddo, signer)` (we add everything that is `active` to the lists)
and then we are doing again the same kind of check , only to exclude the '`non active`'  (actually just to do the opposite)? splicing from the array?
why is that?

Its enough to only add the `active` ones, doesn't make much sense to add only the actives and then try to remove the non actives from there ..
the blocks:
```
if (stat.type === 'dispenser') {
              try {
                const dispensers = await datatoken.getDispensers()
                for (const dispenser of dispensers) {
                 ....
} else if (stat.type === 'fixedrate') {
...
```
are not doing anything, unless the `getPricingStatsForDddo` is not working correctly
so we can safely remove these blocks

src/components/Indexer/processor.ts
```diff
@@ -421,10 +510,69 @@ export class MetadataEventProcessor extends BaseEventProcessor {
...
-      const from = decodedEventData.args[0]
+      const from = decodedEventData.args[0].toString()
 
       // we need to store the event data (either metadata created or update and is updatable)
-      if ([EVENTS.METADATA_CREATED, EVENTS.METADATA_UPDATED].includes(eventName)) {
+      if (
+        [EVENTS.METADATA_CREATED, EVENTS.METADATA_UPDATED].includes(eventName) &&
+        this.isValidDtAddressFromServices(ddo.services)
+      ) {
+        const ddoWithPricing = await getPricingStatsForDddo(ddo, signer)
```
paulo-ocean, line 520: It seems to me, we overcomplicated this part..
we do this a few lines above:
` const ddoWithPricing = await getPricingStatsForDddo(ddo, signer)` (we add everything that is `active` to the lists)
and then we are doing again the same kind of check , only to exclude the '`non active`'  (actually just to do the opposite)? splicing from the array?
why is that?

Its enough to only add the `active` ones, doesn't make much sense to add only the actives and then try to remove the non actives from there ..
the blocks:
```
if (stat.type === 'dispenser') {
              try {
                const dispensers = await datatoken.getDispensers()
                for (const dispenser of dispensers) {
                 ....
} else if (stat.type === 'fixedrate') {
...
```
are not doing anything, unless the `getPricingStatsForDddo` is not working correctly
so we can safely remove these blocks

src/@types/DDO/IndexedMetadata.ts
```diff
@@ -0,0 +1,22 @@
+export type PriceType = 'fixedrate' | 'dispenser'
+
+export interface Price {
```
paulo-ocean, line 3: I think we should probably refactor the interface names to `ServicePrice` (instead of `Price`)
and `ServiceStats` (instead of `Stats`) as it was also suggested on the original issue. This way we can easily understand that these fields are related to the `services`. As IS, its not very clear for others

src/components/Indexer/utils.ts
```diff
@@ -330,3 +434,189 @@ export function buildJobIdentifier(command: string, extra: string[]): JobStatus
     hash: create256Hash(extra.join(''))
   }
 }
+
+export function findServiceIdByDatatoken(ddo: any, datatokenAddress: string): string {
+  let serviceIdToFind = ''
```
paulo-ocean, line 439: why not return `null` to follow the same approach we have on similar functions? If we can't find the service ID.. much cleaner than looking for an empty string

src/components/core/handler/ddoHandler.ts
```diff
@@ -843,11 +845,12 @@ export function validateDDOIdentifier(identifier: string): ValidateParams {
  * @returns validation result
  */
 async function checkIfDDOResponseIsLegit(ddo: any): Promise<boolean> {
-  const { nftAddress, chainId, event } = ddo
-  let isValid = validateDDOHash(ddo.id, nftAddress, chainId)
+  const updatedDdo = deleteIndexedMetadataIfExists(ddo)
+  const { nftAddress, chainId, event } = updatedDdo
```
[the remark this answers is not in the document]
mariacarmina: Yes now, I'll retrieve the event from indexedMetadata before deleting the key when generating ddo hash. I'll store indexedMetadata for event checks.

src/components/Indexer/processor.ts
mariacarmina: > doesn't make much sense to add only the actives and then try to remove the non actives from there

This is writtent in the issue to update the lists to keep only active pricing schemas.

.github/workflows/ci.yml
```diff
@@ -278,7 +278,7 @@ jobs:
         with:
           repository: 'oceanprotocol/ocean-cli'
           path: 'ocean-cli'
-          ref: 'main'
+          ref: 'fix-tests-for-new-ddo'
```
jamiehewitt15, line 281: It's nice that this gets the tests passing but we don't want to merge this into main. Let's get the other PR approved (I noticed it is still a draft) so that we can merge them both at the same time with `ref: 'main'` here

src/components/Indexer/processor.ts
```diff
@@ -118,8 +201,11 @@ class BaseEventProcessor {
   }
 
   protected checkDdoHash(decryptedDocument: any, documentHashFromContract: any): boolean {
+    INDEXER_LOGGER.logMessage(`decrypted document: ${JSON.stringify(decryptedDocument)}`)
     const utf8Bytes = toUtf8Bytes(JSON.stringify(decryptedDocument))
     const expectedMetadata = hexlify(utf8Bytes)
+    INDEXER_LOGGER.logMessage(`hash: ${create256Hash(expectedMetadata.toString())}`)
+    INDEXER_LOGGER.logMessage(`ctr hash: ${documentHashFromContract}`)
```
jamiehewitt15, line 208: What's a ctr hash? I doubt anyone will know when they see this popping up in the logs, so this is essentially just clutter. If you feel that it is important, then explain it: 

```suggestion
    INDEXER_LOGGER.logMessage(`Document Hash From Contract: ${documentHashFromContract}`)
```

src/components/Indexer/processor.ts
```diff
@@ -761,3 +958,517 @@ export class OrderReusedEventProcessor extends BaseEventProcessor {
...
+
+      const savedDDO = this.createOrUpdateDDO(ddo, EVENTS.DISPENSER_DEACTIVATED)
+      return savedDDO
+    } catch (err) {
+      INDEXER_LOGGER.log(LOG_LEVELS_STR.LEVEL_ERROR, `Error retrieving DDO: ${err}`, true)
+    }
+  }
+}
+
+export class ExchangeCreatedEventProcessor extends BaseEventProcessor {
```
bogdanfazakas, line 1129: this processor.ts file tends to get bigger and bigger it might be a good idea to create new files with the classes that extend the BaseEventProcessor and maybe even for that we cand have a new sepparate file

src/components/Indexer/utils.ts
```diff
@@ -330,3 +434,187 @@ export function buildJobIdentifier(command: string, extra: string[]): JobStatus
...
+
+    if (fixedRates) {
+      for (const fixedRate of fixedRates) {
+        const fixedRateContract = new ethers.Contract(
+          fixedRate[0],
+          FixedRateExchange.abi,
+          signer
+        )
+        const exchange = await fixedRateContract.getExchange(fixedRate[1])
+        if (exchange[6] === true) {
```
bogdanfazakas, line 514: If the contract call (getExchange) fails, it will throw an unhandled error, so it would be good to wrap await fixedRateContract.getExchange in a try/catch or check if exchange exists before accessing the 7th element.
also would be nice if we can use destructuring with named variables like const [.., .., .., .., .., .., isActive] = exchange;

src/components/Indexer/processor.ts
```diff
@@ -118,8 +201,11 @@ class BaseEventProcessor {
   }
 
   protected checkDdoHash(decryptedDocument: any, documentHashFromContract: any): boolean {
+    INDEXER_LOGGER.logMessage(`decrypted document: ${JSON.stringify(decryptedDocument)}`)
```
bogdanfazakas, line 204: think we can remove this debug logs

src/components/Indexer/processor.ts
mariacarmina: These will be deleted, should we keep for debug?

src/components/Indexer/utils.ts
mariacarmina: ok, sure, thank you!

src/components/Indexer/processor.ts
mariacarmina: yes, they are for debug purposes, will clean them up in #527 since I have done multiple changes there and discovered some other bugs.

src/components/Indexer/utils.ts
mariacarmina: resolved in #527 PR, since it contains more modifications/fixes.

src/components/Indexer/processor.ts
mariacarmina: should this be tackled within this PR? It already has a lot of files changed, maybe it s a good idea to have this cosmetic change in a separate issue and I'm happy to implement it.

src/test/integration/typesense.test.ts
```diff
@@ -161,19 +161,23 @@ describe('Typesense documents', () => {
...
-      filter_by: 'chainId:<138',
-      sort_by: 'version:desc'
-    })
-
-    expect(result.found).to.equal(1)
-    expect(result.hits[0]).to.not.be.an('undefined')
-    expect(result.hits[0].document).to.not.be.an('undefined')
-  })
+  // TODO: fix search query
+  // it('search document in ddo collection', async () => {
```
bogdanfazakas, line 165: should we put this back on now? i guess it be passing before we merge this ?

mariacarmina: for some reason, if you run typesense tests independently, they pass, otherwise that search test is failing. That's why I let 2 commits, to demonstrate that the query is fixed. This test does not affect the functionality and can be tackle separately.

jamiehewitt15: I think Bogdan has a point, we should figure out why this is failing and add the test back in.

jamiehewitt15: It is quite likely that it is failing on this line:
```expect(result.found).to.equal(1)```
We may have published the same asset more than once and there are two results. In that case you could fix it with just changing the test to check for greater than or equal to 1

Comment (paulo-ocean) [reply src/test/integration/typesense.test.ts:165]: @mariacarmina to me, its quite obvious what changed :-)
you changed the ddo metadata (`data/ddo.ts`).. and the query

BEFORE 
the query was looking for '`DEX`' on `metadata.author` field
AND NOW
you're looking for '`orderbook`' on `metadata.name` field... but the name is actually `'ETH/USDT orderbook'`  (the search query is a full search , not a pattern match)

so you need to change the query to
```
const queryParams: TypesenseSearchParams = {
      q: 'ETH/USDT orderbook',
      query_by: 'metadata.name',
      filter_by: 'chainId:=137',
      sort_by: 'version:desc'
    }
```

paulo-ocean: actually the correct query should be:
`q: 'new metadata name'`
as we changed the metadata name earlier

mariacarmina: yes, the second query made the tests passed. It is a bit strange how with previous queries the tests passed alone tho 😅

.github/workflows/ci.yml
paulo-ocean: btw, i already approved the PR on the CLI, so we can merge that one and update the ci.yml here

src/components/Indexer/processor.ts
mariacarmina: Aha ok, it is clearer now, and it won't need deleting of the items while iterating through pricing schemas, because we already have the active pricing schemas returned.

src/components/Indexer/processor.ts
mariacarmina: deleted

.github/workflows/ci.yml
mariacarmina: ok, perfect, thanks

src/components/Indexer/processor.ts
paulo-ocean: yes, exactly. sorry if it wasn't clear before :-) ... but you already did everything needed inside `getPricingStatsForDddo`

jamiehewitt15 (approved):
Looks good to me

mariacarmina:
> lgtm but don't forget to update the ci.yml after merging [oceanprotocol/ocean-cli#87](https://github.com/oceanprotocol/ocean-cli/pull/87)
> 
> ```
> with:
>           repository: 'oceanprotocol/ocean-cli'
>           path: 'ocean-cli'
>           ref: 'fix-tests-for-new-ddo'
> ```
> 
> put back ` ref: 'main'`

yep, done, merged the PR in ocean-cli, now it's updated

PR #775, opened 2024-12-04
fix corner case issue, cannot start indexing on ganache without ADDRESS_FILE

### LINKED ISSUE #773: Indexer - Minor - Unable to index local network (ganache) if we forget about ADDRESS_FILE
If the `ADDRESS_FILE` is not set AND there is no `startBlock` defined on RPCS, under certain circumstances we cannot get the contract deployed block.
It will start at block 0 if we cannot get the last deployed block as well
But ... if we can get that last block it is not forced to start from 0... So, without a start block and a contract deployed block... it just doesn't start (Should start at block 0 in this particular case)
Minor, but still needs a tweak
Screenshot from 2024-12-04 11-58-46
Screenshot from 2024-12-04 11-58-32

paulo-ocean:
Fixes #773 

Changes proposed in this PR:

- Make sure we always assign a default start block for ganache, even if we forget the env setting
- 
-


jamiehewitt15 (approved):
Looks good to me

bogdanfazakas (approved):

PR #786, opened 2024-12-06
Compute Job - Check container stats, regarding cpu & mem usage

### LINKED ISSUE #768: C2D Docker:  Cpu & Memory Limit
Every compute environment has cpu & memory limits.

paulo-ocean:
Fixes #768 

Changes proposed in this PR:

- limit CPU and Memory when creating container (according to environment settings) 
- 
-


paulo-ocean:
> Looks good to me, although there are quite a few lines of commented out code which I sugged we remove if it's not going to be used.

yeah, you're right. Will remove them.. :+1:

PR #789, opened 2024-12-09
add missing check for ddo DB on republish ddos

### LINKED ISSUE #788: Minor - check if DDO DB exists (P2P) before attempt to  republishStoredDDOS()
Sometimes i see this error on the logs:
`Caught "Cannot read properties of undefined (reading 'search')" on republishStoredDDOS()`
P2P  republishStoredDDOS()
Seems DB instance exists but not with a DDO database so a check is missing
Error is from a try/catch block, so its harmful anyway

paulo-ocean:
Fixes #788

Changes proposed in this PR:

- Avoid error,check if DB exists but also if DDO db exists (we might have instance without the 2nd)
- This error is catched but could be avoided anyway
-

### DIFF
--- a/src/components/P2P/index.ts
+++ b/src/components/P2P/index.ts
@@ -779,7 +779,7 @@ export class OceanP2P extends EventEmitter {
   // related: https://github.com/libp2p/go-libp2p-kad-dht/issues/323
   async republishStoredDDOS() {
     try {
-      if (!this.db) {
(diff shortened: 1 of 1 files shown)


jamiehewitt15 (approved):
Looks good to me

mariacarmina (approved):
lgtm

PR #790, opened 2024-12-09
recalculate code hash if needed

### LINKED ISSUE #772: codeHash (and maybe other fields) are missing on the "status" directCommand
First spotted on the CLI  system tests, see
[https://github.com/oceanprotocol/ocean-cli/issues/77](https://github.com/oceanprotocol/ocean-cli/issues/77) for reference
But even using postman the field seems to be absent some times, while present most of the times (hard to reproduce)

paulo-ocean:
Fixes #772 

Changes proposed in this PR:

- code hash might need to be recalculated if config object does not contain key (e.g: after a config reload)
- this absence can cause the system tests to fail (property `codeHash` missing on the `status` command)
-

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -37,12 +37,6 @@ declare global {
   }
 }
 
-export async function recalculateCodeHash(): Promise<string> {
-  const __filename = fileURLToPath(import.meta.url)
-  const __dirname = path.dirname(__filename)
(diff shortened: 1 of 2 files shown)


jamiehewitt15 (approved):
Looks good to me

mariacarmina (approved):
lgtm

PR #793, opened 2024-12-13
add findPeer & dht refactor

alexcos20:
Changes proposed in this PR:

- add findPeer endpoint, that returns multiaddr for a peerId (if exists in DHT)
- refactor DHT options
-


src/components/httpRoutes/getOceanPeers.ts
```diff
@@ -18,6 +18,25 @@ getOceanPeersRoute.get(
...
+  async (req: Request, res: Response): Promise<void> => {
+    if (!req.query.peerId) {
+      res.sendStatus(400)
+      return
+    }
+    if (hasP2PInterface) {
+      const peers = await req.oceanNode
+        .getP2PNode()
+        .findPeerInDht(String(req.query.peerId), parseInt(String(req.query.timeout)))
+      res.json(peers)
```
paulo-ocean, line 34: if we get nothing from `findPeerInDht` , we're sending `null` as response. I think we should check and have a better response like 'could not find peer  ' +  peerId ... or similar

src/components/P2P/index.ts
```diff
@@ -605,6 +590,22 @@ export class OceanP2P extends EventEmitter {
     return finalmultiaddrs
   }
 
+  async findPeerInDht(peerName: string, timeout?: number) {
+    const peer = peerIdFromString(peerName)
```
paulo-ocean, line 594: this should be inside try/catch .. otherwise it crashes the node if we send an invalid peerName (like '123' for instance)

alexcos20: see latest commit

src/components/httpRoutes/getOceanPeers.ts
alexcos20: see latest commit

paulo-ocean (approved):
lgtm

jamiehewitt15 (approved):

alexcos20:
looks like node image is not present anymore:
image

PR #799, opened 2024-12-17
Issue 768 env cpu mem

### LINKED ISSUE #768: C2D Docker:  Cpu & Memory Limit

paulo-ocean:
Fixes #768 

Changes proposed in this PR:

- set cpu and memory options
- 
-

### DIFF
--- a/src/components/c2d/compute_engine_docker.ts
+++ b/src/components/c2d/compute_engine_docker.ts
@@ -525,7 +525,7 @@ export class C2DEngineDocker extends C2DEngine {
         hostConfig.CpuCount = Math.min(confCPUs, existingCPUs)
         // hostConfig.CpuShares = 1 / hostConfig.CpuCount
         hostConfig.CpuPeriod = 100000 // 100 miliseconds is usually the default
-        hostConfig.CpuQuota = (1 / hostConfig.CpuCount) * 100000
+        hostConfig.CpuQuota = (1 / hostConfig.CpuCount) * hostConfig.CpuPeriod
         // if more than 1 CPU, 	Limit the specific CPUs or cores a container can use.
         if (hostConfig.CpuCount > 1) {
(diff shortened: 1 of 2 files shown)


src/components/c2d/compute_engine_docker.ts
```diff
@@ -1116,3 +1131,32 @@ export function checkManifestPlatform(
...
+ * Helper function to build CPU constraints, also useful for testing purposes
+ * @param environment C2D environment
+ * @returns partial HostConfig object
+ */
+export function buildCPUAndMemoryConstraints(
+  environment: ComputeEnvironment
+): HostConfig {
+  const hostConfig: HostConfig = {}
+  // CPU
+  const existingCPUs = os.cpus().length
```
alexcos20, line 1144: why not take the docker system information from https://docs.docker.com/reference/api/engine/version/v1.47/#tag/System/operation/SystemInfo ?

src/components/c2d/compute_engine_docker.ts
```diff
@@ -664,6 +661,24 @@ export class C2DEngineDocker extends C2DEngine {
     }
   }
 
+  // Seems like monitoring container stats is useles... everything related with cpu/mem is at zeros
+  private async checkResources(job: DBComputeJob, container: Dockerode.Container) {
```
alexcos20, line 665: we can remove this

alexcos20 (changes requested):

src/components/c2d/compute_engine_docker.ts
paulo-ocean: we were already using 'os' on other parts (like status), ans its pretty much the same.
anyway, i've  added the docker system info call + the 'os' one as well, as fallback

alexcos20 (approved):
lgtm

PR #800, opened 2024-12-18
set disk usage

### LINKED ISSUE #769: C2D Docker: Storage limits
Every compute environment has storage limits

paulo-ocean:
Fixes #769 

Changes proposed in this PR:

- set storage limits
- 
-

### DIFF
--- a/src/components/c2d/compute_engine_docker.ts
+++ b/src/components/c2d/compute_engine_docker.ts
@@ -517,15 +517,22 @@ export class C2DEngineDocker extends C2DEngine {
         ]
       }
       if (environment != null) {
+        // TODO the CPU and MEM part is addressed in the PR #799 https://github.com/oceanprotocol/ocean-node/pull/799/
+        // REMOVE AFTER MERGE
         // limit container CPU & Memory usage according to env specs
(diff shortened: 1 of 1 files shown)


paulo-ocean:
do we need to check disk space first? we might need some external dependencies for that

alexcos20 (changes requested):
what about /data volume?  it has no limits now, so I can quickly fill it with garbage -> DDOS attack

paulo-ocean:
> what about /data volume? it has no limits now, so I can quickly fill it with garbage -> DDOS attack

Hi @alexcos20  Not sure if i understand exactly what you're saying
Only possible options i see is here:
```
// create the container
      const mountVols: any = { '/data': {} }
      const hostConfig: HostConfig = {
        Mounts: [
          {
            Type: 'volume',
            Source: volume.Name,
            Target: '/data',
            ReadOnly: false
          }
        ]
      }
```
the `Mounts` proerty (of interface type `MountSettings`) has some options
```
 TmpfsOptions?:
            | {
                SizeBytes: number;
                Mode: number;
            }
```
Is this what youŕe referring to? Can you elaborate a bit more on this one? thanks

alexcos20 (approved):
lgtm

PR #802, opened 2025-01-14
new compute envs specs

### LINKED ISSUE #801: C2D: Environments specs
We inherited the old model of compute envs from previous versions, but it's clumsy and hard to configure.
Why:
 - we are exposing multiple envs, but there is no way of imposing limits (imagine a host with 2 cpus, but exposing 10 envs, with different configs.  At most, you will be able to run 2 jobs in the same time (cpu limits)

So, let's switch to the following structure:
```
{
  id: string
  desc: string
  arch: "x86_64",
  totalCpu: number  // total cpu available for jobs
  maxCpu: number  // max cpu for a single job.  Imagine a K8 cluster with two nodes, each node with 10 cpus.  Total=20, but at most you can allocate 10 cpu for a job
  totalRam: number  // total gb of RAM
  maxRam: number  // max allocatable GB RAM for a single job.  
  maxDisk: number  // max GB of disck allocatable for a single job
  maxJobDuration: number  // max seconds for a job
  consumerAddress: string
  storageExpiry: number
  fees: {  
         "chain_X":[
                       feeToken: "0x123",
                       prices: [
                            { type: "cpu", "price": 0.01 },  // price per cpu per minute
                            { type: "memory", "price": 0.02 }, // price per memory per minute
                            { type: "storage", "price": 0.01 }, // price per 1gb of storage per minute
                       ]
          },
         "chain_Y":[
                       "feeToken": "0x123",
                       "prices": [
                            { "type": "cpu", "price": 0.01 },  // price per cpu per minute
                            { "type": "memory", "price": 0.02 }, // price per memory per minute
                            { "type": "storage", "price": 0.01 }, // price per 1gb of storage per minute
                       ]
          },

}
```

Any user, when starting a compute job, can request resources like:
```json
"resources": [
  "type": "cpu", "amount": "1",
 "type": "memory", "amount": "1",
 ]
 ```
 
 and then ocean-node will compute the pricing, by summing price per resource per minute

Configuring envs (by node owner):
Node owner will define compute engines (like docker socker) and ocean-node will create envs (paid + free if needed), the only required data is type of env (docker, k8, etc) and fees.  Everything else we can detect (no of cpus, memory, etc).  That can be overwritten by node owner (example: host has 20 cpus, but node owner can set totalCpu to 18, so compute will never hog the host)

### PR DESCRIPTION
**Pull request by paulo-ocean**
Fixes #801 

Changes proposed in this PR:

- new compute env specs/structures 
- account for incoming legacy stuff (do struct conversion if needed, for instance from operator service)
-

### CODE REVIEW

File: src/components/c2d/index.ts [hunk 1]
```diff
@@ -79,3 +87,65 @@ export function omitDBComputeFieldsFromComputeJob(dbCompute: DBComputeJob): Comp
   ]) as ComputeJob
   return job
 }
+
+export function isLegacyComputeEnvironment(environment: any): boolean {
```
alexcos20, line 91: why do we need this?

src/components/c2d/index.ts
```diff
@@ -79,3 +87,65 @@ export function omitDBComputeFieldsFromComputeJob(dbCompute: DBComputeJob): Comp
...
+      isDefined(environment.ramGB))
+  )
+}
+
+/**
+ * Converts a lecagy structute of ComputeEnvironment to the new structure representation
+ * @param environment old environment structure
+ * @returns new structure representation
+ */
+export async function convertLegacyComputeEnvironment(
```
alexcos20, line 107: We don't need this

src/components/core/utils/feesHandler.ts
```diff
@@ -28,21 +33,45 @@ import ERC20Template from '@oceanprotocol/contracts/artifacts/contracts/template
...
-    // it's a compute provider fee
-    providerFeeAmount = (seconds * parseFloat(String(computeEnv.priceMin))) / 60
+    if (computeEnv.fees) {
+      // get the fess for the asset chain
+      const feesForChain: ComputeEnvFees = computeEnv.fees[chainId]
+      if (feesForChain && feesForChain.prices.length > 0) {
+        const price =
+          // TODO: check this again
+          // try to get the price from one of the available types; 'cpu', 'memory' or 'storage'
+          getEnvironmentPriceSchemaForType(feesForChain.prices, 'cpu') ||
```
alexcos20, line 65: this is sum of resources, not OR

src/test/integration/compute.test.ts
```diff
@@ -265,15 +269,44 @@ describe('Compute', () => {
...
-      }
+
       assert(computeEnvironment.id.startsWith('0x'), 'id should start with 0x')
-      assert(computeEnvironment.cpuNumber > 0, 'cpuNumber missing in computeEnvironments')
-      assert(computeEnvironment.ramGB > 0, 'ramGB missing in computeEnvironments')
-      assert(computeEnvironment.diskGB > 0, 'diskGB missing in computeEnvironments')
-      assert(computeEnvironment.maxJobs > 0, 'maxJobs missing in computeEnvironments')
+      // cehcks both older and newer versions of ComputeEnvironment structure
+      // new structure
+      if (!isLegacyComputeEnvironment(computeEnvironment)) {
```
alexcos20, line 276: No need for previous version of computeEnvs

alexcos20 (changes requested):

src/components/c2d/index.ts
paulo-ocean: i've kept this for backwards compatibility, because of` Operator Service`. I'll remove it, but we need also to make sure we don't use operator service anymore.
We still have the reference/usage on the K8 environments:
[https://github.com/oceanprotocol/ocean-node/blob/2ed69688f7e142a8b443552dff726158b984b158/src/components/c2d/compute_engine_opf_k8.ts#L44](https://github.com/oceanprotocol/ocean-node/blob/2ed69688f7e142a8b443552dff726158b984b158/src/components/c2d/compute_engine_opf_k8.ts#L44)

alexcos20: we will remove that

alexcos20:
You can remove all code related to old k8 c2d if needed. including tests for that

Issue #770, opened 2024-12-02
DHT expose and search


mihaisc:
- include compute env in dht
- create command to search dht

alexcos20:
We need to expose compute capabilities of a node to DHT  (so users can search for them)
Since dht does not support any types of queries, there is no way to perform something like:  "any node which has between 10 and 20 cpus with or without gpu"

Another caveat is the DHT content provider only works with hashes.

So, we need to expose the full capabilities in small steps, by exposing hashes of well known json structures.

Example:

By calculating the hash of the below string, we can serve as a provider of 2 CPUS, 1 Gb of RAM

 ``json
{ cpus:2 , memory: 1000000000}
```

If a host has 12 cpus and 10 gb of RAM, they will announce all combinations:
 - 1 cpu, 1 GB of RAM
 - 2 cpu, 1 GB of RAM
 - 3 cpu, 1 GB of RAM
 .....
  - 12 cpu, 1 GB of RAM
 - 1 cpu, 2 GB of RAM
 - 2 cpu, 2 GB of RAM
 ...

alexcos20:
let's do #801 first

PR #818, opened 2025-01-23
C2D refactor (Free Compute)

alexcos20:
**_BREAKING CHANGES_**

# Node config #
 We will use DOCKER_COMPUTE_ENVIRONMENTS as a definition of Docker engines & compute envs.
 The full definition is withing interface[ `C2DDockerConfig`](https://github.com/oceanprotocol/ocean-node/blob/feature/docker_refactor/src/%40types/C2D/C2D.ts#L85) 
 
 `free` key is optional, and used to define what resources are available for free within a compute env.

There are no more free and non free compute envs. Just one for each docker engine.

When calling startFreeCompute, node will use resource restrictions from `free` config, if `free` key is defined. if not, startFreeCompute will throw an error.
When calling startCompute, node will use resource restrictions and will ignore `free` config.

This will give us some advantages:
 - easier config
 - resource monitoring
 - at job startup, we can count all used resources and decide if we really have the resources to start the job or not

# Resources #
Every compute env has `resources`.  A resource is defined like:
```json
{
  id: ComputeResourceType
  type?: string
  kind?: string
  total: number // total number of specific resource
  min: number // min number of resource needed for a job
  max: number // max number of resource for a job
  inUse?: number // for display purposes
}
```

There are couple of hardcoded:  `cpu`, `ram` and `disk`, and they can be extended by node owner
Disk resource has to be specified , as for now, there is no safe method of getting the available space.

When starting jobs, user can specify the resources needed, or else, if mins are defined, they are attached automatically

 `Free` resources can be defined for a compute env, this is what anyone can use, for free, using startFreeCompute. They are part of compute env resources, and counted.

## Docker with no free compute ##
```bash
export DOCKER_COMPUTE_ENVIRONMENTS="[{\"socketPath\":\"/var/run/docker.sock\",\"storageExpiry\":604800,\"resources\":[{\"id\":\"disk\",\"total\":1000000000}],\"maxJobDuration\":3600,\"fees\":{\"1\":[{\"feeToken\":\"0x123\",\"prices\":[{\"id\":\"cpu\",\"price\":1}]}]}}]"
```

All other settings are detected automatically, like noOfCpus and RAM:

```
2025-01-23T11:18:01.757Z debug: CORE:   CORE:   ComputeGetEnvironmentsCommand Response: 
[
    {
        "id": "0x6075f99f7962fbd0b122dd62405ee1d09ee6855eaf1095b60c9dbbdc5c76eb05-0xcd0ca6d958fc9f3dbb9a64da37ac638f0d8141319d8679eaf59d53f55a0e487f",
        "runningJobs": 0,
        "consumerAddress": "0xf9C5B7eE7708efAc6dC6Bc7d4b0455eBbf22b519",
        "platform": {
            "architecture": "x86_64",
            "os": "Ubuntu 22.04.3 LTS"
        },
        "fees": {
            "1": [
                [
                    {
                        "feeToken": "0x123",
                        "prices": [
                            {
                                "id": "cpu",
                                "price": 1
                            }
                        ]
                    }
                ]
            ]
        },
        "storageExpiry": 604800,
        "maxJobDuration": 3600,
        "resources": [
            {
                "id": "cpu",
                "total": 16,
                "max": 16,
                "min": 1,
                "inUse": 0
            },
            {
                "id": "ram",
                "total": 33617678336,
                "max": 33617678336,
                "min": 1000000000,
                "inUse": 0
            },
            {
                "id": "disk",
                "total": 1000000000,
                "max": 1000000000,
                "min": 0,
                "inUse": 0
            }
        ],
        "runningfreeJobs": 0
    }
]
```

## Docker with free compute ##
```bash
export  DOCKER_COMPUTE_ENVIRONMENTS="[{\"socketPath\":\"/var/run/docker.sock\",\"resources\":[{\"id\":\"disk\",\"total\":1000000000}],\"storageExpiry\":604800,\"maxJobDuration\":3600,\"fees\":{\"1\":[{\"feeToken\":\"0x123\",\"prices\":[{\"id\":\"cpu\",\"price\":1}]}]},\"free\":{\"maxJobDuration\":60,\"maxJobs\":3,\"resources\":[{\"id\":\"cpu\",\"max\":1},{\"id\":\"ram\",\"max\":1000000000},{\"id\":\"disk\",\"max\":1000000000}]}}]"
```

All other settings are detected automatically, like noOfCpus and RAM:

```
2025-01-23T11:19:33.193Z debug: CORE:   CORE:   ComputeGetEnvironmentsCommand Response: 
[
    {
        "id": "0x7d187e4c751367be694497ead35e2937ece3c7f3b325dcb4f7571e5972d092bd-0x3a5647ecc1b6a3b4eaaeaf818037ca5a7aea23b2302f9c09c2ad716d3b0f94c7",
        "runningJobs": 0,
        "consumerAddress": "0xf9C5B7eE7708efAc6dC6Bc7d4b0455eBbf22b519",
        "platform": {
            "architecture": "x86_64",
            "os": "Ubuntu 22.04.3 LTS"
        },
        "fees": {
            "1": [
                [
                    {
                        "feeToken": "0x123",
                        "prices": [
                            {
                                "id": "cpu",
                                "price": 1
                            }
                        ]
                    }
                ]
            ]
        },
        "storageExpiry": 604800,
        "maxJobDuration": 3600,
        "resources": [
            {
                "id": "cpu",
                "total": 16,
                "max": 16,
                "min": 1,
                "inUse": 0
            },
            {
                "id": "ram",
                "total": 33617678336,
                "max": 33617678336,
                "min": 1000000000,
                "inUse": 0
            },
            {
                "id": "disk",
                "total": 1000000000,
                "max": 1000000000,
                "min": 0,
                "inUse": 0
            }
        ],
        "free": {
            "maxJobDuration": 60,
            "maxJobs": 3,
            "resources": [
                {
                    "id": "cpu",
                    "max": 1,
                    "inUse": 0
                },
                {
                    "id": "ram",
                    "max": 1000000000,
                    "inUse": 0
                },
                {
                    "id": "disk",
                    "max": 1000000000,
                    "inUse": 0
                }
            ]
        },
        "runningfreeJobs": 0
    }
]
```

# ComputeEnv interface & getComputeEnvironments command #

 see new structures above

# Commands #

- `startFreeCompute` requires environmentId now
- for `startFreeCompute` and `startCompute` a new optional parameter `resources` is available.   This will allow user to request what resources to use, instead of using all.   Imagine the following scenario:
      - host has 16 cpus and 4 gpus exposed in a compute env
      - in the current setup, if you want to run a job, you will pay for all   (16 x price_per_cpu_per_min + 4 x price_per_gpu)*minutes
      - by using something like `resources: [{ "type":"cpu",amount: 4}, { "type":"gpu",amount: 1}] ` you will only pay for 4 cpus and gpus , because that is what you are going to get
- 
-

### DIFF
--- a/src/@types/C2D/C2D.ts
+++ b/src/@types/C2D/C2D.ts
@@ -97,6 +97,10 @@ export interface ComputeEnvByChain {
   [chainId: number]: ComputeEnvironment[]
 }
 
+export interface ComputeResourceRequest {
+  type: string
+  amount: number
+}
 export type ComputeResultType =
   | 'algorithmLog'
   | 'output'
--- a/src/@types/C2D/C2D.ts
+++ b/src/@types/C2D/C2D.ts
@@ -34,6 +34,7 @@ export interface ComputeResource {
   total: number // total number of specific resource
   min: number // min number of resource needed for a job
   max: number // max number of resource for a job
+  inUse?: number // for display purposes
 }
 export interface ComputeResourceRequest {
   id: string
@@ -63,7 +64,8 @@ export interface ComputeEnvironmentFreeOptions {
 export interface ComputeEnvironmentBaseConfig {
   description?: string // v1
   storageExpiry?: number // amount of seconds for storage
-  maxJobDuration?: number // v1 max seconds for a paid job
+  minJobDuration?: number // min billable seconds for a paid job
+  maxJobDuration?: number // max duration in seconds for a paid job
   maxJobs?: number // maximum number of simultaneous paid jobs
   fees: ComputeEnvFeesStructure
   resources?: ComputeResource[]
@@ -73,7 +75,8 @@ export interface ComputeEnvironmentBaseConfig {
 
 export interface ComputeEnvironment extends ComputeEnvironmentBaseConfig {
   id: string // v1
-  currentJobs: number
+  runningJobs: number
+  runningfreeJobs?: number
   consumerAddress: string // v1
 }
 
@@ -89,7 +92,7 @@ export interface C2DDockerConfig {
   maxJobDuration?: number
   maxJobs?: number
   fees: ComputeEnvFeesStructure
-  maxDisk: number
+  resources?: ComputeResource[] // optional, owner can overwrite
   free?: ComputeEnvironmentFreeOptions
 }
 
--- a/src/@types/C2D/C2D.ts
+++ b/src/@types/C2D/C2D.ts
@@ -20,18 +20,26 @@ export interface C2DClusterInfo {
   tempFolder?: string
 }
 
-// export type ComputeResourceType = 'cpu' | 'memory' | 'storage'
+export type ComputeResourceType = 'cpu' | 'ram' | 'disk' | any
 
 export interface ComputeResourcesPricingInfo {
-  id: string
-  price: number
+  id: ComputeResourceType
+  price: number // price per unit
 }
 
-export interface ComputeResources {
-  id: string
+export interface ComputeResource {
+  id: ComputeResourceType
   type?: string
   kind?: string
+  total: number // total number of specific resource
+  min: number // min number of resource needed for a job
+  max: number // max number of resource for a job
 }
+export interface ComputeResourceRequest {
+  id: string
+  amount: number
+}
+
 export interface ComputeEnvFees {
   feeToken: string
   prices: ComputeResourcesPricingInfo[]
@@ -47,33 +55,20 @@ export interface RunningPlatform {
 
 export interface ComputeEnvironmentFreeOptions {
   // only if a compute env exposes free jobs
-  maxCpu?: number // max cpu for a single job.
-  maxRam?: number // max allocatable RAM for a single job in bytes.
-  maxDisk?: number // max disk space in bytes allocatable for a single job
   storageExpiry?: number
   maxJobDuration?: number
-  maxJobs: number // maximum number of free jobs in the same time
+  maxJobs?: number // maximum number of simultaneous free jobs
+  resources?: ComputeResource[]
 }
 export interface ComputeEnvironmentBaseConfig {
-  // cpuNumber: number
-  // ramGB: number
-  // diskGB: number
   description?: string // v1
-  // maxJobs: number
-  storageExpiry: number // v1
-  maxJobDuration: number // v1 max seconds for a job
-  // chainId?: number
-  // feeToken: string
-  // priceMin: number
-  totalCpu?: number // total cpu available for jobs
-  totalRam?: number // total bytes of RAM
-  maxCpu?: number // max cpu for a single job.  Imagine a K8 cluster with two nodes, each node with 10 cpus.  Total=20, but at most you can allocate 10 cpu for a job
-  maxRam?: number // max allocatable RAM for a single job in bytes.
-  maxDisk?: number // max disk space in bytes allocatable for a single job
-  free?: ComputeEnvironmentFreeOptions
+  storageExpiry?: number // amount of seconds for storage
+  maxJobDuration?: number // v1 max seconds for a paid job
+  maxJobs?: number // maximum number of simultaneous paid jobs
   fees: ComputeEnvFeesStructure
-  resources?: ComputeResources[]
-  platform?: RunningPlatform[]
+  resources?: ComputeResource[]
+  free?: ComputeEnvironmentFreeOptions
+  platform: RunningPlatform
 }
 
 export interface ComputeEnvironment extends ComputeEnvironmentBaseConfig {
@@ -90,17 +85,18 @@ export interface C2DDockerConfig {
   caPath: string
   certPath: string
   keyPath: string
-  environments: ComputeEnvironment[]
+  storageExpiry?: number
+  maxJobDuration?: number
+  maxJobs?: number
+  fees: ComputeEnvFeesStructure
+  maxDisk: number
+  free?: ComputeEnvironmentFreeOptions
 }
 
 export interface ComputeEnvByChain {
   [chainId: number]: ComputeEnvironment[]
 }
 
-export interface ComputeResourceRequest {
-  type: string
-  amount: number
-}
 export type ComputeResultType =
   | 'algorithmLog'
   | 'output'
@@ -178,6 +174,8 @@ export interface DBComputeJob extends ComputeJob {
   isRunning: boolean
   isStarted: boolean
   containerImage: string
+  resources: ComputeResourceRequest[]
+  isFree: boolean
 }
 
 // make sure we keep them both in sync
--- a/src/@types/C2D/C2D.ts
+++ b/src/@types/C2D/C2D.ts
@@ -45,7 +45,7 @@ export interface ComputeEnvFees {
   prices: ComputeResourcesPricingInfo[]
 }
 export interface ComputeEnvFeesStructure {
-  [chainId: string]: ComputeEnvFees
+  [chainId: string]: ComputeEnvFees[]
 }
 
 export interface RunningPlatform {
@@ -91,10 +91,6 @@ export interface C2DDockerConfig {
   fees: ComputeEnvFeesStructure
   maxDisk: number
   free?: ComputeEnvironmentFreeOptions
-}
-
-export interface ComputeEnvByChain {
-  [chainId: number]: ComputeEnvironment[]
 }
 
 export type ComputeResultType =
--- a/src/@types/C2D/C2D.ts
+++ b/src/@types/C2D/C2D.ts
@@ -20,13 +20,18 @@ export interface C2DClusterInfo {
   tempFolder?: string
 }
 
-export type ComputeResourceType = 'cpu' | 'memory' | 'storage'
+// export type ComputeResourceType = 'cpu' | 'memory' | 'storage'
 
 export interface ComputeResourcesPricingInfo {
-  type: ComputeResourceType
+  id: string
   price: number
 }
 
+export interface ComputeResources {
+  id: string
+  type?: string
+  kind?: string
+}
 export interface ComputeEnvFees {
   feeToken: string
   prices: ComputeResourcesPricingInfo[]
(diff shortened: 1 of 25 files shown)


jamiehewitt15:
I ran this branch with:

```
 export  DOCKER_COMPUTE_ENVIRONMENTS="[{\"socketPath\":\"/var/run/docker.sock\",\"resources\":[{\"id\":\"disk\",\"total\":1000000000}],\"storageExpiry\":604800,\"maxJobDuration\":3600,\"fees\":{\"1\":[{\"feeToken\":\"0x123\",\"prices\":[{\"id\":\"cpu\",\"price\":1}]}]},\"free\":{\"maxJobDuration\":60,\"maxJobs\":3,\"resources\":[{\"id\":\"cpu\",\"max\":1},{\"id\":\"ram\",\"max\":1000000000},{\"id\":\"disk\",\"max\":1000000000}]}}]"
```

I got the error below when calling `GET /api/services/computeEnvironments`. Is there something else that is needed in the setup?

```
2025-01-29T10:21:15.395Z debug: HTTP:	HTTP:	GET computeEnvironments request received with query: {}
2025-01-29T10:21:15.396Z info: CORE:	Checking received command data for Command "getComputeEnvironments": {
    "command": "getComputeEnvironments",
    "chainId": null,
    "node": null
}
2025-01-29T10:21:15.467Z debug: CORE:	CORE:	ComputeGetEnvironmentsCommand Response: [
  {
    "id": "0x7d187e4c751367be694497ead35e2937ece3c7f3b325dcb4f7571e5972d092bd-0x071ead74e903edeb2ad40d196f03db09f70811ede01f3e111fd5106f52b388ee",
    "runningJobs": 0,
    "consumerAddress": "0xe2DD09d719Da89e5a3D0F2549c7E24566e947260",
    "platform": {
      "architecture": "x86_64",
      "os": "Ubuntu 22.04.4 LTS"
    },
    "fees": null,
    "storageExpiry": 604800,
    "maxJobDuration": 3600,
    "resources": [
      {
        "id": "cpu",
        "total": 2,
        "max": 2,
        "min": 1,
        "inUse": 0
      },
      {
        "id": "ram",
        "total": 4103393280,
        "max": 4103393280,
        "min": 1000000000,
        "inUse": 0
      },
      {
        "id": "disk",
        "total": 1000000000,
        "max": 1000000000,
        "min": 0,
        "inUse": 0
      }
    ],
    "free": {
      "maxJobDuration": 60,
      "maxJobs": 3,
      "resources": [
        {
          "id": "cpu",
          "max": 1,
          "inUse": 0
        },
        {
          "id": "ram",
          "max": 1000000000,
          "inUse": 0
        },
        {
          "id": "disk",
          "max": 1000000000,
          "inUse": 0
        }
      ]
    },
    "runningfreeJobs": 0
  }
]
2025-01-29T10:21:15.469Z error: Error: TypeError: Cannot read properties of undefined (reading 'length')
```

src/@types/commands.ts
```diff
@@ -166,14 +171,18 @@ export interface ComputeStartCommand extends Command {
...
   datasets?: ComputeAsset[]
   output?: ComputeOutput
+  resources?: ComputeResourceRequest[]
+  chainId?: number // network used by payment
 }
 export interface FreeComputeStartCommand extends Command {
   consumerAddress: string
   signature: string
   nonce: string
+  environment: string
```
jamiehewitt15, line 181: Could we make this optional? In case there is only one free compute environment, then we don't need to send the environment ID. We just use the only one available. I'm not sure exactly why someone would have multiple free environments? Seems more likely that they would have multiple paid environments and then just a single free one.

src/@types/commands.ts
```diff
@@ -166,14 +171,18 @@ export interface ComputeStartCommand extends Command {
...
   algorithm: ComputeAlgorithm
   datasets?: ComputeAsset[]
   output?: ComputeOutput
+  resources?: ComputeResourceRequest[]
```
jamiehewitt15, line 185: I don't see why someone would use this on free compute? It's free, so everyone will just always want the maximum possible resources.

jamiehewitt15:
Despite the issue I'm getting when calling the `GET /api/services/computeEnvironments` it works fine for starting a compute job.

paulo-ocean:
have the same issue, problem is on `areEmpty` function, this line:
`if (computeEnvs[supportedNetwork].length === 0) {`

jamiehewitt15:
When I set the `maxJobDuration` to 600 on the free compute I get: `500 Invalid C2D Environment` when trying to start a compute job via direct command.

src/@types/commands.ts
paulo-ocean: i would agree, now its mandatory to know the id in advance, while before it was not... and usually there is only 1 free env

paulo-ocean:
ok, so for now, it only works with  a directCommand... but we need to know the env id in advance..

src/@types/C2D/C2D.ts
```diff
@@ -20,56 +20,64 @@ export interface C2DClusterInfo {
...
 export interface ComputeResourcesPricingInfo {
-  type: ComputeResourceType
-  price: number
+  id: ComputeResourceType
+  price: number // price per unit
+}
+
+export interface ComputeResource {
+  id: ComputeResourceType
+  type?: string
```
paulo-ocean, line 32: what's the `type` and `kind`? it can be a bit confusing with the `ComputeResourceType` type that is now the `id` ?

alexcos20:
How do you even now if a node has free compute envs?  By calling getComputeEnvironments
How do you know if the architecture is a match(x64, arm)? By calling getComputeEnvironments
How do you know that free job duration is usable for your algo? By calling getComputeEnvironments
How do you know if free resources are enough for your algo? By calling getComputeEnvironments

So, in all cases, you have the compute env id, so passing it it's not a big deal.
What if I have two k8 clusters (one x64, the other one arm)? -> multiple C2D environments

src/@types/C2D/C2D.ts
alexcos20: kind can be  "Discrete" or "Named"
type will describe the resource

ie: id='gpu' , type='RX580'  or    id='cpu', type='AMD Ryzen 7 5800X 8-Core Processor'

This will make our life easier when dealing with ASICS and FGPA devices

alexcos20:
> have the same issue `2025-01-29T10:21:15.469Z error: Error: TypeError: Cannot read properties of undefined (reading 'length')` , seems the problem is on `areEmpty` function, this line: `if (computeEnvs[supportedNetwork].length === 0) {` BUT i don't see any job starting, are you sure? Do we need changes on CLI + SDK or something? cause with the CLI we don't pass the last error and if i fix that i got others.. so not working for me at all

@paulo-ocean @jamiehewitt15 

SDK needs to be updated, there is no  `computeEnvs[supportedNetwork]`  anymore,  just look above.
The return is an array of envs, each env supporting one or more chains for payments.

There is no sane logic, if you have an env supporting payment on chains 1 and 2, to see it listed as :
```
{
'1':  computeEnv (with fees on 1 and 2)
'2':  theSameComputeEnv(with fees on 1 and 2)
}
```

alexcos20:
Actually, http endpoint will always return an array.  
 - can be empty `[]` -> no envs found
 - can contain elements
 
 We need to be consistent, because calling `getComputeEnvironments` using directCommand always returns an array, so why have  HTTP 404 if the array is empty?
 
 Fixed

alexcos20:
> > When I increase the maxJobDuration to 600 on the free compute I get: 500 Invalid C2D Environment when trying to start a compute job via direct command. It works when I have it set at 60.
 
 Can you paste your config and node output pls? Most likely you need to config storageExpiry as well..
 Also, what is the result of getComputeEnvironments?

paulo-ocean:
> > > When I increase the maxJobDuration to 600 on the free compute I get: 500 Invalid C2D Environment when trying to start a compute job via direct command. It works when I have it set at 60.
> 
> Can you paste your config and node output pls? Most likely you need to config storageExpiry as well (see https://github.com/oceanprotocol/ocean-node/blob/feature/docker_refactor/src/utils/config.ts#L411)
> 
> Also, what is the result of getComputeEnvironments?

yeah, we mentioned that possibility on the meeting

jamiehewitt15:
getComputeEnvironments is working now. The issue with it failing to start a compute job when I increase the `maxJobDuration` also seems to be fixed as well. I'll continue to do some more testing.

jamiehewitt15:
> How do you even now if a node has free compute envs? 

From the perspective of the vscode extension, we will be predominantly using a list of default nodes which we know have free compute environments. So the `getComputeEnvironments` doesn't add any value in this respect. It's just an extra HTTP request to tell us something we already know. For other unknown nodes it would be needed, but that supports the idea that it could be optional. 

> How do you know if the architecture is a match(x64, arm)? 
> How do you know that free job duration is usable for your algo? 
> How do you know if free resources are enough for your algo? 

It's free, so the answer to all of these questions for most people will not be to read the response of `getComputeEnvironments`, they will just run the algorithm and find out if it works or not. Again, thinking from the perspective of the vscode extension - they only have to click a single "Run algorithm" button to find out if it works or not. They will most likely only consider the response of `getComputeEnvironments` if the algorithm fails on the first try.

alexcos20:
> > How do you even now if a node has free compute envs?
> 
> From the perspective of the vscode extension, we will be predominantly using a list of default nodes which we know have free compute environments. So the `getComputeEnvironments` doesn't add any value in this respect. It's just an extra HTTP request to tell us something we already know. For other unknown nodes it would be needed, but that supports the idea that it could be optional.

That is just for the beginning. The main idea is not to use hardcoded nodes, but search DHT for providers that do match your requirements (as paid/free, no of cpus, certain gpu, etc). See https://github.com/oceanprotocol/ocean-node/issues/770

> > How do you know if the architecture is a match(x64, arm)?
> > How do you know that free job duration is usable for your algo?
> > How do you know if free resources are enough for your algo?
> 
> It's free, so the answer to all of these questions for most people will not be to read the response of `getComputeEnvironments`, they will just run the algorithm and find out if it works or not. Again, thinking from the perspective of the vscode extension, they only have to click a single "Run algorithm" button to find out if it works or not. They will only consider the response of `getComputeEnvironments` if the algorithm fails on the first try.
> 

vscode extension should check if a certain node is able to compute your job, not blindly throw anything to the node

> I don't really see an issue with the environment ID being optional and then using a default free environment?

I will add it, but, again, breaks a lot of logic that we should handle before calling startCompute or freeStartCompute

src/@types/commands.ts
alexcos20: it's optional.  While on community, chances of somebody using it are low (unless you want to test your algo and see how it behaves with low resources), on enterprise can be useful

src/utils/config.ts
```diff
@@ -468,66 +389,65 @@ function getDockerFreeComputeOptions(
...
   if (
     existsEnvironmentVariable(
       ENVIRONMENT_VARIABLES.DOCKER_COMPUTE_ENVIRONMENTS,
       isStartup
     )
   ) {
     try {
-      const options: ComputeEnvironmentBaseConfig[] = JSON.parse(
+      const configs: C2DDockerConfig[] = JSON.parse(
         process.env.DOCKER_COMPUTE_ENVIRONMENTS
```
paulo-ocean, line 402: we're missing the `DOCKER_COMPUTE_ENVIRONMENTS` var reference on `env.md` and `.env.example`

jamiehewitt15:
Right now, the signature isn't being validated. The compute jobs run perfectly fine with this request: 

```
    const requestBody = {
      command: 'freeStartCompute',
      consumerAddress: consumerAddress,
      environment: environmentId,
      nonce: nonce,
      signature: '0x123',
      datasets: dataset ? [dataset] : [],
      algorithm: {
        meta: {
          rawcode: algorithmContent,
          container: containerConfig
        }
      }
    }
```

Do we plan to change that for the release?

paulo-ocean (approved):
i think we can merge it.. created a couple of new issues for the things mentioned on the comments

PR #823, opened 2025-01-27
Issue 814 credentials types

### LINKED ISSUE #814: Credentials types
With OPE, we have new credentials types(SSI).  But we should allow more users to create their own types (LDAP, NTLM, Active Directory, Oauth, etc).
Of course, node cannot handle those types, so we need to change the logic.

Flow:
  - if POLICY_SERVER_URL exists, then ocean-node will **_NOT_** perform any checks. It will just use the existing code and let PolicyServer decide.
  - if POLICY_SERVER_URL is not defined then ocean-node will perform checks only for 'known' types. The remaining rules **_WILL BE IGNORED_**!

Known types:
 - `address`  - web3 address
 - `accessList`  - accessList contract (see #808)

paulo-ocean:
Fixes #814 

Changes proposed in this PR:

- 
- 
-


src/utils/credentials.ts
```diff
@@ -45,3 +50,27 @@ export function checkCredentials(credentials: Credentials, consumerAddress: stri
   }
   return true
 }
+
+export function areKnownCredentialTypes(credentials: Credentials): boolean {
+  let known = true
```
mariacarmina, line 55: I think we can return simply true or false, no need to store in this variable

src/utils/constants.ts
```diff
@@ -328,6 +328,11 @@ export const ENVIRONMENT_VARIABLES: Record<any, EnvVariable> = {
     name: 'IS_BOOTSTRAP',
     value: process.env.IS_BOOTSTRAP,
     required: false
+  },
+  POLICY_SERVER_URL: {
```
mariacarmina, line 332: Can we have this in ocean-quickstart as well? `POLICY_SERVER_URL`

mariacarmina (commented):
some small suggestions from my side, otherwise looks good :)

bogdanfazakas (approved):

PR #826, opened 2025-01-29
add AUTHORIZED_PUBLISHERS_*** env variables

### LINKED ISSUE #815: DDO Index:  Isolated node
In order to implement the concept of isolated node, we need some restrictions on what assets are indexed.  We already have validators, but we need to add a list of publishers as well.

To do:
- add AUTHORIZED_PUBLISHERS env variable
- add AUTHORIZED_PUBLISHERS _LIST env variable

When indexing a new asset, if any of those envs is defined, check the publisher address and index only if it's a match (either on AUTHORIZED_PUBLISHERS , or in AUTHORIZED_PUBLISHERS_LIST with a matching chainId)

- AUTHORIZED_PUBLISHERS : List of address of accounts Example: "[\"0x967da4048cD07aB37855c090aAF366e4ce1b9F48\",\"0x388C818CA8B9251b393131C08a736A67ccB19297\"]"

- AUTHORIZED_PUBLISHERS_LIST looks like
```json
{
 "1": ["0x123","0x124"],
 "2": ["0x345"]
}
```
(Which means:  If an asset is published on chainId 1, then see if the publisher is on the accessList contract 0x123 or 0x124

paulo-ocean:
Fixes #815 

Changes proposed in this PR:

- add `AUTHORIZED_PUBLISHERS_LIST` and `AUTHORIZED_PUBLISHERS`  env variables
- index accordingly
-


jamiehewitt15 (approved):
Looks good to me

bogdanfazakas (approved):
LGTM

PR #828, opened 2025-01-30
Feature/escrow_payments

alexcos20:
Changes proposed in this PR:

- remove OPF_k8
- add escrow payments
- update interfaces for paid Compute Jobs (BREAKING CHANGES)

How to use:
 - start barge with 
 ```bash
 export CONTRACTS_VERSION=escrow
./start_ocean.sh --no-aquarius --no-provider --no-dashboard --with-typesense
 ```

Nice TODOs  (can be done later):
- more tests for lock/claim
- docs


src/components/core/utils/escrow.ts
```diff
@@ -0,0 +1,244 @@
...
+    token: string,
+    payer: string,
+    payee: string
+  ): Promise<EscrowAuthorization[]> {
+    const { rpc, network, chainId, fallbackRPCs } = this.networks[chain]
+    const blockchain = new Blockchain(rpc, network, chainId, fallbackRPCs)
+    const signer = blockchain.getSigner()
+    const contract = await this.getContract(chainId, signer)
+    try {
+      return await contract.getAuthorizations(token, payer, payee)
```
mariacarmina, line 98: Those functions, at least most of them are implemented [here](https://github.com/oceanprotocol/ocean.js/pull/1910), when a new ocean.js version will be released, I would say that contract object creation and calling contract function to be replaced with the methods in ocean.js. In ocean.js I can also add the additional checks that you managed to write within ocean-node.

mariacarmina (commented):
I left some comments, still need to review feesHandler & and test the end2end flow, because it contains many breaking changes.

src/components/c2d/compute_engine_base.ts
```diff
@@ -288,4 +293,70 @@ export abstract class C2DEngine {
...
+  ): ComputeResourcesPricingInfo[] {
+    console.log('getEnvPricesForToken')
+    console.log(env)
+    if (!env.fees || !(chainId in env.fees) || !env.fees[chainId]) {
+      return null
+    }
+    console.log(env.fees)
+    for (const fee of env.fees[chainId]) {
+      console.log(fee)
+      console.log(fee.feeToken)
```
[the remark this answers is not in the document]
alexcos20: for debug only, will clean up later

src/components/core/utils/escrow.ts
alexcos20: what's the relation?  ocean-node does not use ocean.js

jamiehewitt15:
We will need some documentation on how the escrow payments work, is that going to come in a separate PR?

src/components/c2d/compute_engine_docker.ts
```diff
@@ -762,8 +767,9 @@ export class C2DEngineDocker extends C2DEngine {
         // is running, we need to stop it..
         console.log('running, need to stop it?')
         const timeNow = Date.now() / 1000
-        console.log('timeNow: ' + timeNow + ' , Expiry: ' + job.expireTimestamp)
-        if (timeNow > job.expireTimestamp || job.stopRequested) {
+        const expiry = parseFloat(job.algoStartTimestamp) + job.maxJobDuration
+        console.log('timeNow: ' + timeNow + ' , Expiry: ' + expiry)
```
jamiehewitt15, line 771: Why are you using `console.log` rather than our CustomNodeLogger?

src/components/core/compute/initialize.ts
```diff
@@ -65,16 +65,117 @@ export class ComputeInitializeHandler extends Handler {
...
+        if (!env) {
+          return {
+            stream: null,
+            status: {
+              httpStatus: 500,
+              error: 'Invalid C2D Environment'
+            }
+          }
+        }
+
```
jamiehewitt15, line 108: It would be good if each of these 500 responses had a different error message so we know exactly where it is coming from and which check failed.

src/components/core/compute/initialize.ts
```diff
@@ -267,66 +350,27 @@ export class ComputeInitializeHandler extends Handler {
...
-              result.providerFee = await createProviderFee(
-                ddo,
-                service,
-                bestValidUntil,
-                env,
-                task.compute.validUntil
-              )
-              foundValidCompute = { txId: null, chainId: ddo.chainId }
+              // TO DO:  Edge case when this asset is served by a remote provider.
+              // We should connect to that provider and get the fee
```
jamiehewitt15, line 365: Looks like this was left here from before your changes, is it still the plan? If so we need an issue for it

src/components/database/sqliteCompute.ts
```diff
@@ -127,12 +127,11 @@ export class SQLiteCompute implements ComputeDatabaseProvider {
       statusText, 
       inputDID, 
       algoDID, 
-      agreementId, 
       expireTimestamp, 
       environment, 
       body
       )
-      VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?);
+      VALUES (?, ?, ?, ?,  ?, ?, ?, ?, ?, ?, ?);
```
jamiehewitt15, line 134:
```suggestion
      VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?);
```

src/OceanNode.ts
```diff
@@ -29,33 +30,45 @@ export class OceanNode {
...
   private c2dEngines: C2DEngines
+  // escrow
+  public escrow: Escrow
   // requester
   private remoteCaller: string | string[]
   private requestMap: Map<string, RequestLimiter>
+
   // eslint-disable-next-line no-useless-constructor
   private constructor(
+    private config: OceanNodeConfig,
```
paulo-ocean, line 41: why do we need to add the config? we already did the same change a couple of times in the past.. but the config is something that change over time (very frequent on the tests).. The initial idea was to create a node Instance based on the  config 
Afaik, its only usage here is to "decide" if we should also add both C2D engines and Escrow components... so why not add those on the constructor (like we do for other optional components, like; P2P, DB, Indexer, etc ?)...

docs/env.md
```diff
@@ -28,6 +28,10 @@ Environmental variables are also tracked in `ENVIRONMENT_VARIABLES` within `src/
 - `MAX_CHECKSUM_LENGTH`: Define the maximum length for a file if checksum is required (Mb). Example: `10`
 - `IS_BOOTSTRAP`: Is this node to be used as bootstrap node or not. Default is `false`.
 
+## Payments
+
+- `ESCROW_CLAIM_TIMEOUT`: Amount of time reserved to claim a escrow payment, in seconds. Defaults to `600`. Example: `600`
```
paulo-ocean, line 33: i think we should also add the variable reference to `constants/ENVIRONMENT_VARIABLES`, `env.example` and `scripts/ocean-node-quickstart.sh`

src/@types/C2D/C2D.ts
```diff
@@ -120,8 +124,7 @@ export interface ComputeJob {
   results: ComputeResult[]
   inputDID?: string[]
   algoDID?: string
-  agreementId?: string
```
paulo-ocean, line 123: What happened to the `agreementId` ? if its not to be part of a compute job anymore, then there is a lot of code yet to be refactored, as the `agreementId` is used in a lot of current logic

src/components/core/utils/escrow.ts
mariacarmina: @alexcos20 it is helpful to use the lib when we interact with smart contracts, at least that's a suggestion, we also have ethers in ocean-node to interact with smart contracts. I've just seen that some similar functions from ocean-node for escrow payments (and not only) that can be imported from ocean.js

paulo-ocean: well, i actually disagree on this.. IMO is better to NOT use the SDK on the nodes, AS IS... otherwise we might have to keep SDK updated (and release a new version) every time we do a big change on the Nodes... Every contract is different and has different functions, we don't need the SDK to interact with a smart contract.. 

we have ethers for that.. If we're gonna use the SDK on the Nodes... we might have a "trilema", as we need to keep always 3 different pieces of software in sync. the SDK, the CLI (which we also use for the system tests) and the Nodes

src/@types/C2D/C2D.ts
alexcos20: `agreementId`  is the tx in which ocean-node creates the lock

src/components/core/utils/escrow.ts
paulo-ocean: Its ok that we have different views :-)
I personally don't see any need for SDK on the nodes, and i think it will bring more way more issues than solutions.
Afaik, the SDK is mainly a library to abstract indexer, provider, aquarius, etc.. deal with configurations, URLS, etc..
This is all part of the nodes monolithic infrastructure... don´t see how its helpful here

Regrading the smart contracts in particular, I don't think we need an SDK to encapsulate  every smart contract functionality into a typescript class, just to eventually call a smart contract function. And i definitely don't see that as a code duplication... its just an enclosing function to call the smart contract level function.. If we use the SDK for this particular case its exactly the same thing... except that we have an extra Instance in memory, and we call the encapsulating function that actually does the smart contract call... And if we need a bit more logic on the node side, what do we do then? We update the SDK?

alexcos20: Why we would integrate ocean-js in ocean-node?  For 3 (three) functions in a smartcontract?    Which, for now, don't even overlap? (ie: they do not exist in ocean.js)

mariacarmina (commented):
At initializeCompute https://github.com/oceanprotocol/ocean-node/pull/828/files#diff-71521de5df77d93ae85ff75f78504f70beec812c2e776bf25e990a0d5d4c37b8R284 we do not create any task object specific for compute initialize, I'd suggest to have some checks for the following keys: 'payment', 'consumerAddress', 'environment' -> then we can update also the API.md for compute routes to be easier to follow the parameters.

giurgiur99 (approved):

.github/workflows/ci.yml
```diff
@@ -125,23 +125,25 @@ jobs:
               DOCKERHUB_PASSWORD: ${{ secrets.DOCKERHUB_PASSWORD }}
       - name: Run Barge
         working-directory: ${{ github.workspace }}/barge
+        env:
```
mariacarmina, line 128: Can we can update barge with default version to `escrow`? Or after we have this escrow integration in place and see if works on all impacted repos?

mariacarmina (approved):
One question from my side, otherwise it's good to start with this baseline and not keeping in review for too much time.

.github/workflows/ci.yml
alexcos20: I will take care of this

PR #829, opened 2025-02-03
add ALLOWED_VALIDATORS_LIST

### LINKED ISSUE #813: Config env ALLOWED_VALIDATORS_LIST
We already have ALLOWED_VALIDATORS env, which decides which addresses are allowed validators to verify asset signatures before indexing.

- add `ALLOWED_VALIDATORS_LIST` env variable, which looks like
```json
{
 "1": ["0x123","0x124"],
 "2": ["0x345"]
}
```

which means:
  
Ocean-node will check both `ALLOWED_VALIDATORS` and `ALLOWED_VALIDATORS_LIST` (if they are defined) and if an address is a match on any , then it's allowed

paulo-ocean:
Fixes #813

Changes proposed in this PR:

- Add ALLOWED_VALIDATORS_LIST, using access lists (who can validate assets before indexing)
- 
-


src/components/Indexer/utils.ts
```diff
@@ -157,6 +160,7 @@ export const processChunkLogs = async (
           event.type === EVENTS.METADATA_UPDATED ||
           event.type === EVENTS.METADATA_STATE
         ) {
+          // ref: https://github.com/oceanprotocol/ocean-node/issues/257
```
mariacarmina, line 163: this seems as an acient issue, seems to be merged, is there any follow-up issue which is more actual?

src/components/Indexer/utils.ts
```diff
@@ -185,8 +191,47 @@ export const processChunkLogs = async (
...
+                        `Metadata validator: ${metaproofValidator} is NOT part of the access list group: ${accessListAddress}.`
+                      )
+                    } else {
+                      isAllowed = true
+                      break
+                    }
+                  }
+                }
+              } else {
+                isAllowed = true // no rules for this specific chain, so ignore this
```
mariacarmina, line 229: Can you detail here why isAllowed is true anyways? There is set to true before when meeting a condition.

mariacarmina (commented):
some comments on my side.

mariacarmina (commented):
Can we have some simple tests as well?

src/components/Indexer/utils.ts
paulo-ocean: yes, its merged. and that comment is a reference to it, cause that is the place where the logic was defined..

src/components/Indexer/utils.ts
paulo-ocean: sure, its simple, and its written on the comment.. IF there is NO rule defined for that chain, it behaves the same way of NOT having the `ALLOWED_VALIDATORS_LIST` var at all, which is to `ALLOW`

jamiehewitt15 (approved):
Looks good to me

PR #834, opened 2025-02-04
fix: return correct message on policy server call

AdriGeorge:
Fixes # .

Changes proposed in this PR:

- add return message if call is 200

need to set the env POLICY_SERVER_URL for example equal to http://ps.oceanenterprise.io:8000/


mariacarmina (approved):
lgtm

paulo-ocean (approved):

PR #836, opened 2025-02-05
add AUTHORIZED_DECRYPTERS_LIST

### LINKED ISSUE #811: Config env AUTHORIZED_DECRYPTERS_LIST
We already have AUTHORIZED_DECRYPTERS env, which decides which addresses are authorized to decrypt data.

- add `AUTHORIZED_DECRYPTERS_LIST` env variable, which looks like
```json
{
 "1": ["0x123","0x124"],
 "2": ["0x345"]
}
```

which means:
  
Ocean-node will check both `AUTHORIZED_DECRYPTERS` and `AUTHORIZED_DECRYPTERS_LIST` (if they are defined) and if an address is a match on any , then it's allowed

paulo-ocean:
Fixes #811 

Changes proposed in this PR:

- add  `AUTHORIZED_DECRYPTERS_LIST` support (accessList) 
- 
-


mariacarmina (approved):
lgtm

mariacarmina:
If you can, it would be nice to have some tests to check the functionality? create an accesslist for one authorized decryptor, then remove it to see if 403 error pops up when the same decryptor is mentioned in the env.

.env.example
```diff
@@ -20,6 +20,7 @@ export FEE_AMOUNT=
 export ADDRESS_FILE=
 export NODE_ENV=
 export AUTHORIZED_DECRYPTERS=
+export AUTHORIZED_DECRYPTERS_LIST=
```
jamiehewitt15, line 23: I'm not a huge fan of the name `AUTHORIZED_DECRYPTERS_LIST`, it's a bit too similar to `AUTHORIZED_DECRYPTERS` and not immediately obvious what the difference is. `ACCESS_LIST_CONTRACT_ADRESSES` would be a lot clearer. It's not a deal breaker though

jamiehewitt15 (approved):
Other than the minor comment above, it all looks good

.env.example
paulo-ocean: well, i'm not a big fan of neither one :-) but we're adding the `_LIST`  suffix to all, when its related with access lists, so at least we have a standard...

jamiehewitt15: Yeah, I realise that. It's not a huge deal

PR #841, opened 2025-02-12
add allowed admins access list

### LINKED ISSUE #812: Config env ALLOWED_ADMINS_LIST
We already have ALLOWED_ADMINS env, which decides which addresses are allowed admins.

- add `ALLOWED_ADMINS` env variable, which looks like
```json
{
 "1": ["0x123","0x124"],
 "2": ["0x345"]
}
```

which means:
  
Ocean-node will check both `ALLOWED_ADMINS` and `ALLOWED_ADMINS_LIST` (if they are defined) and if an address is a match on any , then it's allowed

paulo-ocean:
Fixes #812 

Changes proposed in this PR:

- add support for access lists on .env var ALLOWED_ADMINS_LIST
- when we check for Admins... we check also on Smart Contracts (if defined in config) - Concatenate both ALLOWED_ADMINS_LIST and ALLOWED_ADMINS
- refactor of Handler interfaces now we have `Basehandler` + `CommandHandler` + `AdminCommandHandler` ( interfaces  `IValidatecommandHandler` + `IValidateAdminCommandHandler`) 

For the admin commands we need SC access list checks on command validation, since these are `asynchronous` operations and before all the parameters validations were `synchronous`, had to refactor the interfaces to avoid an even bigger refactor.. so `regular` commands and `admin` ones have slightly different validation function signatures (the `IValidateXXX`) interfaces


src/utils/auth.ts
```diff
@@ -40,3 +47,55 @@ export function validateAdminSignature(
...
+    const supportedChains: string[] = Object.keys(config.supportedNetworks)
+    const accessListsChainsListed = Object.keys(config.allowedAdminsList)
+    for (const chain of supportedChains) {
+      const { chainId, network, rpc, fallbackRPCs } = RPCS[chain]
+      const blockchain = new Blockchain(rpc, network, chainId, fallbackRPCs)
+
+      // check the access lists for this chain
+      if (accessListsChainsListed.length > 0 && accessListsChainsListed.includes(chain)) {
+        for (const accessListAddress of config.allowedAdminsList[chainId]) {
+          // instantiate contract and check addresses present + balanceOf()
```
jamiehewitt15, line 82: You say that you are checking addresses present + balanceOf() but I don't see where you are checking balanceOf here? Also, I'm not sure why we the balance of the account is relevant?

jamiehewitt15: yeah, makes sense

jamiehewitt15 (approved):
Looks good to me

mariacarmina (approved):
lgtm

PR #842, opened 2025-02-13
Test if dashboard changes are already committed

### LINKED ISSUE #838: Add test to check if the dashboard files are up to date
Add test to check if the dashboard files are up to date.

jamiehewitt15:
Fixes #838 

Changes proposed in this PR:

- Test if dashboard changes are already committed
- 
-


dashboard/src/pages/index.tsx
```diff
@@ -9,7 +9,7 @@ export default function Home() {
   return (
     <>
       <Head>
-        <title>Ocean nodes</title>
+        <title>Ocean nodes - TEST</title>
```
paulo-ocean, line 12: why this name change?

jamiehewitt15: I'll change this back. I wanted to test if this would cause the test to fail, i.e. committing a change to the dashboard without building it... It should have failed but didn't so I need to look it.

jamiehewitt15: The issue was due to running it as part of the integration tests, where the CI does `npm run build` - so that meant the tests would never fail. I've moved it to a separate CI flow now

paulo-ocean (approved):
LGTM

jamiehewitt15:
Looking for a second review on this @bogdanfazakas @mariacarmina @alexcos20

mariacarmina (approved):
lgtm

PR #846, opened 2025-02-24
add signature check to free compute

alexcos20:
Changes proposed in this PR:

- add signature check for free compute
- 
-


jamiehewitt15 (commented):
I think you probably need to update the tests?

paulo-ocean (approved):
lgtm

Issue #804, opened 2025-01-16
Credentials:  Define how rules are applied (match `all` or match `any`)


MBadea17:
Currently, the way Ocean Node handles the _web3 address-based access_ credentials is to allow access to everybody if no credentials are defined for an asset. 

To increase the security of the assets shared using Ocean technology in enterprise-grade environments, we need to make the Ocean Node more restrictive when handling the assets with no address-based credentials defined. Therefore, the behavior of the Ocean Node needs to be updated as follows:
- if no address-based credentials are defined (both _allow_ and _deny_ lists are empty), access to the asset is restricted to everybody;
- to allow access to everybody, the symbol * will be used in the _allow_ list;
- if a web3 address is present on both _deny_ and _allow_ lists, the _deny_ list takes precedence, and access to the asset is denied for the respective address.

Also, the structure of the _credentials_ object within the DDO needs to be updated to accommodate the credentials used by the newly added Policy Server component, as follows:
```
{
  "credentials": {
    "allow": [
      {
        "type": "address",
        "values": ["0x123", "0x456"]
      },
      {
	"type": "PS-specific Type",
	"values": [{}]
      }    
    ],
    "deny": [
      {
        "type": "address",
        "values": ["0x2222", "0x333"]
      }
    ]
  }
}
```
The "_address_" type object includes the addresses-based credentials (directly handled by Ocean Node) while the "_PS-specific Type_" object includes the credentials that the  Policy Server will assess. The type of Policy Server access credentials will differ from one Policy Server implementation to another.  For instance, the Policy Server developed for walt.id SSI stack will handle credentials of type "_SSIpolicy_".

paulo-ocean:
Adding some comments, as discussed on the slack channel.
The  '*' can be an element of the list:
```
"allow": [
      {
        "type": "address",
        "values": ["*"]
      },
```
(this would allow all)

Also, we might want a env var to enable/disable the new behavior

alexcos20:
The agreed structure will look like:

```json
{
  "credentials": {
    "match_allow": "all",
    "match_deny": "any",
    "allow": [
      {
        "type": "address",
        "values": ["0x123", "0x456"]
      },
      {
	"type": "PS-specific Type",
	"values": [{}]
      }    
    ],
    "deny": [
      {
        "type": "address",
        "values": ["0x2222", "0x333"]
      }
    ]
  }
}
```

Where:
[...]

PR #849, opened 2025-02-26
start refactoring creds behaviour for address

### LINKED ISSUE #810: Credentials address: Change default behaviour
With the addition of #809 , we need to change the default behavior, in order to have more strict access.

In a nutshell,  when type=='address':
  - for allow rules:  if value does not exist or it's empty -> no one has access.  if value list has at least one element, check it
  - for deny rules:  if value does not exist or it's empty -> there is no deny list.  if value list has at least one element, check it

paulo-ocean:
Fixes #810 

Changes proposed in this PR:

- change creds behaviour for type 'address'
- refactor + some aux functions
-

### DIFF
--- a/src/utils/credentials.ts
+++ b/src/utils/credentials.ts
@@ -13,7 +13,7 @@ import { isDefined } from './util.js'
 export function findCredential(
   credentials: Credential[],
   consumerCredentials: Credential
-) {
+): Credential | undefined {
   return credentials.find((credential) => {
     if (Array.isArray(credential?.values)) {
       if (credential.values.length > 0) {


src/utils/credentials.ts
```diff
@@ -52,29 +73,69 @@ export function hasAddressMatchAllRule(credentials: Credential[]): boolean {
...
-    type: 'address',
+    type: CREDENTIAL_TYPES.ADDRESS, // 'address',
     values: [String(consumerAddress)?.toLowerCase()]
   }
 
-  const accessGranted = true
+  // if no address-based credentials are defined (both allow and deny lists are empty), access to the asset is restricted to everybody;
+  // to allow access to everybody, the symbol * will be used in the allow list;
+  // if a web3 address is present on both deny and allow lists, the deny list takes precedence
+  // and access to the asset is denied for the respective address.
```
jamiehewitt15, line 89: I'm kind of surprised that we agreed to implement this change. So the default behaviour is that if an asset published, no one can access it? Seems contrary to our mission statement 🤔

paulo-ocean: that's copied from the specs yes [https://github.com/oceanprotocol/ocean-node/issues/804](https://github.com/oceanprotocol/ocean-node/issues/804)

jamiehewitt15: It makes to enterprise users... In the past Ocean has always taken the approach that everything is open by default and  Enterprises must take extra steps if they don't like the default. So this is a fundamental shift in that approach

paulo-ocean: i'm not gonna discuss that... please check with @alexcos20 . I just followed the specs and they make sense to me

mariacarmina (approved):
LGTM!

PR #867, opened 2025-03-07
re-indexing old DDOs

### LINKED ISSUE #806: Indexer: decide when to reindex
As soon as we merge #525 and derivates(526-528), we will change the structure of the DDO.
But existing nodes will not reindex old assets, only new ones, leading to inconsistency .

So, node must know when to reindex, based on version.

Proposal:
 - use node version(take it from package.lock) to decide if action is needed
 - hardcode a "min_version" in the code.  At startup, if db.version is null, or lower that "min_version", then reset all chains and store current version (take it from package.lock)

jamiehewitt15:
Fixes #806

Changes proposed in this PR:

- Saving node version in DB
- Checking if version number in DB is older than `0.2.2`
- Initiate re-indexing old DDOs if the version is old or null
- Adding tests


src/components/Indexer/index.ts
```diff
@@ -405,4 +414,60 @@ export class OceanIndexer {
       }
     }
   }
+
+  /**
+   * Gets the current node version from package.json
+   */
+  private getCurrentVersion(): string {
```
mariacarmina, line 421: the current version can be retrieved from `process.env.npm_package_version`, no need to create this function to retrieve the current version of the node

mariacarmina (commented):
I'll create a test with reindex command too

src/components/Indexer/index.ts
```diff
@@ -405,4 +414,60 @@ export class OceanIndexer {
...
+        `Reindexing needed: DB version ${
+          dbVersion || 'not set'
+        } is older than minimum required ${this.MIN_REQUIRED_VERSION}`
+      )
+
+      // Reindex all chains
+      for (const chainID of this.supportedChains) {
+        const chainIdNum = Number(chainID)
+        INDEXER_LOGGER.info(`Triggering reindexing for chain ${chainIdNum}`)
+        await this.resetCrawling(chainIdNum)
```
mariacarmina, line 463: should we call instead addReindexTask instead? We should call resetCrawling through the handler.

mariacarmina, line 463: we should check also the reindex job status

paulo-ocean: we already have a `reindexChainHandler` why not use it?
and yes, we should watch job status as well

src/components/Indexer/index.ts
paulo-ocean: imo, it looks a bit cumbersome to read the version from package.json
how will this work for tests for instance?
It looks that if the reading of the json file fails for some reason, we always default to trigger a reindexing ?

src/components/Indexer/index.ts
mariacarmina: reindexChainHandler requires admin signature, I have understood that we need to reindex the chain not depending of the role.

src/components/database/ElasticSearchDatabase.ts
```diff
@@ -17,6 +17,9 @@ export class ElasticsearchIndexerDatabase extends AbstractIndexerDatabase {
   private client: Client
   private index: string
 
+  // constant for the node version document ID
+  private static readonly VERSION_DOC_ID = 'node_version'
```
paulo-ocean, line 21: Can't we have this field on `AbstractIndexerDatabase` instead?
We have this field duplicated on both  Elastic and Typesense classes, while both extend from `AbstractIndexerDatabase`, 
we could have for instance
`protected static readonly VERSION_DOC_ID = 'node_version'` there

src/components/Indexer/index.ts
```diff
@@ -35,13 +38,19 @@ export class OceanIndexer {
...
   private supportedChains: string[]
   private workers: Record<string, Worker> = {}
+  private MIN_REQUIRED_VERSION = '0.2.2'
 
   constructor(db: Database, supportedNetworks: RPCS) {
     this.db = db
     this.networks = supportedNetworks
     this.supportedChains = Object.keys(supportedNetworks)
     INDEXING_QUEUE = []
     this.startThreads()
```
paulo-ocean, line 46: Should we even start the threads if we are going to reindex ? I mean.. I know the reindex/reset is caught on the while() loop of the crawler, so they should be running... but here we don't even wait for `startThreads` completion
Imo, i think we should check this logic more carefully... maybe wait until we start the threads, or maybe some timeout ??

src/components/Indexer/index.ts
```diff
@@ -35,13 +36,19 @@ export class OceanIndexer {
...
+
+    // Check if reindexing is needed
+    this.checkAndTriggerReindexing().catch((error) => {
```
paulo-ocean, line 49: Should we even start the threads if we are going to reindex ? I mean.. I know the reindex/reset is caught on the while() loop of the crawler, so they must be running... but here we don't even wait for `startThreads` completion
Imo, i think we should check this logic more carefully... maybe wait until we start the threads, or maybe some timeout ??

src/components/database/ElasticSearchDatabase.ts
jamiehewitt15: I think it might be cleaner to store the version in a separate DB. This is not really what the indexer db is for so it's a bit cumbersome

src/test/integration/database.test.ts
```diff
@@ -201,12 +201,6 @@ describe('IndexerDatabase CRUD', () => {
   })
 
   describe('Node Version Management', () => {
-    it('should have null version initially', async () => {
```
jamiehewitt15: @mariacarmina why did you remove this test? The db should start up with `nodeVersion === null` it's quite a simple test...

mariacarmina: It retrieved by default the latest version, maybe from other indexer tests,the version was already set to latest.

jamiehewitt15: So that means there is a bug that needs to be fixed. We should fix the bug, rather than removing the test

mariacarmina: If the indexer was already created in previous tests, that version might be already set, or the code from indexer takes by default the latest version, why should we demonstrate if initially is null, from now on we must always have a version for indexer, so I do not see the scope for it.

src/components/database/ElasticSearchDatabase.ts
mariacarmina: @jamiehewitt15 the version is used for indexer in order to know what type of ddos will index, as clarifying with Alex, we agreed to keep in a separate column/document from indexer db. why should the version be separate from indexer?

src/test/integration/database.test.ts
jamiehewitt15: I added it back in and the tests are passing, so it doesn't look there was any issue

src/components/database/ElasticSearchDatabase.ts
jamiehewitt15: Storing two different document types in the indexer DB isn't following separation of concerns. It creates a messy situation as lastBlock is required field, that won't be in the documents storing the node version.

src/components/Indexer/index.ts
jamiehewitt15: I did a refactor of this so that we wait for the `startThreads` completion. Let me know what you think

paulo-ocean: looks good

src/components/Indexer/index.ts
```diff
@@ -35,13 +36,18 @@ export class OceanIndexer {
...
 
   constructor(db: Database, supportedNetworks: RPCS) {
     this.db = db
     this.networks = supportedNetworks
     this.supportedChains = Object.keys(supportedNetworks)
     INDEXING_QUEUE = []
-    this.startThreads()
+
+    // Thread initialization is await-able
+    this.initializationPromise = this.initializeThreads()
```
paulo-ocean, line 50: i think we have an issue here... will detail it later bellow

src/components/Indexer/index.ts
```diff
@@ -306,10 +312,60 @@ export class OceanIndexer {
...
 
-  public async resetCrawling(
-    chainId: number,
-    blockNumber?: number
-  ): Promise<JobStatus | null> {
+  /**
+   * Initialize threads and check for reindexing
+   */
+  private async initializeThreads(): Promise<boolean> {
+    const threadsStarted = await this.startThreads()
```
paulo-ocean, line 319: this only returns "true" IF ALL the threads started successfully... we we have 2 chains to index and 1 fails... it will never return true and we keep repeating recursively ... The initial idea was just "wait" for the promises to finish .. but not exactly for a positive value (when number of threads started === number of chains)

paulo-ocean:
```suggestion
    const threadsStarted = await this.startThreads()
    await this.startThreads()
    //  if (threadsStarted) {
```

src/components/Indexer/index.ts
```diff
@@ -306,10 +312,60 @@ export class OceanIndexer {
...
+
+    if (threadsStarted) {
```
paulo-ocean, line 321:
```suggestion
    if (threadsStarted) {
    //  if (threadsStarted) {
```

src/components/Indexer/index.ts
```diff
@@ -306,10 +312,60 @@ export class OceanIndexer {
...
+  public getInitializationPromise(): Promise<boolean> {
+    return this.initializationPromise
+  }
+
+  /**
+   * Reset crawling with initialization check
+   */
+  public async resetCrawling(chainId: number, blockNumber?: number): Promise<JobStatus> {
+    // Ensure threads are initialized before running commands
+    if (!this.threadsInitialized) {
```
paulo-ocean, line 365: not sure if we need any of this..
when we call the resetCrawling, we already check if the thread is running, and restart if needed:
```
const isRunning = runningThreads.get(chainId)
    // not running, but still on the array
    if (!isRunning && this.workers[chainId]) {
```

paulo-ocean: i would just call:
`this.initializeThreads()` on the constructor and remove the need for assignment/variable `initializationPromise`
and then on initializeThreads:
private async initializeThreads(): Promise<void> {
    await this.startThreads()

      // Only check for reindexing if threads started successfully
      try {
        await this.checkAndTriggerReindexing()
      } catch (error) {
        INDEXER_LOGGER.error(
          `Error during version check and reindexing: ${error.message}`
        )
      }
    }

jamiehewitt15: Ok yeah, good point. I have simplified it and removed `initializationPromise`. I just have a check now to ensure we don't try to reset crawling before the basic initialization process is complete:

```
    if (!this.threadsInitialized) {
      await new Promise((resolve) => {
        const checkInterval = setInterval(() => {
          if (this.threadsInitialized) {
            clearInterval(checkInterval)
            resolve(true)
          }
        }, 100)
      })
```

src/components/Indexer/index.ts
jamiehewitt15: Yeah, I see what you're saying. I have implemented this.

paulo-ocean:
adding to the stuff we discussed on the meeting, if i got it right... it seems the approach could be much simpler ..
basically, we just need to check if we need to reindex (based on the version) BEFORE calling `startThreads` as usual
In case we do need to to it... we just need to reset the `lastIndexedBlock` (maybe set it to null)
probably call `deleteAllAssetsFromChain` too
and that's it

jamiehewitt15:
> adding to the stuff we discussed on the meeting, if i got it right... it seems the approach could be much simpler .. basically, we just need to check if we need to reindex (based on the version) BEFORE calling `startThreads` as usual In case we do need to to it... we just need to reset the `lastIndexedBlock` (maybe set it to null) probably call `deleteAllAssetsFromChain` too and that's it when the crawler thread starts.. we just start indexing from whatever block it was specified on the RPCS ... a fresh start with new ddo structure

Yeah, that makes sense. I'll update it

mariacarmina:
> adding to the stuff we discussed on the meeting, if i got it right... it seems the approach could be much simpler .. basically, we just need to check if we need to reindex (based on the version) BEFORE calling `startThreads` as usual In case we do need to to it... we just need to reset the `lastIndexedBlock` (maybe set it to null) probably call `deleteAllAssetsFromChain` too and that's it when the crawler thread starts.. we just start indexing from whatever block it was specified on the RPCS ... a fresh start with new ddo structure

Why do we need to delete the old assets with a different structure, can you please explain @paulo-ocean ? Based on the version we can differentiate what will index.

paulo-ocean:
> > adding to the stuff we discussed on the meeting, if i got it right... it seems the approach could be much simpler .. basically, we just need to check if we need to reindex (based on the version) BEFORE calling `startThreads` as usual In case we do need to to it... we just need to reset the `lastIndexedBlock` (maybe set it to null) probably call `deleteAllAssetsFromChain` too and that's it when the crawler thread starts.. we just start indexing from whatever block it was specified on the RPCS ... a fresh start with new ddo structure
> 
> Why do we need to delete the old assets with a different structure, can you please explain @paulo-ocean ? Based on the version we can differentiate what will index.

?? How does the reindex work? Didn't we do it for the reindexChain as well?
you want to keep 2 different versions of the indexed asset?

mariacarmina:
> > > adding to the stuff we discussed on the meeting, if i got it right... it seems the approach could be much simpler .. basically, we just need to check if we need to reindex (based on the version) BEFORE calling `startThreads` as usual In case we do need to to it... we just need to reset the `lastIndexedBlock` (maybe set it to null) probably call `deleteAllAssetsFromChain` too and that's it when the crawler thread starts.. we just start indexing from whatever block it was specified on the RPCS ... a fresh start with new ddo structure
> 
> > 
> 
> > Why do we need to delete the old assets with a different structure, can you please explain @paulo-ocean ? Based on the version we can differentiate what will index.
> 
> 
> 
> ?? How does the reindex work? Didn't we do it for the reindexChain as well? (actually you worked on it :-) )
> 
> you want to keep 2 different versions of the indexed asset?
> 
> that's not a reindex. We would not need this task if we're going to keep old assets and just starting indexing the new ones with different structure

I think you might misunderstood, I didn't ask how reindex works and I know it was assigned to me that reindexChain command, I was asking why we are trying to replicate the logic from that command including deletion of the assets, since it is not needed, we'll continue indexing the assets with different version, that was my understanding, so we do not need to remove the old assets.

mariacarmina:
> > > > adding to the stuff we discussed on the meeting, if i got it right... it seems the approach could be much simpler .. basically, we just need to check if we need to reindex (based on the version) BEFORE calling `startThreads` as usual In case we do need to to it... we just need to reset the `lastIndexedBlock` (maybe set it to null) probably call `deleteAllAssetsFromChain` too and that's it when the crawler thread starts.. we just start indexing from whatever block it was specified on the RPCS ... a fresh start with new ddo structure
> > 
> > 
> > > 
> > 
> > 
> > > Why do we need to delete the old assets with a different structure, can you please explain @paulo-ocean ? Based on the version we can differentiate what will index.
> > 
> > 
> > ?? How does the reindex work? Didn't we do it for the reindexChain as well? (actually you worked on it :-) )
> > you want to keep 2 different versions of the indexed asset?
> > that's not a reindex. We would not need this task if we're going to keep old assets and just starting indexing the new ones with different structure
> 
> I think you might misunderstood, I didn't ask how reindex works and I know it was assigned to me that reindexChain command, I was asking why we are trying to replicate the logic from that command including deletion of the assets, since it is not needed, we'll continue indexing the assets with different version, that was my understanding, so we do not need to remove the old assets.

After our discussion I understood the flow because we enforce the latest version for all DDOs, all good :).

paulo-ocean (approved):
LGTM
The way i see it we're just missing this line
`await this.checkAndTriggerReindexing()` at the beginning of `startThreads` function
```
public async startThreads(): Promise<boolean> {
    await this.checkAndTriggerReindexing()
...

```

mariacarmina:
> adding to the stuff we discussed on the meeting, if i got it right... it seems the approach could be much simpler .. basically, we just need to check if we need to reindex (based on the version) BEFORE calling `startThreads` as usual In case we do need to to it... we just need to reset the `lastIndexedBlock` (maybe set it to null) probably call `deleteAllAssetsFromChain` too and that's it when the crawler thread starts.. we just start indexing from whatever block it was specified on the RPCS ... a fresh start with new ddo structure

Why set the lastIndexedBlock should be set to null? It will throw error in the `update` db function, shouldn't be the deployed block depending on the chain? getDeployedContractBlock(chainId) to reindex the assets from the default start, mentioned in address.json

jamiehewitt15:
> Why set the lastIndexedBlock should be set to null? It will throw error in the update db function, shouldn't be the deployed block depending on the chain? getDeployedContractBlock(chainId) to reindex the assets from the default start, mentioned in address.json

I've updated this

paulo-ocean:
> > Why set the lastIndexedBlock should be set to null? It will throw error in the update db function, shouldn't be the deployed block depending on the chain? getDeployedContractBlock(chainId) to reindex the assets from the default start, mentioned in address.json
> 
> I've updated this

why?
Its not that simple. What about the startBlock on the RPCS? we just ignore that?

mariacarmina:
> > > adding to the stuff we discussed on the meeting, if i got it right... it seems the approach could be much simpler .. basically, we just need to check if we need to reindex (based on the version) BEFORE calling `startThreads` as usual In case we do need to to it... we just need to reset the `lastIndexedBlock` (maybe set it to null) probably call `deleteAllAssetsFromChain` too and that's it when the crawler thread starts.. we just start indexing from whatever block it was specified on the RPCS ... a fresh start with new ddo structure
> > 
> > 
> > Why set the lastIndexedBlock should be set to null? It will throw error in the `update` db function, shouldn't be the deployed block depending on the chain? getDeployedContractBlock(chainId) to reindex the assets from the default start, mentioned in address.json
> 
> And what about the `startBlock` ?

That's what I wanted to propose either deployedBlock if startBlock is not defined we have function in Indexer.

mariacarmina (approved):
lgtm

alexcos20 (approved):

PR #869, opened 2025-03-07
fix status code if policy server not available

### LINKED ISSUE #868: Check policy server passthrough response
When the passThrough request fails (for instance because the server is down)
we should return a 400 status.
Seems we are returning a 200 at the moment

paulo-ocean:
Fixes #868

Changes proposed in this PR:

- fix status code if policy server fails or not available (400)
- 
-


alexcos20 (approved):

jamiehewitt15 (approved):
Looks good to me

PR #872, opened 2025-03-10
Setup SQLite for config storing version.

mariacarmina:
Fixes https://github.com/oceanprotocol/s1-dapps/issues/142 .

Changes proposed in this PR:

- 
- 
-


jamiehewitt15:
You get this error when you reuse the same database in different tests as it creates a race condition:

> TypeError: Cannot read properties of null (reading 'version')

You can solve it by using a different DB: 

https://github.com/oceanprotocol/ocean-node/blob/9d2538d9c56882b1d1e05f39d2fdb7b42e115b0e/src/test/integration/database.test.ts#L27

jamiehewitt15:
But you should actually just remove the tests for the Typesense database, as it's been removed

mariacarmina:
> You get this error when you reuse the same database in different tests as it creates a race condition:
> 
> > 1. Version Database
> >    should have null version initially:
> >    TypeError: Cannot read properties of null (reading 'version')
> 
> You can solve it by using a different DB:
> 
> https://github.com/oceanprotocol/ocean-node/blob/9d2538d9c56882b1d1e05f39d2fdb7b42e115b0e/src/test/integration/database.test.ts#L27

But that database is used for indexer database from my tests

src/test/integration/database.test.ts
```diff
@@ -521,38 +524,45 @@ describe('MetadataQuery', () => {
 
 describe('Version Database', () => {
```
jamiehewitt15, line 525: It's a minor point but it would be nice to move the SQLite tests to a separate file. This file is very big

package.json
```diff
@@ -32,7 +32,7 @@
     "mocha-light": "mocha --node-env=test --config .mocharc.json --exclude \"./dist/test/integration/compute.test.js\"",
     "test": "npm run lint && npm run test:unit:cover && npm run test:integration:cover",
     "test:unit": "npm run build-tests && npm run mocha \"./dist/test/unit/**/*.test.js\"",
-    "test:integration": "npm run build-tests && npm run mocha \"./dist/test/integration/**/*.test.js\"",
+    "test:integration": "npm run build-tests && npm run mocha \"./dist/test/integration/database.test.js\"",
```
jamiehewitt15, line 35: Can you change this back?

mariacarmina: not yet, still debugging

src/test/integration/database.test.ts
mariacarmina: Ok will do

package.json
mariacarmina: done, fixed

src/test/integration/database.test.ts
mariacarmina: done.

src/components/Indexer/index.ts
```diff
@@ -468,23 +468,24 @@ export class OceanIndexer {
    */
   public async checkAndTriggerReindexing(): Promise<void> {
     const currentVersion = process.env.npm_package_version
-    const dbActive = this.getDatabase()
-    if (!dbActive || !(await isReachableConnection(dbActive.getConfig().url))) {
-      INDEXER_LOGGER.error(`Giving up reindexing. DB is not online!`)
+    if (!this.db.version) {
+      INDEXER_LOGGER.error(`Giving up reindexing...`)
       return
     }
```
jamiehewitt15, line 474: This is not right, If `this.db.version == null` it will stop here. That's not what we want, if we have a null value we want to reindex.

mariacarmina: No, this means that version db does not exist, it is not checking if a specific version exists, that checks if the database of versions exist. Should we remove? Because database class creates a db from scratch if it does not exist

jamiehewitt15: ah ok I see. Yeah, that check will cause a problem. They only will have the version DB when a value has been inputted for the first time, so we should treat no version DB the same null.

src/components/Indexer/index.ts
```diff
@@ -502,7 +503,7 @@ export class OceanIndexer {
       }
 
       // Update the version in the database
-      await dbActive.version.setNodeVersion(currentVersion)
+      await this.db.version.update(currentVersion, dbVersion.version)
```
jamiehewitt15, line 506: Why are we using update if it doesn't exist yet?

src/components/database/SQLLiteConfigDatabase.ts
```diff
@@ -0,0 +1,108 @@
+import fs from 'fs'
+import path from 'path'
+import { OceanNodeDBConfig } from '../../@types/OceanNode.js'
+import { DATABASE_LOGGER } from '../../utils/logging/common.js'
+import { GENERIC_EMOJIS, LOG_LEVELS_STR } from '../../utils/logging/Logger.js'
+import { AbstractVersionDatabase } from './BaseDatabase.js'
+import { SQLiteProvider } from './sqlite.js'
+import { TypesenseSchema } from './TypesenseSchemas.js'
+
+export class SQLLiteConfigDatabase extends AbstractVersionDatabase {
```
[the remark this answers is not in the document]
jamiehewitt15: Yeah, it could be a lot simpler

src/components/database/TypesenseSchemas.ts
```diff
@@ -62,6 +62,11 @@ export const typesenseSchemas: TypesenseSchemas = {
     enable_nested_fields: true,
     fields: [{ name: 'nonce', type: 'int64' }]
   },
+  configSchemas: {
+    name: 'config',
+    enable_nested_fields: true,
+    fields: [{ name: 'version', type: 'string' }]
+  },
```
jamiehewitt15, line 69: Do we really want the schema to be here in the Typesense schema? I understand that we did it that way with the nonce schema, so that the nonce could be stored in either Typesense or SQLite. But with the config DB it is clear that we will never store it in Typesense.

src/components/database/sqlite.ts
```diff
@@ -62,19 +94,73 @@ export class SQLiteProvider implements DatabaseProvider {
     })
   }
 
+  // eslint-disable-next-line require-await
+  async retrieveLatestVersion() {
+    const selectSQL = `
+      SELECT version FROM ${this.schemaConfig.name} ORDER BY id DESC LIMIT 1;
```
jamiehewitt15, line 100: If you go ahead and follow @paulo-ocean suggestion of making it a simple key/value store, you don't need this selection by order... There will only be one value.

src/components/database/TypesenseSchemas.ts
paulo-ocean: we don't.. that relates to my previous comments as well... and its true that at the beginning the nonce was on Typesense (like everything else)... but the Nonce is currently just SQL Lite... so its both that need a refactor: Nonce and Config DB... we don't need schemas neither half the abstractions that only made sense for Elastic/Typesense

mariacarmina: fixed

src/components/database/SQLLiteConfigDatabase.ts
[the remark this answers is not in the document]
mariacarmina: fixed

src/components/Indexer/index.ts
mariacarmina: fixed

src/components/database/sqlite.ts
```diff
@@ -62,6 +95,19 @@ export class SQLiteProvider implements DatabaseProvider {
...
+      SELECT value FROM ${this.configSchema};
+    `
+    return new Promise<{ version: string | null }>((resolve, reject) => {
+      this.db.get(selectSQL, [], (err, row: { value: string } | undefined) => {
+        if (err) reject(err)
+        else resolve(row ? { version: row.value } : { version: null }) // Returns null if no version exists
+      })
+    })
+  }
+
```
jamiehewitt15, line 110: Rather than retrieveVersion, we should have `retrieveValue(key: string)` which takes the `key` as an argument. That way we can use this same method for retrieving all future values e.g. `retrieveValue("version")` / `retrieveValue("otherKey")`

src/test/integration/versionDatabase.test.ts
```diff
@@ -0,0 +1,84 @@
+import { Database } from '../../components/database/index.js'
```
jamiehewitt15, line 1: Can we rename this file to `/configDatabase.test.ts`

src/components/database/sqlite.ts
jamiehewitt15: Also looks like this is duplicated in [src/components/database/SQLLiteConfigDatabase.ts](https://github.com/oceanprotocol/ocean-node/pull/872/files#diff-a2ce10b0bc97914e724b5733016e16ca07bbba18327806f5ba7daf29d0d6d300)

src/components/database/SQLLiteConfigDatabase.ts
```diff
@@ -0,0 +1,56 @@
...
+import { DATABASE_LOGGER } from '../../utils/logging/common.js'
+import { GENERIC_EMOJIS, LOG_LEVELS_STR } from '../../utils/logging/Logger.js'
+import { SQLiteProvider } from './sqlite.js'
+
+export class SQLLiteConfigDatabase {
+  private provider: SQLiteProvider
+
+  constructor() {
+    return (async (): Promise<SQLLiteConfigDatabase> => {
+      DATABASE_LOGGER.info('Version Database initiated with SQLite provider')
```
jamiehewitt15, line 12:
```suggestion
      DATABASE_LOGGER.info('Config Database initiated with SQLite provider')
```

src/components/database/SQLLiteConfigDatabase.ts
```diff
@@ -0,0 +1,56 @@
...
+      DATABASE_LOGGER.logMessageWithEmoji(
+        errorMsg,
+        true,
+        GENERIC_EMOJIS.EMOJI_CROSS_MARK,
+        LOG_LEVELS_STR.LEVEL_ERROR
+      )
+      return null
+    }
+  }
+}
```
jamiehewitt15, line 56: We just need two methods here: `createOrUpdateValue(key: string, value: string)` and `getValue(key: string)` and then we will be able to use them for all future use cases with the config DB. By going down a path of implementing `retrieveLatestVersion` you are setting yourselve up to have a new getter function for every different value which over complicates it.

src/components/database/sqlite.ts
```diff
@@ -33,10 +35,26 @@ export class SQLiteProvider implements DatabaseProvider {
...
+    return new Promise<void>((resolve, reject) => {
+      this.db.run(createTableSQL, (err) => {
+        if (err) reject(err)
+        else resolve()
+      })
+    })
+  }
+
   // eslint-disable-next-line require-await
   async create(address: string, nonce: number) {
```
paulo-ocean, line 55: looking good, but better if we we refactor some function names to be more clear...
example:
for config we have:
`createConfig`
maybe for nonce should be 
`createNonce` instead of `create` only
same for create table
for config we have:
`createTableForConfig`
and for nonce only `createTable` maybe `createTableForNonce` ?
my point is that the names for 'config' are quite explanatory.... while for nonce they are quite generic...
maybe we should align them

src/components/database/SQLLiteConfigDatabase.ts
mariacarmina: `createOrUpdateValue(key: string, value: string)`, I would rename createOrUpdateConfig, because you might also want to add new key along with value, not only the value

src/components/database/sqlite.ts
mariacarmina: done.

src/components/database/SQLLiteConfigDatabase.ts
mariacarmina: fixed

src/components/database/SQLLiteConfigDatabase.ts
mariacarmina: done

src/components/database/sqlite.ts
mariacarmina: done.

src/test/integration/configDatabase.test.ts
```diff
@@ -0,0 +1,88 @@
...
+import { expect, assert } from 'chai'
+import { OceanIndexer } from '../../components/Indexer/index.js'
+import { getMockSupportedNetworks } from '../utils/utils.js'
+import { SQLLiteConfigDatabase } from '../../components/database/SQLLiteConfigDatabase.js'
+import { DB_TYPES } from '../../utils/constants.js'
+import { OceanNodeDBConfig } from '../../@types/OceanNode.js'
+
+const versionConfig: OceanNodeDBConfig = {
+  url: 'http://localhost:8108/test-version?apiKey=xyz',
+  dbType: DB_TYPES.TYPESENSE
```
jamiehewitt15, line 11: Why are you still using Typesense?

mariacarmina: I tried with both, emtyConfig and with Typesense Config just to see that it does not impact

src/utils/config.ts
```diff
@@ -201,6 +201,22 @@ function getAuthorizedDecrypters(isStartup?: boolean): string[] {
     isStartup
   )
 }
+
+function getAuthorizedDecryptersList(isStartup?: boolean): AccessListContract | null {
```
paulo-ocean, line 205: what are these changes doing here?
seems we might need a rebase no?

.env.example
```diff
@@ -20,6 +20,7 @@ export FEE_AMOUNT=
 export ADDRESS_FILE=
 export NODE_ENV=
 export AUTHORIZED_DECRYPTERS=
+export AUTHORIZED_DECRYPTERS_LIST=
```
paulo-ocean, line 23: why is all the code from `AUTHORIZED_DECRYPTERS_LIST` (already on main branch) showing up on this PR?

jamiehewitt15: Yeah, looks like there may have been a pull from main

mariacarmina: fixed

src/utils/config.ts
mariacarmina: fixed

src/utils/constants.ts
```diff
@@ -247,6 +247,11 @@ export const ENVIRONMENT_VARIABLES: Record<any, EnvVariable> = {
     value: process.env.AUTHORIZED_DECRYPTERS,
     required: false
   },
+  AUTHORIZED_DECRYPTERS_LIST: {
```
paulo-ocean, line 250: still seeing a lot of accessList related stuff in this PR, seems its not clean yet

src/components/Indexer/index.ts
```diff
@@ -464,9 +464,7 @@ export class OceanIndexer {
           )
         }
       }
-
-      // Update the version in the database
-      await dbActive.version.setNodeVersion(currentVersion)
+      await this.db?.version?.createOrUpdateConfig('version', currentVersion)
```
paulo-ocean, line 467: do we really need these optionals => ?
`await this.db?.version?.createOrUpdateConfig('version', currentVersion)`
and
above
`const dbVersion = await this.db?.version?.retrieveValue()`
why wouldn't the DB NOT exist? But if we need to check it, just do it once.. when we check` this.getDatabase()`

src/components/database/sqlite.ts
```diff
@@ -34,9 +36,25 @@ export class SQLiteProvider implements DatabaseProvider {
   }
 
   // eslint-disable-next-line require-await
-  async create(address: string, nonce: number) {
+  async createTableForConfig() {
+    const createTableSQL = `
+        CREATE TABLE IF NOT EXISTS ${this.configSchema} (
+          key TEXT PRIMARY KEY DEFAULT 'version',
```
paulo-ocean, line 42: i don't think we need the `DEFAULT 'version'` if  we define it as a primary key
when we have a primary key, we have a conflict when we try to introduce the same key again... here we are defaulting the 'version' , which will cause the conflict more likely (if we don't provide a key, it will default to the same over and over)

src/components/Indexer/index.ts
jamiehewitt15: I don't think that's quite right as the SQLite DB doesn't need a connection URL, so no reason to check for `!(await isReachableConnection(dbActive.getConfig().url))`. Should be fine just to check for `dbActive`:

```
    const dbActive = this.getDatabase()
    if (!dbActive) {
      INDEXER_LOGGER.error(`Giving up reindexing. DB is not online!`)
      return
    }
   const dbVersion = await dbActive.version.retrieveValue()
```

jamiehewitt15: Although, on second thoughts... The indexer DB needs to be online otherwise the reindexing will fail. So probably best to do as you suggested @paulo-ocean

src/components/database/index.ts
```diff
@@ -27,12 +27,13 @@ export class Database {
...
   order: AbstractOrderDatabase
   ddoState: AbstractDdoStateDatabase
-  version: AbstractVersionDatabase
+  version: SQLLiteConfigDatabase
 
   constructor(private config: OceanNodeDBConfig) {
     return (async (): Promise<Database> => {
       try {
         this.nonce = await DatabaseFactory.createNonceDatabase(this.config)
+        this.version = await DatabaseFactory.createConfigDatabase()
```
jamiehewitt15, line 36: Why do you call it `this.version` when it is the config database? It should be `this.config`

mariacarmina: I'll rename into sqliteConfig instead of config, because I do not want to mess with config DB

src/utils/constants.ts
mariacarmina: I tried to follow the [commit](https://github.com/oceanprotocol/ocean-node/pull/872/commits/d2bcc38a2569dd12779f6f12c3394b7b554414d9) changes  and those accessList stuff weren't all of them captured :(... I think there are in processEvent too if I am not wrong

mariacarmina: But maybe it is a good idea to have main branch merged with @jamiehewitt15's and then replicate in mine and we will have in the end all branches in sync with main

src/components/Indexer/index.ts
paulo-ocean: allow me to disagree @jamiehewitt15 ..
```
const dbActive = this.getDatabase()
    if (!dbActive) {
```
Only tells you that you have an `Instance` of the database, but that does not mean it is reachable
That is exactly why you also need to
`(await isReachableConnection(dbActive.getConfig().url)).`
Otherwise you try to start threads and reindex on something that is not running

docs/env.md
```diff
@@ -29,7 +29,6 @@ Environmental variables are also tracked in `ENVIRONMENT_VARIABLES` within `src/
 - `MAX_CHECKSUM_LENGTH`: Define the maximum length for a file if checksum is required (Mb). Example: `10`
 - `IS_BOOTSTRAP`: Is this node to be used as bootstrap node or not. Default is `false`.
 - `AUTHORIZED_PUBLISHERS`: Authorized list of publishers. If present, Node will only index assets published by the accounts in the list. Example: `"[\"0x967da4048cD07aB37855c090aAF366e4ce1b9F48\",\"0x388C818CA8B9251b393131C08a736A67ccB19297\"]"`
-- `AUTHORIZED_PUBLISHERS_LIST`: AccessList contract addresses (per chain). If present, Node will only index assets published by the accounts present on the given access lists. Example: `"{ \"8996\": [\"0x967da4048cD07aB37855c090aAF366e4ce1b9F48\",\"0x388C818CA8B9251b393131C08a736A67ccB19297\"] }"`
```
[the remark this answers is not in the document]
mariacarmina: I'll add back, this is on as-is and I made a confusion when cleaning up the merge issues

jamiehewitt15 (approved):
Looks good to me

PR #873, opened 2025-03-11
add support for accessList on download checkCredentials

### LINKED ISSUE #871: extend checkCredentials to support 'accessList' type (on asset download)
Right now we only `checkCredentials`  (ddo and service level) when trying to download an asset.
But we only check for **address based** types. It would be nice to extend `checkCredentials` to also check for smart contract accessList credentials (since we already support accessLists for specific roles, like decrypter, admin, publisher, etc)

paulo-ocean:
Fixes #871 

Changes proposed in this PR:

- support accessList check on checkCredentials (DDO download)
- 
-


mariacarmina (approved):
LGTM, let's merge this into #873

PR #874, opened 2025-03-11
Update error message for invalid peer connection

bogdanfazakas:
Fixes # .

Changes proposed in this PR:

- 
- 
-

### DIFF
--- a/src/components/P2P/index.ts
+++ b/src/components/P2P/index.ts
@@ -668,7 +668,7 @@ export class OceanP2P extends EventEmitter {
       const connection = await this._libp2p.dial(multiaddrs, options)
       if (connection.remotePeer.toString() !== peerId.toString()) {
         response.status.httpStatus = 404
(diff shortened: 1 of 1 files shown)


src/components/P2P/index.ts
```diff
@@ -668,7 +668,7 @@ export class OceanP2P extends EventEmitter {
       const connection = await this._libp2p.dial(multiaddrs, options)
       if (connection.remotePeer.toString() !== peerId.toString()) {
         response.status.httpStatus = 404
-        response.status.error = `Invalid peer on the other side: ${peerId}`
+        response.status.error = `Invalid peer on the other side: ${connection.remotePeer.toString()}`
```
mariacarmina, line 671: Can we include both peer ids? like source and destination?

paulo-ocean: isn't the other one our/current peer? not sure why we need that

paulo-ocean (approved):
lgtm

PR #879, opened 2025-03-12
add some access/credentials tests + emit events if unauthorized

### LINKED ISSUE #840: Create tests for AccessLists new variables (ADMIN, PUBLISHER, VALIDATOR, DECRYPTOR) + refactor common code
We are adding a few accessLists .env variables for different roles and they share pretty much the same format.
Once all the related opened PR are merged we should do some refactoring to use more common / generic functions, and make some integration unit or integration tests

paulo-ocean:
Fixes #840 

Changes proposed in this PR:

- 
- 
-


Maungmaunglwin096062286:
paulo-ocean pushed 1 commit.

cef5ce4 more

Maungmaunglwin096062286:
[paulo-ocean](https://github.com/paulo-ocean) pushed 1 commit.

[cef5ce4](https://github.com/oceanprotocol/ocean-node/commit/cef5ce480b3fbcc52b8a0ccdfa9793306a360e06) more

src/utils/constants.ts
```diff
@@ -91,6 +91,9 @@ export const EVENTS = {
   EXCHANGE_DEACTIVATED: 'ExchangeDeactivated'
 }
 
+// when something at access/credentials level fails
+export const UNAUTHORIZED_ACTION_EVENT = 'UnauthorizedAction'
+
 export const INDEXER_CRAWLING_EVENTS = {
   CRAWLING_STARTED: 'crawlingStarted',
   REINDEX_QUEUE_POP: 'popFromQueue', // this is for reindex tx, not chain
```
Maungmaunglwin096062286, line 99: paulo-ocean pushed 1 commit.

cef5ce4 more

https://github.com/user-attachments/assets/ba641b94-89a9-46f4-8ea5-e63ed31d642f

Maungmaunglwin096062286 (changes requested):

alexcos20:
why do we need to emit events?

paulo-ocean:
> why do we need to emit events?

we don't necessarily need them for the functionality, but its important to properly test access related stuff... for the tests infrastructure basically... without this its impossible to know what happened if something went wrong.. 

consider the following:
i can try to publish an asset, but a lot of things could fail, either i'm not allowed to publish it... or i cannot validate it, or i cannot decrypt it. etc... we have now multiple roles and a lot of credentials configurations involved...
 
If i try to look for that asset, and something failed, its not there, its not indexed, but how do i know what happened?... Where was the issue?..

If we emit some AUTH related events, we can catch those (if we want) and properly test the access/credentials/roles changes..

Without it, we can't.. 

That was the main idea

giurgiur99:
Should we keep the events mentioned here? I don't find them mandatory. @alexcos20

mariacarmina:
> > why do we need to emit events?
> 
> we don't necessarily need them for the functionality, but its important to properly test access related stuff... for the tests infrastructure basically... without this its impossible to know what happened if something went wrong..
> 
> consider the following: i can try to publish an asset, but a lot of things could fail, either i'm not allowed to publish it... or i cannot validate it, or i cannot decrypt it. or some other generic error... we have now multiple roles and a lot of credentials configurations involved as well...
> 
> If i try to look for that asset, and something failed, its not there, its not indexed, but how do i know what happened?... Where was the issue?..
> 
> If we emit some AUTH related events, we can catch those (if we want) and properly test the access/credentials/roles changes.. we can at least distinguish if it was access related
> 
> Without it, we can't..
> 
> That was the main idea. Not necessary but useful

Regarding this comment, I think maintaining events for authorisation checks is overengineering/overcomplicated, the error should be traceable via logs as we do for decryption errors, validation errors and so on. My suggestion is to not over complicate and remove them.

giurgiur99 (approved):
LGTM now

alexcos20 (approved):
lgtm

PR #882, opened 2025-03-18
add some basic docs for c2d

alexcos20:
Changes proposed in this PR:

- add some basic docs for c2d
- 
-

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -1174,3 +1174,101 @@ Forwards request to PolicyServer (if any)
   }
 }
 ```
+
+# Compute
+
+## Compute object definitions
+
+### Dataset (`ComputeAsset` Interface)
+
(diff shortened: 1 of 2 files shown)


giurgiur99 (approved):

jamiehewitt15:
I'm pretty sure we need to update `docs/C2DV2.md`. That file hasn't been updated in a while, and it doesn't mention free C2d.

alexcos20:
> I'm pretty sure we need to update `docs/C2DV2.md`. That file hasn't been updated in a while, and it doesn't mention free C2d.

will remove it for now, as it was mostly an internal document to describe the arhitecture

jamiehewitt15:
> will remove it for now, as it was mostly an internal document to describe the architecture

ok, but let's consider adding it back in at some point. It's useful to have a document that properly explains C2D, rather than just the API and envs.

PR #886, opened 2025-03-20
fix issue with empty  nft fields

### LINKED ISSUE #885: Node is Not Indexing new DDOS
Image

Related with the nft/indexedMetadataFields
when we try to create a ddo in DB it fails because nft fields are empty

paulo-ocean:
Fixes #885 

Changes proposed in this PR:

- have some nft fields data due to schema validation
- 
-

### DIFF
--- a/src/components/database/ElasticSearchDatabase.ts
+++ b/src/components/database/ElasticSearchDatabase.ts
@@ -588,7 +588,8 @@ export class ElasticsearchDdoDatabase extends AbstractDdoDatabase {
     }
     try {
       // avoid issue with nft fields, due to schema
-      ddo.nft = ddo?.indexedMetadata?.nft
+      // ddo.nft = ddo?.indexedMetadata?.nft
+      if (ddo?.indexedMetadata?.nft) delete ddo.nft
       const validation = await this.validateDDO(ddo)
       if (validation === true) {
         const response = await this.client.index({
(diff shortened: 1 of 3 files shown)


paulo-ocean:
still need to check the tests

src/components/database/ElasticSearchDatabase.ts
```diff
@@ -587,6 +587,8 @@ export class ElasticsearchDdoDatabase extends AbstractDdoDatabase {
       throw new Error(`Schema for version ${ddo.version} not found`)
     }
     try {
+      // avoid issue with nft fields, due to schema
+      ddo.nft = ddo?.indexedMetadata?.nft
```
mariacarmina, line 591: I think this looks like a hot fix, I am not sure if we need to enforce the old format structure of the ddo into the abstract database. I've seen from the issue that it complains about the nft created datetime format. I am not saying it is wrong, but maybe we can check in other place to see where the error appears instead of enforcing old format ddo.

mariacarmina: Can we purpose to have a change within the schemas directly?

paulo-ocean: Well, the fields are still present (but empty), so i'm **not enforcing anything.**
The point is that the schemas validates the nft fields and we can't go ahead with empty fields.
It could be a hot fix, but at least it fixes the current state, which is a regression :-).
IMO, We could address the schemas here [https://github.com/oceanprotocol/ddo.js/pull/11](https://github.com/oceanprotocol/ddo.js/pull/11)

paulo-ocean: fixed

mariacarmina: > don't know all the details, BUT if the '`DDO.nft`' is not needed anymore (and everything should be under `DDO.indexedMetadata.nft`) , we can probably just delete it? Don't think its about the schema

yes, that was my concern all the nft data should be now under indexedMetadata

paulo-ocean:
> lgtm now, I suppose it indexes the new DDOs with this fix.

yes, it does :-)

bogdanfazakas (approved):

PR #895, opened 2025-03-24
fix docker-compose

### LINKED ISSUE #714: Docker script:  add persistent volume
Ocean-node is using sqlite to store various internal data.
We need a volume option added to docker-compose, in order to preserve that data between restarts.

### LINKED ISSUE #894: C2D Free: config
Right now, in main, there is no config for C2D, unless the config script is executed. But, since we have > 500.000 nodes online, it's expected that a simple update will also activate c2d.

It would be nice to have an 'update' or 'migrate' script , along with scripts/ocean-node-quickstart.sh
- If we release a new version of ocean-node, they will restart their instances, but there is no C2D
- If they want C2D, they need to run scripts/ocean-node-quickstart.sh  and make sure they enter the correct private key & existing settings

To do:
 - create default options
 - proper configs (as socket is not exposed)

alexcos20:
Fixes #714

Changes proposed in this PR:

- expose docker socket to node through docker-compose.yml
- add persistent volume for /database folder
-

### DIFF
--- a/scripts/ocean-node-quickstart.sh
+++ b/scripts/ocean-node-quickstart.sh
@@ -215,6 +215,9 @@ services:
 
     networks:
       - ocean_network
+    volumes:
+      - node-sqlite:/databases
+      - /var/run/docker.sock:/var/run/docker.sock
     depends_on:
       - typesense
 
@@ -231,6 +234,8 @@ services:
 
 volumes:
   typesense-data:
+    driver: local
+  node-sqlite:
     driver: local
 
 networks:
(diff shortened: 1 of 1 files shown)


mariacarmina (approved):
lgtm

alexcos20:
Previous commit was mounting /databases as volume, but our db was actually in /usr/src/app/databases , so it was wrong.
Fixed in c0889be3afb94efd99481cd0aeddbb1df71c2e36

md00ux (approved):

PR #898, opened 2025-03-28
Import v0.0.6 of ddo.js.

### LINKED ISSUE #856: Integrate DDO library to Ocean node

mariacarmina:
Fixes #856  .

Changes proposed in this PR:

- Integrate ddo-js in ocean-node


src/components/Indexer/processor.ts
```diff
@@ -769,6 +769,7 @@ export class MetadataStateEventProcessor extends BaseEventProcessor {
           // in case their state changes back to active.
           const shortDdoInstance = DDOManager.getDDOClass({
             id: ddo.id,
+            version: ddo.version,
```
mariacarmina, line 772: DDO version here will be a constant, `short`.

src/components/database/ElasticSearchDatabase.ts
```diff
@@ -489,15 +491,17 @@ export class ElasticsearchDdoDatabase extends AbstractDdoDatabase {
...
       // Skipping validation for short DDOs as it currently doesn't work
       // TODO: DDO validation needs to be updated to consider the fields required by the schema
       // See github issue: https://github.com/oceanprotocol/ocean-node/issues/256
       return true
-    } else if ('nft' in ddo && ddo.nft?.state !== 0) {
+    } else if ('nft' in ddoInstance.getDDOData() && nft?.state !== 0) {
       return true
     } else {
-      const validation = await validateObject(ddo, ddo.chainId, ddo.nftAddress)
+      const validation = await ddoInstance.validate()
```
mariacarmina, line 504: I suggest to have a method which retrieves the DDO instance based on DDO manager and validate the record, we can create an util function in utils folder for validate DDO handler since we use this code in multiple places from code base:
```javascript
const ddoInstance = DDOManager.getDDOClass(task.ddo)
const validation = await ddoInstance.validate()

```

src/components/policyServer/index.ts
```diff
@@ -38,7 +38,7 @@ export class PolicyServer {
   }
 
   async checknewDDO(
-    rawDDO: DDO,
+    rawDDO: Record<string, any>,
```
mariacarmina, line 41: Maybe we can use the types from ddo.js here instead of Record.

mariacarmina (commented):
Some cosmetic adjustments from my side, need to test this PR too.
Thanks!

src/test/integration/database.test.ts
```diff
@@ -63,6 +63,7 @@ describe('DdoDatabase CRUD', () => {
 
   before(async () => {
     database = await new Database(typesenseConfig)
+    await database.ddo.deleteAllAssetsFromChain(ddoWithInvalidDid.chainId)
```
mariacarmina, line 66: If this line gets removed, what tests impact? In the past, cleanup of ddos was not necessary for CRUD operations tests, that's why I'm asking.

src/test/integration/indexer.test.ts
```diff
@@ -593,19 +593,20 @@ describe('Indexer stores a new metadata events and orders.', () => {
   })
 
   it('Deprecated asset should have a short version of ddo', async function () {
+    this.timeout(DEFAULT_TEST_TIMEOUT * 3)
```
mariacarmina, line 596: `DEFAULT_TEST_TIMEOUT` is too short for this particular test execution? I suggest to not increase the timeouts if the test can be executed without extra time, because it will impact the CI testing time. If we remove this, it fails time to time in the pipeline?

src/components/Indexer/utils.ts
```diff
@@ -659,9 +671,10 @@ export async function getPricingStatsForDddo(ddo: any, signer: Signer): Promise<
                 contract: dispenser,
                 token: service.datatokenAddress
               })
-              ddo.indexedMetadata.stats.push({
+              stats.push({
                 datatokenAddress: service.datatokenAddress,
                 name: await datatoken.name(),
+                symbol: await datatoken.symbol(),
```
mariacarmina, line 677: Good catch! 👍🏻

mariacarmina (commented):
Much better! 💯 I let the final round of review, after we can merge this!

src/test/integration/indexer.test.ts
giurgiur99: Good catch, this was changed while debugging, I will fix it.

src/test/integration/database.test.ts
mariacarmina: Previously for clear state, we used to stop indexer threads in `after()` but the db was not cleared up before.

> This ensures an empty state each time for that particular chainId.
Do we need this, if the tests are dedicated only for ganache chain? I suggest to remove this and check if the file before has `indexer.stopThreads()` implemented. If the CI fails, maybe the problemis raised from somewhere else.

src/@types/commands.ts
```diff
@@ -78,7 +77,7 @@ export interface FindDDOCommand extends DDOCommand {
 // this one gets the raw ddo
 // https://github.com/oceanprotocol/ocean-node/issues/47
 export interface ValidateDDOCommand extends Command {
-  ddo: DDO
+  ddo: Record<string, any>
```
bogdanfazakas, line 80: why cant we use here the DDO from  ? 
import { DDO } from '@oceanprotocol/ddo-js'

src/components/core/handler/downloadHandler.ts
```diff
@@ -194,7 +193,7 @@ export async function handleDownloadUrlCommand(
 
 export function validateFilesStructure(
   ddo: DDO,
-  service: Service,
+  service: any,
```
bogdanfazakas, line 196: is there a reason to remove the typing here ?
cant we use the service type exported by the ddo.js lib ?

bogdanfazakas (approved):
looks good, nice job 👍 
there are some small question that you can take a look at

src/@types/commands.ts
giurgiur99: Good catch 👍

src/components/core/handler/downloadHandler.ts
giurgiur99: Good catch 👍

PR #910, opened 2025-04-15
Improve build dashboard process

### LINKED ISSUE #909: Dependabot cannot update dasboard deps on it's own
**Describe the bug**
Dependabot cannot update dasboard deps on it's own since we use the hashing technique. This is complex and redundant, we can rely on the ci to check that the build has been done and not have extra checks and tests.

giurgiur99:
Fixes #909 

Changes proposed in this PR:
- Remove hashing technique for dashboard build
- Use simpler checks in ci for dashboard build


.github/workflows/ci.yml
```diff
@@ -309,5 +315,11 @@ jobs:
...
-      - name: Run dashboard build test
-        run: npm run test:dashboard
+      - name: Build dashboard
+        run: npm run build:dashboard
+      - name: Upload dashboard build
+        uses: actions/upload-artifact@v4
+        with:
+          name: dashboard-build
+          path: dist/dashboard
+          retention-days: 1
```
mariacarmina, line 325: When an automatically build is fetched, the previous dashboard build gets overwritten, right? Can you specify why we need to define a retention period in this case? Does it affects the runner resources if the dashboard build persists more than one day?

mariacarmina (commented):
One question from my side, otherwise looks pretty smooth!

giurgiur99:
> The only issue with this pr is that you need to build the dashboard manually if you done any changes into that, this was a process that was done automatically with the check changes script before.
> 
> Inmo we have 2 options:
> 
> 1. keep the build process as it was and only update the ci to test the dashboard build rather that testing the script mentioned above
> 2. move fwd with the logic from this PR and let users know that they to manually build the node dashboard if they do any change by updating the README as well

I would go with the second point since `dasboard/dist` is already commited in the repo. What do you think?

mariacarmina:
My suggestion is to also keep these changes from this PR for dashboard build, since it has a difference, when updating a package, the pipeline won't fail in this case. For breaking changes, manual flow for building the dashboard will be required.
The only observation was to not upload artifacts from dashboard if nothing was modified within the dashboard in the flow, which was fixed previously, thanks @giurgiur99 !

To summarize, next steps will be:
- to document the observation above into the README for dashboard build
- keep Bogdan's suggestion to rename the node dahsboard into control panel

After we can merge this PR.
Thanks all!

alexcos20 (approved):
lgtm

mariacarmina (approved):
LGTM

PR #918, opened 2025-04-17
Fix/ Use safe headers for Arweave storage type

bogdanfazakas:
Fixes # .

Changes proposed in this PR:

- delete 'content-length' header if 'Transfer-Encoding' header is set since having both present at same time is not allowed by the http protocol


giurgiur99 (approved):

mariacarmina (approved):
LGTM

PR #929, opened 2025-04-28
use Ocean as default token for fees

alexcos20:
Changes proposed in this PR:

- use Ocean as default token for fees, if no token is defined
- more checks on fees validation
-


bogdanfazakas (approved):

mariacarmina (approved):
LGTM!

PR #934, opened 2025-05-01
Fix consistency for job id generation within codebase.

mariacarmina:
Fixes # .

Changes proposed in this PR:

- 
- 
-


giurgiur99 (commented):
Don't we need to add the jobId in FreeComputeStartHandler too? And we can make it non-optional.

mariacarmina:
> Don't we need to add the jobId in FreeComputeStartHandler too? And we can make it non-optional.

We can generate the jobId within free start compute handler and remove the generation from `engine.startComputeJob`

giurgiur99 (approved):
LGTM!

bogdanfazakas (approved):

PR #947, opened 2025-05-28
Check wallet signature

### LINKED ISSUE #946: Client signature is not validated/used
**Describe the bug**
Clients should be able to sign for their own messages when publishing an asset. Implementation for this feature has been started in ocean.js but no support at the moment in node.

We should allow the users to sign for their assets and use that signature when publishing an asset

giurgiur99:
Fixes #946 

Changes proposed in this PR:

- If users provides a message signature, check it
- keep backward compatibility

### DIFF
--- a/src/components/core/handler/ddoHandler.ts
+++ b/src/components/core/handler/ddoHandler.ts
@@ -877,14 +877,19 @@ export function validateDdoSignedByPublisher(
   signature: string,
   publisherAddress: string
 ): boolean {
-  const message = ddo.id + nonce
-  const messageHash = ethers.solidityPackedKeccak256(
-    ['bytes'],
-    [ethers.hexlify(ethers.toUtf8Bytes(message))]
-  )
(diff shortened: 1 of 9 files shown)


src/components/core/handler/ddoHandler.ts
```diff
@@ -817,6 +809,22 @@ export class ValidateDDOHandler extends CommandHandler {
       const ddoInstance = DDOManager.getDDOClass(task.ddo)
       const validation = await ddoInstance.validate()
 
+      const { ddo, publisherAddress, nonce, signature: signatureFromRequest } = task
+      if (publisherAddress && nonce && signatureFromRequest) {
```
alexcos20, line 813: address, nonce, signature should be mandatory.

giurgiur99: Added configurable behaviour using VALIDATE_UNSIGNED_DDO env var in last commit.

mariacarmina (approved):
LGTM!

PR #954, opened 2025-06-06
chore: return schema

AdriGeorge:
Issue:
nft is not in ddoInstance.getDDOFields() becouse now is in indexedMetadata, now the schema is always op_short so the indexer will store into that index instead of the index op_ddo_v${ddo.version}

Fixes # .

Changes proposed in this PR:

- getting nft from ddoInstance.getDDOData().indexedMetadata
- returning correct schema
- indexer store in correct index


mariacarmina (commented):
One small observation, otherwise looks good and can be merged. 

Thank you!

PR #960, opened 2025-06-12
fix: update indexedMetadata event

AdriGeorge:
Fixes # .

Changes proposed in this PR:

- update event with updateFields (is inside indexedMetadata
image

- indexedMetadata.event.txid


mariacarmina (approved):
LGTM!
Nice catch

Issue #980, opened 2025-07-16
Cleanup stuck c2d jobs


mariacarmina:
From https://github.com/oceanprotocol/ocean-node/pull/971#issuecomment-3068874432:
_1. Cleanup jobs_
If the algorithm job did not finished its execution (did not reach "Publishing Results" status), after quiting node, container and volumes assigned to the job are still up.
If job container and volumes are manually deleted and start again the node, node will fetch the stuck jobs from sqlite and it stops its execution because no resources are assigned for that job (again if the docker container and volumes are manually deleted).
When node is restarted, it should check for stuck jobs if they have assigned container and volumes, if they do not exist -> delete the job from sqlite. For testing purposes, I deleted manually the sqlite before starting again the node.

**Flow Description:**
   - Check whether the containers and volumes still exist for that job.
   - If they no longer exist → check if job is in Running phase, Configuring Volumes - resume job execution
   - PublishingResults phase -> delete the job from the SQLite database.
   - If they still exist → resume the execution of the stuck job.
  
**Fix**:
Handle exception thrown in try-catch block which prevent ocean-node execution to crash and make sure the flow description from above is respected.

PR #989, opened 2025-07-19
Handle stuck jobs on Running phase. Fix scheduling cleanup for c2d jobs orphans

### LINKED ISSUE #988: Scheduler for expired c2d jobs cleanup throws error when invoked
Scheduler for expired c2d jobs cleanup throws error when invoked at every 5 mins (the interval set in the code by default):

It is triggered in cleanup orphans jobs

mariacarmina:
Fixes #988 #980  .

Changes proposed in this PR:

- fix node crash error when unexisting container was trying to be fetched by c2d monitor loop without deleting sqlite before -> in "RunningAlgorithm" phase, added try-catch and set job status back to ConfiguringVolumes step to re-create the container and resume the job lifecycle
- fixed: if the algorithm job gets stuck in "PublishingResults" phase -> delete the job from sqlite
- surround with try-catch scheduler operations of deleting old logs & deleting expired and orphans c2d jobs
- fixed cleanup orphans jobs:  `OCEANNODE:	Error deleting expired jobs: Cannot read properties of undefined (reading 'id')` by passing environment as parameter to `getFinishedJobs()`
Screenshot 2025-07-19 at 13 56 27

### DIFF
--- a/src/components/c2d/compute_engine_docker.ts
+++ b/src/components/c2d/compute_engine_docker.ts
@@ -1,5 +1,6 @@
 /* eslint-disable security/detect-non-literal-fs-filename */
 import { Readable } from 'stream'
+import os from 'os'
 import {
   C2DStatusNumber,
   C2DStatusText,
@@ -175,7 +176,8 @@ export class C2DEngineDocker extends C2DEngine {
       type: 'cpu',
       total: sysinfo.NCPU,
       max: sysinfo.NCPU,
-      min: 1
+      min: 1,
+      description: os.cpus()[0].model
     })
     this.envs[0].resources.push({
       id: 'ram',
--- a/src/components/c2d/compute_engine_docker.ts
+++ b/src/components/c2d/compute_engine_docker.ts
@@ -866,13 +866,31 @@ export class C2DEngineDocker extends C2DEngine {
       }
     }
     if (job.status === C2DStatusNumber.RunningAlgorithm) {
-      const container = await this.docker.getContainer(job.jobId + '-algoritm')
-      const details = await container.inspect()
-      console.log('Container inspect')
-      console.log(details)
+      let container
+      let details
+      try {
+        container = await this.docker.getContainer(job.jobId + '-algoritm')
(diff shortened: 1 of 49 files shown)


src/components/database/C2DDatabase.ts
```diff
@@ -117,15 +117,19 @@ export class C2DDatabase extends AbstractDatabase {
...
-    const finishedOrExpired: DBComputeJob[] = await this.provider.getFinishedJobs()
-    const envIds: string[] = existingEnvironments.map((env) => {
-      return env.id
-    })
     let cleaned = 0
-    for (const job of finishedOrExpired) {
-      if (job.environment && !envIds.includes(job.environment)) {
-        if (await c2dDatabase.deleteJob(job.jobId)) {
-          cleaned++
+    const envIds: string[] = existingEnvironments
```
alexcos20, line 121: this is not working as it should

alexcos20 (changes requested):

src/components/database/C2DDatabase.ts
mariacarmina: What do you mean exactly?
The reason why I changed the existing approach to iterate through computeEnvironments is to pass as parameter the environment to this call (which previously threw error `Error deleting expired jobs: Cannot read properties of undefined (reading 'id')`): await this.provider.getFinishedJobs(computeEnvironment).
It runs as expected the cron job at 5 mins when tested locally.

src/components/c2d/compute_engine_docker.ts
```diff
@@ -976,8 +976,15 @@ export class C2DEngineDocker extends C2DEngine {
...
+        try {
+          const algoLogFile =
+            this.getC2DConfig().tempFolder + '/' + job.jobId + '/data/logs/algorithm.log'
+          writeFileSync(algoLogFile, String(e.message))
+        } catch (e) {
+          console.log('Failed to write')
+          console.log(e)
+        }
         await this.db.updateJob(job)
-        await this.cleanupJob(job)
```
alexcos20: why was this removed?

alexcos20 (approved):
lgtm

PR #1009, opened 2025-07-29
Allow dataset on c2d

giurgiur99:
Changes proposed in this PR:

- Datasets should also be allowed on compute jobs

### DIFF
--- a/src/components/core/compute/utils.ts
+++ b/src/components/core/compute/utils.ts
@@ -80,8 +80,11 @@ export async function validateAlgoForDataset(
       throw new Error('Dataset service not found')
     }
     const { compute } = datasetService
-    if (datasetService.type !== 'compute' || !compute) {
-      throw new Error('Service not compute')
+    if (datasetService.type !== 'compute' && datasetService.type !== 'dataset') {
+      throw new Error('Service is not compute or dataset')
+    }
+    if (!compute) {
+      throw new Error('Compute service not found')
     }
     const publishers = compute.publisherTrustedAlgorithmPublishers || []
     const algorithms = compute.publisherTrustedAlgorithms || []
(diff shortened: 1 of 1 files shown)


src/components/core/compute/utils.ts
```diff
@@ -80,8 +80,11 @@ export async function validateAlgoForDataset(
       throw new Error('Dataset service not found')
     }
     const { compute } = datasetService
-    if (datasetService.type !== 'compute' || !compute) {
-      throw new Error('Service not compute')
+    if (datasetService.type !== 'compute' && datasetService.type !== 'dataset') {
+      throw new Error('Service is not compute or dataset')
+    }
+    if (!compute) {
```
alexcos20, line 86: this will fail for datasets, as they don't have compute options

src/components/core/compute/utils.ts
```diff
@@ -80,8 +80,11 @@ export async function validateAlgoForDataset(
       throw new Error('Dataset service not found')
     }
     const { compute } = datasetService
-    if (datasetService.type !== 'compute' || !compute) {
-      throw new Error('Service not compute')
+    if (datasetService.type !== 'compute' && datasetService.type !== 'dataset') {
```
mariacarmina, line 83: 'Dataset' represents asset type, it is not included within `compute` settings and it does not represent a service type. Supported service types are `access` for download/consume flow and `compute` for starting c2d jobs.

src/components/core/compute/utils.ts
```diff
@@ -79,6 +79,9 @@ export async function validateAlgoForDataset(
     if (!datasetService) {
       throw new Error('Dataset service not found')
     }
+    if (datasetService.type === 'access') {
```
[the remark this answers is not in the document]
mariacarmina: We discussed today to keep this check for access service type, since access service can be used also for download and compute flows.

mariacarmina (approved):
LGTM

PR #1034, opened 2025-09-01
Parse JSON config file.

### LINKED ISSUE #1020: Node JSON config file
As Node features continue to grow,  so is number of envs needed.

Proposal:
- have a file (config.json) that holds entire config, and keep only PRIVATE_KEY as env

At startup, we should validate the config, and exit on invalid config

mariacarmina:
Fixes #1020  .

Changes proposed in this PR:

- implement function for loading json file as config if CONFIG_PATH variable is specified
- unit tests for `getConfiguration()` function
- added new .env test file for config updated docs.


giurgiur99 (commented):
We can add a config.example.json, similar to .env.example

src/test/.env.test2
```diff
@@ -0,0 +1,2 @@
+PRIVATE_KEY='0xc594c6e5def4bab63ac29eed19a134c130388f74f019bc74b8f4389df2837a58'
+CONFIG_PATH="$HOME/config.json"
```
giurgiur99, line 2: $HOME/config.json is not the root of the folder

src/utils/config.ts
```diff
@@ -861,17 +861,105 @@ export function loadConfigFromEnv(envVar: string = 'CONFIG_PATH'): OceanNodeConf
...
+      connectionsMaxParallelDials: parseInt(process.env.P2P_connectionsMaxParallelDials),
+      connectionsDialTimeout: parseInt(process.env.P2P_connectionsDialTimeout),
+      upnp: process.env.P2P_ENABLE_UPNP,
+      autoNat: process.env.P2P_ENABLE_AUTONAT,
+      enableCircuitRelayServer: process.env.P2P_ENABLE_CIRCUIT_RELAY_SERVER,
+      enableCircuitRelayClient: process.env.P2P_ENABLE_CIRCUIT_RELAY_CLIENT,
+      circuitRelays: parseInt(process.env.P2P_CIRCUIT_RELAYS),
+      announcePrivateIp: process.env.P2P_ANNOUNCE_PRIVATE,
+      filterAnnouncedAddresses: JSON.parse(process.env.P2P_FILTER_ANNOUNCED_ADDRESSES),
+      minConnections: parseInt(process.env.P2P_MIN_CONNECTIONS),
```
alexcos20, line 936: what happens if there is no env called **P2P_MIN_CONNECTIONS** ?

PR #1047, opened 2025-09-28
Update job model.

### LINKED ISSUE #1043: Update job model to include container stats

mariacarmina:
Fixes #1043 .

Changes proposed in this PR:

- add OOMKilled and exit codes attributes to c2d job model

### DIFF
--- a/src/@types/C2D/C2D.ts
+++ b/src/@types/C2D/C2D.ts
@@ -146,7 +146,7 @@ export type DBComputeJobMetadata = {
   [key: string]: string | number | boolean
 }
 
-export interface ComputeJobError {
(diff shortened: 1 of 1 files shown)


src/@types/C2D/C2D.ts
```diff
@@ -160,6 +165,7 @@ export interface ComputeJob {
   agreementId?: string
   environment?: string
   metadata?: DBComputeJobMetadata
+  error?: ComputeJobError
```
alexcos20, line 168: I would name it terminationDetails.

cause we might have OOMKilled..  or we might have an algo that finished fine, and we have an exit code (OOMKilled is false)

PR #1074, opened 2025-10-20
set job duration and return all jobs on getJobs handler

dnsi0:
Fixes # .

Changes proposed in this PR:

- add algorithm duration
- return all jobs instead of finished ones on getJobs handler

image


ndrpp (approved):

giurgiur99 (approved):
LGTM

PR #1124, opened 2025-12-16
fix: update admin handlers validation to accept smart wallet signatures

ndrpp:
Fixes # .

Changes proposed in this PR:

- add support for smart wallet signatures validation in admin handlers (smart wallets from alchemy are [going to be used in the nodes dashboard](https://github.com/oceanprotocol/nodes-dashboard/pull/219))


ndrpp:
> use a specific logger (like CORE), or create a new one, but do not use console
> 
> also, use verbose -> errors level, depending on the criticality

They were for testing purposes only, will revert them afterwards or use the logger as you've mentioned in case they are needed :+1:

PR #1127, opened 2025-12-19
Fix/p2p get push node config

ndrpp:
Fixes # .

Changes proposed in this PR:

- pass address body param in http handlers for erc1271 admin validation
- return an error when pushing config to a node that has running jobs

### DIFF
--- a/src/components/core/admin/pushConfigHandler.ts
+++ b/src/components/core/admin/pushConfigHandler.ts
@@ -51,20 +51,6 @@ export class PushConfigHandler extends AdminCommandHandler {
     }
 
     try {
-      const hasRunningJobs =
-        (await this.getOceanNode().getDatabase().c2d.getRunningJobs()).length > 0
-      if (hasRunningJobs) {
-        return new Promise<P2PCommandResponse>((resolve) => {
-          resolve({
-            status: {
-              httpStatus: 400,
-              error: 'Error pushing config: Node has running jobs'
-            },
-            stream: null
-          })
-        })
-      }
-
       const configPath = getConfigFilePath()
       const configContent = await fs.promises.readFile(configPath, 'utf-8')
       const currentConfig = JSON.parse(configContent)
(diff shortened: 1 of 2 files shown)


alexcos20 (commented):

Overall risk: low

Summary:
This pull request introduces the ability to dynamically update the OceanNode's configuration via an admin P2P command. It adds a `setConfig` method to the `OceanNode` class and integrates this into the `PushConfigHandler`. A crucial safety measure has been added to prevent configuration updates if there are active C2D jobs running. Additionally, the `address` parameter is now passed from HTTP handlers to the P2P admin command handlers for enhanced validation, and the private key is properly hidden in the configuration response. The C2D engines are also re-added after a configuration update to ensure the new settings take effect.

Comments:
• [INFO][other] The `setConfig` method re-initializes the `escrow` instance. This is generally safe if the `escrow` object is designed to be fully recreated from the configuration. Could you confirm that re-initializing it this way, rather than updating its internal properties, is the intended and correct behavior, and doesn't lead to resource leaks or unexpected state changes if it held complex state or external connections?
• [INFO][other] The check for running jobs before updating the configuration is an excellent safety addition. Consider adding a more detailed log message when a configuration update is rejected due to running jobs, indicating which jobs are active (e.g., their IDs or status) to aid in debugging.
• [INFO][other] Calling `this.getOceanNode().addC2DEngines()` after a config update is critical for the changes to take effect. It's assumed this method handles the lifecycle gracefully, e.g., stopping existing engines if their configuration has changed or is no longer valid, and starting new ones. A comment here clarifying this behavior or confirming that `addC2DEngines` is idempotent and handles re-initialization properly would be beneficial.

src/components/core/admin/pushConfigHandler.ts
```diff
@@ -51,6 +51,20 @@ export class PushConfigHandler extends AdminCommandHandler {
     }
 
     try {
+      const hasRunningJobs =
+        (await this.getOceanNode().getDatabase().c2d.getRunningJobs()).length > 0
```
alexcos20: why do we have this condition?

alexcos20 (approved):

PR #1143, opened 2026-01-09
fix validateDDO signature logic

### LINKED ISSUE #816: ValidateDDOCommand - getValidationSignature only for authentificated requests
When calling `ValidateDDOCommand` we always add our signature if the DDO is valid, no matter who is the published. In order to implement the concept of isolated markets, we need to change that.

So, we will add our signature **_ONLY IF_**:
 - command contains parameter `publisherAddress`
 - command contains parameter `nonce`
 - command contains parameter `signature` which is the signed message based on nonce by publisherAddress
 - ddo.nftAddress exists and it's valid
 - publisherAddress  has updateMetadata role on ddo.nftAddress contract
 - publisherAddress  has publishing rights on this node (see #815)

alexcos20:
Fixes #816

Changes proposed in this PR:

- sign ddo only if config.validateUnsignedDDO is true  or if publisher is checked (nonce, signature, etc(
- 
-

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -158,8 +158,6 @@ jobs:
           RPCS: '{ "8996": {"rpc": "http://127.0.0.1:8545", "chainId": 8996, "network": "development", "chunkSize": 100}}'
           DB_URL: 'http://localhost:9200'
           DB_TYPE: 'elasticsearch'
-          DB_USERNAME: 'elastic'
-          DB_PASSWORD: 'changeme'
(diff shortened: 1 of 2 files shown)


src/test/integration/elasticsearch.test.ts
```diff
@@ -13,7 +13,9 @@ import { SQLLiteNonceDatabase } from '../../components/database/SQLLiteNonceData
 
 const dbConfig = {
   url: 'http://localhost:9200',
-  dbType: DB_TYPES.ELASTIC_SEARCH
+  dbType: DB_TYPES.ELASTIC_SEARCH,
+  username: 'elastic',
+  password: 'changeme'
```
bogdanfazakas: same as above or here maybe use the values from `.env.test` file

alexcos20: fixed in https://github.com/oceanprotocol/ocean-node/pull/1143/commits/2905d592f91d35ec61f99404a8afa9d26a259ffb

PR #1149, opened 2026-01-13
feat: update node quickstart script with gpu auto detect

ndrpp:
Fixes # .

Changes proposed in this PR:

- update node quickstart script to include UPnP setup
- automatically detect gpus and add them to docker compute environments


alexcos20:
same for AMD, you are not passing the devices

PR #1167, opened 2026-01-21
fix: refactor blockchain class

ndrpp:
Fixes # .

Changes proposed in this PR:

- refactor blockchain class to pass config object to it
- update signer caching mechanism to hash keys instead of storing them in plain text

### DIFF
--- a/src/components/Indexer/crawlerThread.ts
+++ b/src/components/Indexer/crawlerThread.ts
@@ -9,7 +9,12 @@ import { Blockchain } from '../../utils/blockchain.js'
 import { BlocksEvents, SupportedNetwork } from '../../@types/blockchain.js'
 import { LOG_LEVELS_STR } from '../../utils/logging/Logger.js'
 import { isDefined, sleep } from '../../utils/util.js'
-import { EVENTS, INDEXER_CRAWLING_EVENTS, INDEXER_MESSAGES } from '../../utils/index.js'
+import {
+  EVENTS,
+  getConfiguration,
+  INDEXER_CRAWLING_EVENTS,
+  INDEXER_MESSAGES
+} from '../../utils/index.js'
 import { INDEXER_LOGGER } from '../../utils/logging/common.js'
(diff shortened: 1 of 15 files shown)


ndrpp:
>• [ERROR][bug] The private key conversion `Buffer.from(this.config.keys.privateKey).toString('hex')` is likely incorrect if `this.config.keys.privateKey` is already a hex-encoded string (e.g., '0x...').
> 
> `ethers.Wallet` expects a 0x-prefixed hex string or a 32-byte Uint8Array directly. If `this.config.keys.privateKey` is '0xdeadbeef...', `Buffer.from('0xdeadbeef...')` will treat the '0', 'x', 'd', 'e', 'a', 'd', ... as ASCII characters, not hex digits. Converting this buffer back to hex will result in a hex string that is the ASCII representation of the input string, not the actual private key bytes.
> 
> Assuming `this.config.keys.privateKey` should be a standard 0x-prefixed hex string (like `process.env.PRIVATE_KEY` typically is), you should pass it directly to the `ethers.Wallet` constructor:
> 
> ```ts
> this.signer = new ethers.Wallet(this.config.keys.privateKey, this.provider)
> ```
> 
> If the intention is for `this.config.keys.privateKey` to be a raw byte string (which is unusual for private keys in config), please clarify and add a comment, but even then, `Buffer.from().toString('hex')` is unlikely to be the correct conversion for `ethers.Wallet`. • [ERROR][bug] Same issue as the previous comment regarding private key conversion. Please verify the expected format of `this.config.keys.privateKey` and adjust the `ethers.Wallet` instantiation accordingly. • [ERROR][bug] Same issue as the previous comments regarding private key conversion. Please verify the expected format of `this.config.keys.privateKey` and adjust the `ethers.Wallet` instantiation accordingly.

`this.config.keys.privateKey` is stored as a Uint8Array(32) when merging the config from the env  and config.json (see `src/utils/config/builder.ts`), so this is not a bug

alexcos20 (approved):

PR #1189, opened 2026-02-03
add signature for encrypt handlers

### LINKED ISSUE #762: Command: encrypt  - Add user signature
Now, the encrypt command can be called by anyone, without any type of auth.
This can lead to some DDOS attacks.

Solution:
 - add consumer & signature required parameters
 - check signature when somebody is trying to use this handler
 - update http route
 - add policyServer command, so an owner can decide who can have access

Warning:
 - this has breaking changes for ocean.js/cli, so they need to be updated

alexcos20:
Fixes #762

Changes proposed in this PR:

- require signature for `encrypt` and `encryptFiles` endpoints
- ocean.js and ocean-cli already updated
-

### DIFF
--- a/src/components/core/handler/handler.ts
+++ b/src/components/core/handler/handler.ts
@@ -182,19 +182,23 @@ export abstract class CommandHandler
   ): Promise<P2PCommandResponse> {
     const oceanNode = this.getOceanNode()
     const auth = oceanNode.getAuth()
-    if (auth) {
-      const isAuthRequestValid = await auth.validateAuthenticationOrToken({
-        token: authToken,
-        address,
-        nonce,
-        signature,
-        message
-      })
-      if (!isAuthRequestValid.valid) {
-        return {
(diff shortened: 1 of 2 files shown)


src/components/core/handler/handler.ts
```diff
@@ -182,17 +182,19 @@ export abstract class CommandHandler
...
-      address,
-      nonce,
-      signature,
-      message
-    })
-    if (!isAuthRequestValid.valid) {
-      return {
-        stream: null,
-        status: { httpStatus: 401, error: isAuthRequestValid.error }
+    if (auth) {
```
ndrpp: shouldn't we return 401 if there is no auth?

alexcos20: see https://github.com/oceanprotocol/ocean-node/pull/1189/commits/46296d900ce10ffeb20f1b96b9c757c493e94c61

PR #1239, opened 2026-03-01
Update DNS entries for bootstraps

alexcos20:
Update DNS entries for bootstraps
- 
- 
-


bogdanfazakas (approved):
lgtm, lets merge #1238 first it should fix the system tests as well

PR #1241, opened 2026-03-02
allow override cpu and ram

giurgiur99:
Fixes # .

Changes proposed in this PR:

- 
- 
-

### DIFF
--- a/src/components/c2d/compute_engine_docker.ts
+++ b/src/components/c2d/compute_engine_docker.ts
@@ -226,8 +226,16 @@ export class C2DEngineDocker extends C2DEngine {
 
     if (envConfig.resources) {
       for (const res of envConfig.resources) {
-        // allow user to add other resources
-        if (res.id !== 'cpu' && res.id !== 'ram') {
+        if (res.id === 'cpu' || res.id === 'ram') {
+          // allow config to override cpu/ram per env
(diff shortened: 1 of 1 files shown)


alexcos20 (commented):

Overall risk: low

Summary:
This pull request enhances the C2D engine's resource management by allowing CPU and RAM resource configurations (total, max, min) to be overridden via `envConfig.resources`. Previously, only non-CPU/RAM resources could be added or configured this way. The changes enable more granular control over compute resources per environment, aligning with the intent to allow custom overrides for core resources.

Comments:
• [INFO][other] The logic `else if (res.total != null) existing.max = res.total` effectively sets `existing.max` to `existing.total` if `res.max` is not explicitly provided but `res.total` is. This ensures `max` is always updated to at least `total` when `total` is present. This seems to be a deliberate design choice to infer `max` from `total`.
• [INFO][other] If `res.id` is 'cpu' or 'ram' but `existing` is `undefined` (i.e., 'cpu' or 'ram' is not found in `this.envs[0].resources`), the current logic silently skips the update. This is likely acceptable given that CPU and RAM are expected to be default resources always present. However, if there's a scenario where they might not be, this would prevent configuration. Consider if an `else` branch to add the resource if `existing` is not found, or a warning/error, would be beneficial, though probably not required for the current design.

PR #1267, opened 2026-03-17
getFile: return proper type

alexcos20:
Changes proposed in this PR:

- use StorageObject as return for getFile function
- 
-

### DIFF
--- a/src/components/core/compute/utils.ts
+++ b/src/components/core/compute/utils.ts
@@ -44,7 +44,7 @@ export async function getAlgoChecksums(
         file as StorageObject,
         true // force checksum
       )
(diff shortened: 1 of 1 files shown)


alexcos20:
> [WARNING][bug] The fileInfo.checksum ?? '' handles cases where checksum might be null or undefined. While this prevents runtime errors, please confirm if an empty string '' is an acceptable value for a content checksum in all scenarios, especially if the checksum is critical for data integrity validation. Ideally, fetchSpecificFileMetadata should either guarantee a non-empty checksum when forceChecksum is true, or propagate an error if it cannot be obtained.

Fixed in https://github.com/oceanprotocol/ocean-node/pull/1267/commits/5c5dc773ddb5a23adf7d6fb2837afbc17a4e6f11

PR #1320, opened 2026-04-08
feat(#1317): update quickstart script & networking docs

ndrpp:
Fixes # .

Changes proposed in this PR:

- set tls multiaddrs in p2p announce addresses during quickstart script run
- update networking docs with more information of fallbacks

### DIFF
--- a/docs/networking.md
+++ b/docs/networking.md
@@ -1,107 +1,147 @@
 # Ocean Node Networking
 
-## Networking in cloud environments or DMZ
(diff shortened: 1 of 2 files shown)


giurgiur99 (commented):
One small improvement that we can have:
Autocomplete user's public ip, something like 
Public ip (<x.x.x.x>): 

We can get this with `curl ifconfig.me`

dnsi0 (approved):

giurgiur99 (approved):

PR #1357, opened 2026-04-28
Fix accessing bucket files in c2d jobs

giurgiur99:
Fixes # .

Changes proposed in this PR:

- Check isFree before throwing err
- Fix error in docker: 2026-04-28T10:24:31.994Z error: CORE:	Unable to create docker container: (HTTP code 400) bad parameter - invalid mount config for type "bind": invalid mount path: 'persistent_storage/buckets/11a75ea6-8b52-418f-a935-b87bd4ae6ad2/package.json' mount path must be absolute
- 

image

### DIFF
--- a/src/components/c2d/compute_engine_docker.ts
+++ b/src/components/c2d/compute_engine_docker.ts
@@ -1252,6 +1252,7 @@ export class C2DEngineDocker extends C2DEngine {
       }
     }
     if (
+      isFree &&
       algorithm.meta.container &&
       algorithm.meta.container.dockerfile &&
       !env.free?.allowImageBuild
--- a/src/components/persistentStorage/PersistentStorageLocalFS.ts
+++ b/src/components/persistentStorage/PersistentStorageLocalFS.ts
@@ -3,6 +3,7 @@ import fsp from 'fs/promises'
 import path from 'path'
 import { pipeline } from 'stream/promises'
 import { randomUUID } from 'crypto'
+import Dockerode from 'dockerode'
 
 import type { AccessList } from '../../@types/AccessList.js'
 import type {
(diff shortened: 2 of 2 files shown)


alexcos20 (commented):

Overall risk: low

Summary:
The PR introduces a targeted fix to ensure that the Docker image build restriction (`env.free?.allowImageBuild`) is exclusively applied to free compute jobs. Previously, the logic incorrectly applied this free-tier restriction to paid jobs, causing unintended build failures.

Comments:
• [INFO][bug] Good catch! By scoping the `!env.free?.allowImageBuild` check with `isFree &&`, you've correctly isolated the free-tier limitations so they no longer restrict paid workloads. This aligns the behavior with the configuration's intended namespace (`env.free`). The logic fix is sound.

LGTM!

alexcos20 (commented):

Overall risk: low

Summary:
This PR fixes a bug related to Docker image build checks in free environments and introduces a solid Docker-in-Docker (sibling) volume mount mapping mechanism for persistent storage. The host mapping approach is well-designed but contains a bug related to missing imports and synchronous file reading.

Comments:
• [INFO][bug] Good logic fix. Ensuring the `allowImageBuild` rule only applies when the job `isFree` prevents unintended restrictions on paid compute jobs.
• [ERROR][bug] `fs.readFileSync` is used here, but `fs` is not natively imported at the top of the file (only `fs/promises` as `fsp`). This will likely throw a `ReferenceError` that gets silently swallowed by the catch block below, causing the host path detection to silently fail in all environments.

Additionally, synchronous file reads block the Node.js event loop, which is an anti-pattern inside `async` functions.

Suggested fix:
```diff
-      const containerId = fs.readFileSync('/etc/hostname', 'utf8').trim()
+      const containerId = (await fsp.readFile('/etc/hostname', 'utf8')).trim()
```
• [INFO][other] Excellent implementation of host path resolution using Docker's `Mounts`. This perfectly solves the classic "Docker sibling" bind-mount issue where in-container paths do not map to the host's actual filesystem paths.
• [INFO][style] Good job catching errors and falling back to the default `baseFolder` gracefully without crashing the node. This ensures safe behavior in non-Docker or custom deployments.

alexcos20 (approved):

PR #1394, opened 2026-06-19
add persistentStorage as Storage first class

alexcos20:
# Persistent storage in Compute (C2D)

Makes node persistent storage (localfs buckets) usable as datasets and algorithms in compute jobs, and turns `nodePersistentStorage` into a first-class storage type instead of a scattered special case.

## What changed

- **Datasets & algorithms can now use persistent storage in compute** — regardless of how the file reference arrives: plaintext, encrypted, or via a published DDO (`documentId` + `serviceId`). Previously only a plaintext file object worked, and persistent-storage algorithms were rejected outright.
- **`nodePersistentStorage` is now a first-class storage class** (alongside `url`, `ipfs`, `s3`, …). `Storage.getStorageClass()` returns a working backend for it, so algorithm provisioning and checksum computation go through the same path as every other storage type — removing the special-case branches.
- **Download is explicitly denied for persistent storage** — these files are compute-only and are never served through the download or file-info endpoints.
- **Access control is enforced early** — a consumer not on a bucket's ACL is rejected at compute *initialize/start* (for every reference form), in addition to the existing enforcement at job execution.
- **Bug fix:** a relative `persistentStorage.options.folder` produced a relative Docker bind-mount path and failed container creation (`mount path must be absolute`). The folder is now resolved to an absolute path.

## Technical notes

- A single shared `resolveComputeFileObject` helper resolves the effective (decrypted/DDO-derived) file object, reused by the compute engine and the initialize/start pre-checks.
- The persistent-storage backend gained `getReadableStream` / `getFileInfo` / `getFileChecksum`; the algorithm trusted-checksum gate uses a real content hash so trust matching still works.
- Persisted job records no longer store decrypted bucket references — they are re-resolved on demand.
- File-info gating checks both the command type and the embedded file type (case-insensitive) to prevent a routing bypass.

## Tests

Added integration/unit coverage for: encrypted and published persistent-storage datasets, persistent-storage algorithms, mixed persistent/non-persistent datasets, ACL denial, the storage class itself, absolute bind-mount paths, and download denial.

## Summary by CodeRabbit

* **New Features**
  * Added a node-local persistent storage backend with SHA-256 checksums, file metadata, and readable-stream support
  * Added consumer-aware handling throughout compute and file-info flows for persistent-storage items
  * Blocked direct persistent-storage downloads (returns HTTP 403)
* **Bug Fixes**
  * Fixed persistent-storage authorization for encrypted and referenced file objects across compute initialization/execution
* **Tests**
  * Expanded persistent-storage integration coverage (mounting, checksums, ACL gating) and strengthened unit tests
* **Chores**
  * Centralized persistent-storage file-object resolution and unified ACL validation across algorithms and datasets

### DIFF
--- a/src/@types/PersistentStorage.ts
+++ b/src/@types/PersistentStorage.ts
@@ -1,5 +1,5 @@
 import type { AccessList } from './AccessList'
-import type { BaseFileObject } from './fileObject.js'
+export type { PersistentStorageObject } from './fileObject.js'
 export type PersistentStorageType = 'localfs' | 's3'
 
 export interface PersistentStorageLocalFSOptions {
@@ -32,10 +32,4 @@ export interface DockerMountObject {
   Source: string
   Target: string
   ReadOnly: boolean
-}
-
-export interface PersistentStorageObject extends BaseFileObject {
-  type: 'nodePersistentStorage'
-  bucketId: string
-  fileName: string
 }
--- a/src/@types/commands.ts
+++ b/src/@types/commands.ts
@@ -75,6 +75,8 @@ export interface FileInfoCommand extends Command {
   fileIndex?: number
   file?: StorageObject
   checksum?: boolean
+  // required only for nodePersistentStorage files, to gate on the bucket ACL
+  consumerAddress?: string
 }
 // group these 2
 export interface DDOCommand extends Command {
--- a/src/@types/fileObject.ts
+++ b/src/@types/fileObject.ts
@@ -49,12 +49,19 @@ export interface FtpFileObject extends BaseFileObject {
   url: string
 }
 
+export interface PersistentStorageObject extends BaseFileObject {
+  type: 'nodePersistentStorage'
+  bucketId: string
+  fileName: string
+}
+
 export type StorageObject =
   | UrlFileObject
   | IpfsFileObject
   | ArweaveFileObject
   | S3FileObject
   | FtpFileObject
+  | PersistentStorageObject
 
 export interface StorageReadable {
   stream: Readable
@@ -68,7 +75,8 @@ export enum FileObjectType {
   IPFS = 'ipfs',
   ARWEAVE = 'arweave',
   S3 = 's3',
-  FTP = 'ftp'
+  FTP = 'ftp',
+  NODE_PERSISTENT_STORAGE = 'nodePersistentStorage'
 }
 
 export interface FileInfoRequest {
--- a/src/@types/fileObject.ts
+++ b/src/@types/fileObject.ts
@@ -79,6 +79,16 @@ export enum FileObjectType {
   NODE_PERSISTENT_STORAGE = 'nodePersistentStorage'
 }
 
+// Case-insensitive match for the persistent-storage type. getStorageClass routes on
+// `type?.toLowerCase()`, so all guard checks must normalize too or a casing variant
+// (e.g. "NodePersistentStorage") slips past the guard and is still routed as PS.
+export function isPersistentStorageType(type: unknown): boolean {
+  return (
+    typeof type === 'string' &&
+    type.toLowerCase() === FileObjectType.NODE_PERSISTENT_STORAGE.toLowerCase()
+  )
+}
+
 export interface FileInfoRequest {
   type: FileObjectType
   fileIndex?: number
--- a/src/components/c2d/compute_engine_docker.ts
+++ b/src/components/c2d/compute_engine_docker.ts
@@ -59,7 +59,11 @@ import { ValidateParams } from '../httpRoutes/validateCommands.js'
 import { Service } from '@oceanprotocol/ddo-js'
 import { getOceanTokenAddressForChain } from '../../utils/address.js'
 import { dockerRegistryAuth, OceanNodeConfig } from '../../@types/OceanNode.js'
-import { BaseFileObject, EncryptMethod } from '../../@types/fileObject.js'
+import {
+  BaseFileObject,
+  EncryptMethod,
+  isPersistentStorageType
+} from '../../@types/fileObject.js'
 import { getAddress, ZeroAddress } from 'ethers'
 import { AccessList } from '../../@types/AccessList.js'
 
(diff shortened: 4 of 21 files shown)


alexcos20:
@coderabbitai full review

alexcos20 (commented):

Overall risk: high

Summary:
The PR successfully integrates `nodePersistentStorage` into the main `Storage` abstraction and cleanly expands its functionality to compute algorithms. However, a critical authorization bypass vulnerability exists due to inconsistent string case validation between the ACL pre-checks and the underlying storage factory. Adding strict case-insensitive checks and defense-in-depth ACL enforcement in the new storage class is required before merging.

Comments:
• [ERROR][security] **Critical Security Vulnerability (Authorization Bypass):** `Storage.getStorageClass()` evaluates `file.type` case-insensitively, but the pre-checks here and throughout the file evaluate `type` case-sensitively.

If an attacker submits a compute job with a dataset or algorithm `type: "NODEpersistentStorage"`, it will fail this `!== 'nodePersistentStorage'` check. This circumvents the bind-mount skips and ACL validations. However, it will successfully be instantiated by `getStorageClass()` later in the file and be fully downloaded into the container without ACL enforcement (since `consumerAddress` is not passed to `getStorageClass` in the download path).

**Fix:** Use a case-insensitive comparison for ALL `file.type` checks here and on lines 2938, 2948, and 2989.
```diff
-        if (!resolved || resolved.type !== 'nodePersistentStorage') {
+        if (!resolved || resolved.type?.toLowerCase() !== 'nodepersistentstorage') {
```
• [ERROR][security] This guard is case-sensitive. An attacker or client supplying `type: "NODEpersistentStorage"` would bypass this throw, but `getStorageClass` would still instantiate the class without a `consumerAddress`. This leads to a silent logic failure where the checksum evaluates to `undefined` instead of throwing the required error.

```diff
-      if ((file as any)?.type === 'nodePersistentStorage' && !consumerAddress) {
+      if ((file as any)?.type?.toLowerCase() === 'nodepersistentstorage' && !consumerAddress) {
```
• [WARNING][security] As a defense-in-depth measure, it is highly recommended to strictly enforce the presence of `consumerAddress` when attempting to read or stat persistent storage files. Because `Storage.getStorageClass` makes `consumerAddress` optional, any internal code that accidentally calls `getReadableStream()` on this class without it will silently bypass ACL enforcement entirely.

```diff
   override async getReadableStream(): Promise<StorageReadable> {
+    if (!this.consumerAddress) {
+      throw new Error('Access Denied: consumerAddress is strictly required to read persistent storage files')
+    }
     const { bucketId, fileName } = this.getFile() as PersistentStorageObject
```
Consider adding the same strict check to `fetchSpecificFileMetadata`.
• [INFO][style] Excellent defensive programming using a case-insensitive check here! To permanently resolve the security inconsistencies mentioned in `compute_engine_docker.ts` and `utils.ts`, please export this `isPersistentStorageType` helper (e.g., to `src/@types/fileObject.ts` or a shared utility) and use it universally across the codebase (including in `PersistentStorageFactory.ts` ACL checks).
• [WARNING][security] The user-supplied `fileName` is joined directly using `path.join`. While `ensureFileExists` may perform validation, it is always a best practice to explicitly sanitize filenames to prevent Path Traversal attacks (e.g., passing `fileName: "../../etc/passwd"`).

```diff
-    const targetPath = path.join(this.bucketPath(bucketId), fileName)
+    const targetPath = path.join(this.bucketPath(bucketId), path.basename(fileName))
```
• [INFO][architecture] The `resolveComputeFileObject` abstraction is an excellent design improvement. It cleans up the file-resolution pipeline and prevents code duplication effectively.

giurgiur99 (approved):
LGTM

alexcos20:
@coderabbitai full review

alexcos20 (commented):

Overall risk: high

Summary:
The PR successfully elevates persistent storage to a first-class storage type, unifying its access through the `Storage` class and improving ACL resolutions for compute assets. However, a critical security vulnerability exists due to a combination of a case-sensitivity bypass in `downloadHandler.ts` and a 'fail-open' ACL check in `PersistentStorageLocalFS.ts`, which together allow unauthenticated downloading of any persistent storage file.

Comments:
• [ERROR][security] The check for persistent storage uses strict equality against `FileObjectType.NODE_PERSISTENT_STORAGE`. If an attacker submits a download command with `type: 'NodePersistentStorage'` (capitalized), this check will evaluate to false and the file will be passed to `Storage.getStorageClass()`. Since `getStorageClass()` performs a case-insensitive match, it will route the request to `NodePersistentStorage` without a `consumerAddress`, effectively bypassing this block and allowing unauthorized direct downloads. Use the new helper function to ensure case-insensitivity.

```diff
-    if (
-      (task.fileObject as { type?: string })?.type ===
-      FileObjectType.NODE_PERSISTENT_STORAGE
-    ) {
+    if (isPersistentStorageType((task.fileObject as { type?: string })?.type)) {
```
• [ERROR][security] This implements a 'fail-open' mechanism for file metadata. If `consumerAddress` is omitted or `undefined`, the ACL check is entirely skipped and file metadata is returned! This allows internal functions that inadvertently omit `consumerAddress` to expose file information. To ensure a secure 'fail-closed' posture, this method must throw an error if `consumerAddress` is missing.

```diff
-    if (consumerAddress) {
-      await this.assertConsumerAllowedForBucket(consumerAddress, bucketId)
-    }
+    if (!consumerAddress) {
+      throw new Error('Access denied: consumerAddress is required to access persistent storage')
+    }
+    await this.assertConsumerAllowedForBucket(consumerAddress, bucketId)
```
• [ERROR][security] Similarly, `getReadableStream` fails open when `consumerAddress` is undefined, skipping the ACL check. This is exactly what makes the bypass in `downloadHandler.ts` so critical, as `NodePersistentStorage` passes `undefined` when instantiated outside a compute job. Enforce the ACL strictly by requiring the consumer address.

```diff
-    if (consumerAddress) {
-      await this.assertConsumerAllowedForBucket(consumerAddress, bucketId)
-    }
+    if (!consumerAddress) {
+      throw new Error('Access denied: consumerAddress is required to access persistent storage')
+    }
+    await this.assertConsumerAllowedForBucket(consumerAddress, bucketId)
```
• [WARNING][security] The `/api/services/fileinfo` route is inherently unauthenticated, but it now accepts `consumerAddress` from the request body to fulfill persistent storage queries. Since the backend strictly trusts this `consumerAddress` for ACL verification without a cryptographic signature (unlike Compute commands), any unauthenticated user can provide the address of a known bucket owner and successfully enumerate metadata (size, last modified time) of arbitrary files within that bucket. Consider whether metadata leakage is acceptable here, or require a signature validation.
• [INFO][other] Excellent defensive programming. Guarding against a missing `consumerAddress` here prevents unexpected downstream failures and clearly communicates the context requirement for persistent storage.

PR #1404, opened 2026-06-30
cross-node auth tokens

giurgiur99:
Changes proposed in this PR:

- Enable cross-node auth tokens
- Ask another node for confirmation that this token is legit + recover the address of the user that signed it to prove ownership
- 

## Summary by CodeRabbit

* **New Features**
  * Auth tokens now support extra claims and can be validated across peers when not found locally.
  * Added support for a new auth-token validation command for distributed environments.
  * Successful authentication can now return the resolved consumer address.

* **Bug Fixes**
  * Improved token validation for signatures, issuer checks, and loop prevention.
  * Tightened streamed validation responses to avoid overly large payloads.

* **Tests**
  * Expanded authentication coverage for delegated validation, invalid signatures, missing claims, and local-only token lookup.


alexcos20 (commented):

Overall risk: high

Summary:
The PR introduces cross-node token verification but contains a critical security flaw. The implementation uses `jwt.decode()` to parse remote tokens, bypassing cryptographic signature verification. Since the user's embedded signature only authenticates the `address`, `nonce`, and `command`, an attacker can forge the JWT payload to alter or remove the `validUntil` field, creating tokens that never expire. Additionally, there are bypass vulnerabilities in how `validUntil` and `createdAt` are validated.

Comments:
• [ERROR][security] Using `jwt.decode()` evaluates the payload without verifying its cryptographic signature (CWE-345). Because the embedded user signature (`signatureMatchesAddress`) only covers `address`, `nonce`, and `command`, the `validUntil` and `createdAt` fields in the JWT payload are completely unauthenticated.

An attacker can take a legitimate token, extract the `address`, `nonce`, and `signature`, and generate a forged JWT payload with an altered or missing `validUntil` field. The `validateRemoteToken` method will incorrectly accept this forged token.

**Recommendations for architectural fix:**
1. **Cryptographic Trust:** If nodes share trust, verify the JWT using `jwt.verify()` with a shared secret or the issuing node's public key (JWKS).
2. **Stateless Verification (Client-Signed Expiration):** If validation must be stateless and trustless, the user's original signature MUST include the expiration time (e.g., signing `address + nonce + validUntil + command`).
3. **Nonce Freshness:** If the `nonce` is guaranteed to be a timestamp, the server must enforce a strict maximum time-to-live against it (e.g., rejecting if `Date.now() - Number(nonce) > MAX_TTL`), and completely ignore the unauthenticated `validUntil` field in the payload.
• [ERROR][security] The expiration check is flawed and can be bypassed. An attacker can omit `validUntil` from a forged JWT payload (making it `null` or `undefined`), or pass a non-numeric string (e.g., `"infinity"`) which causes `Number(validUntil)` to evaluate to `NaN`. Because `Date.now() >= NaN` evaluates to `false`, the token will be considered valid indefinitely.

Require `validUntil` to be a valid number and reject tokens that fail this check.

```diff
-    if (validUntil != null && Date.now() >= Number(validUntil)) {
+    const expTime = Number(validUntil)
+    if (validUntil == null || Number.isNaN(expTime) || Date.now() >= expTime) {
       return null
     }
```
• [WARNING][bug] Similar to `validUntil`, the `createdAt` value is pulled from an unverified payload. If it is omitted or a non-numeric string is provided, `Number(createdAt)` will be `NaN`, resulting in an invalid `Date` object being returned inside the `AuthToken`. This could lead to downstream crashes if other components expect a valid `Date` object.

Ensure `createdAt` is validated before constructing the `Date` object.

```diff
+    const createdTime = Number(createdAt)
+    if (Number.isNaN(createdTime)) {
+      return null
+    }
     return {
       token,
       address,
-      created: new Date(Number(createdAt)),
+      created: new Date(createdTime),
       validUntil: validUntil != null ? new Date(Number(validUntil)) : null,
       isValid: true
     }
```
• [INFO][style] Good use of falling back to verify both the hex string and the raw byte array with `toBeArray`. This ensures compatibility with different client-side signing implementations (e.g., clients that sign the raw bytes vs. clients that sign the literal hex representation).

giurgiur99:
@coderabbitai review

alexcos20 (commented):

Overall risk: high

Summary:
This PR implements cross-node authentication token validation. While the feature is well-structured, there are critical security vulnerabilities that need to be addressed. The primary issue is an authentication bypass in the remote token validation logic due to trusting the `issuerPeerId` from an unverified JWT. Additionally, there is an IDOR/privilege escalation flaw in the Persistent Storage handlers, as well as potential DDoS vectors via unbounded stream reading and long network timeouts.

Comments:
• [ERROR][security] Critical Authentication Bypass: `jwt.decode` is used without cryptographic verification of the issuing node. An attacker who intercepts or observes a victim's valid `CREATE_AUTH_TOKEN` signature (which is node-agnostic) can craft a forged JWT containing the victim's address and signature, but setting `issuerPeerId` to the attacker's own malicious node. Node B will successfully verify the victim's consumer signature locally, then query the attacker's node. The attacker's node will reply `{"valid": true}`, causing Node B to fully authenticate the attacker as the victim. To resolve this, the issuing node MUST sign the JWT asymmetrically (e.g., using its libp2p private key), and Node B MUST verify the token's signature against the `issuerPeerId`'s public key before making any P2P calls.
• [ERROR][security] Privilege Escalation / IDOR: By using `task.consumerAddress ? getAddress(task.consumerAddress) : ...`, you allow an authenticated user to act on behalf of another user. An attacker can supply their own valid token in `task.authorization` (which sets `isAuthRequestValid.consumerAddress` to the attacker's address) but pass a victim's address in `task.consumerAddress`. Since `task.consumerAddress` is truthy, the system will use it and grant access to the victim's storage. You should strictly use the authenticated identity:
```diff
-        ownerNormalized = task.consumerAddress
-          ? getAddress(task.consumerAddress)
-          : getAddress(isAuthRequestValid.consumerAddress)
+        ownerNormalized = getAddress(isAuthRequestValid.consumerAddress)
```
Note: This issue is also present in other PersistentStorage handlers (e.g., lines 163, 271, 315, 373, 423). Please update all occurrences.
• [WARNING][security] Unbounded Stream Reading: `streamToString` consumes the entire stream into memory. A malicious peer could exploit this by sending an endless stream of data in response to `VALIDATE_AUTH_TOKEN`, causing an Out-Of-Memory (OOM) crash on the node. You should enforce a maximum byte limit (e.g., 1024 bytes) when reading this JSON response.
• [WARNING][performance] A 30-second timeout for remote token validation (`VALIDATE_TOKEN_DIAL_TIMEOUT_MS = 30_000`) is very long. Since this validation blocks incoming API requests, it can lead to resource exhaustion if an attacker supplies a slow or unresponsive `issuerPeerId`. Consider reducing this to a much shorter timeout (e.g., 5000ms).

alexcos20 (commented):

Overall risk: medium

Summary:
This PR implements cross-node verifiable JWT authentication by embedding the original authentication signature and a `validUntil` timestamp into the token payload. The architecture allows nodes to validate tokens statelessly while maintaining strong cryptographic guarantees. The refactoring to reuse `consumerAddress` in storage handlers is also excellent. A few security hardening suggestions are provided, specifically around enforcing a max token TTL during stateless validation to prevent permanently valid tokens, and adding explicit type checking for the decoded JWT payload.

Comments:
• [WARNING][security] Cross-node tokens bypass the `MAX_AUTH_TOKEN_TTL_MS` limit that is normally enforced during token creation. A malicious dApp could prompt a user to sign a token valid for many years, construct the JWT locally, and use it indefinitely across nodes since this stateless validation lacks TTL bounds (and stateless tokens cannot be easily revoked by a single node). Consider exporting `MAX_AUTH_TOKEN_TTL_MS` from `authHandler.ts` and enforcing it here as well:
```diff
-    if (!Number.isFinite(validUntilNum) || Date.now() >= validUntilNum) {
+    if (
+      !Number.isFinite(validUntilNum) || 
+      Date.now() >= validUntilNum || 
+      validUntilNum - Date.now() > MAX_AUTH_TOKEN_TTL_MS
+    ) {
       return null
     }
```
• [INFO][security] The decoded JWT payload is cast to an object without strict type checking. Although downstream functions (like `verifyConsumerSignature`) handle non-string values safely by throwing catchable errors in `ethers`, it is best practice to explicitly validate types to prevent prototype pollution or edge-case coercion bugs.
```diff
     const { address, nonce, signature, createdAt, validUntil, chainId } = decoded as {
       address?: string
       nonce?: string
       signature?: string
       createdAt?: number
       validUntil?: number | null
       chainId?: string | null
     }
-    if (!address || !nonce || !signature || validUntil == null) {
+    if (
+      typeof address !== 'string' || 
+      typeof nonce !== 'string' || 
+      typeof signature !== 'string' || 
+      typeof validUntil !== 'number'
+    ) {
       return null
     }
```
• [INFO][security] Hardcoding `PROTOCOL_COMMANDS.CREATE_AUTH_TOKEN` during signature verification is an excellent security measure. It ensures that signatures intended for other operational commands (e.g., `UPDATE_BUCKET_LABEL`) cannot be maliciously repurposed to mint cross-node authentication tokens.
• [INFO][performance] Excellent refactoring here! Reusing the `consumerAddress` returned from `validateTokenOrSignature` completely avoids a redundant token database lookup (`getAddressFromToken`), improving overall performance and keeping the code DRY.

alexcos20 (commented):

Overall risk: low

Summary:
Implement cross-node self-verifiable authentication tokens. This PR embeds the user's original signature, `validUntil`, and `chainId` inside the JWT payload, allowing any node in the network to cryptographically verify the token without needing a shared database or shared JWT secret. Additionally, a strict 24-hour maximum TTL constraint (`MAX_AUTH_TOKEN_TTL_MS`) is introduced for token creation and cross-node validation.

Comments:
• [INFO][style] Replacing `this.getAddressFromToken(task.authorization)` with `isAuthRequestValid.consumerAddress` across all persistent storage handlers is an excellent refactor. It prevents redundant token database lookups, simplifies error handling, and safely deprecates `getAddressFromToken` entirely.
• [INFO][security] The constraint `validUntil - Date.now() > MAX_AUTH_TOKEN_TTL_MS` is a highly effective and elegant way to enforce maximum continuous validity without requiring a signed `createdAt` timestamp. This reliably caps the token's remaining lifespan at 24 hours at any given verification point. Well done!
• [INFO][architecture] **Architectural Note:** Since cross-node tokens are verified statelessly via `validateSelfContainedToken`, they inherently bypass the database nonce increment check on the verifying node. Consequently, these offline tokens cannot be explicitly revoked network-wide before they hit their `validUntil` expiry. Given the newly introduced strict 24h TTL cap, this is an optimal and standard trade-off for decentralized JWT verification. 

LGTM!

giurgiur99:
@coderabbitai review

alexcos20 (commented):

Overall risk: high

Summary:
This PR introduces cross-node authentication token validation, enabling a federated-style P2P auth mechanism. The codebase structural changes, such as injecting the P2P node into the Auth layer and refactoring `validateTokenOrSignature` to return the `consumerAddress`, are well-designed and clean. However, the implementation introduces a critical security vulnerability (Signature Replay leading to Auth Bypass) because the user's signature does not bind to the `issuerPeerId`. Additionally, there are severe performance/DoS concerns regarding network overhead and unbounded stream reading that need to be addressed before merging.

Comments:
• [ERROR][security] **CRITICAL VULNERABILITY: Signature Replay leading to Authentication Bypass**

The cross-node authentication mechanism trusts the `issuerPeerId` extracted from the unverified JWT payload. Because the consumer's signature payload (`consumer + nonce + command`) does not cryptographically commit to the `issuerPeerId`, an attacker who intercepts a valid `createAuthToken` signature (or operates a malicious node the user connects to) can craft a forged JWT. 

The attacker can set `issuerPeerId` to their own rogue node's ID. When a victim node processes this token, it verifies the consumer's ECDSA signature (which succeeds), and then queries the rogue node. The rogue node unconditionally replies `{"valid": true}`, tricking the victim node into granting the attacker full access to the user's resources.

**Recommendation:** 
To securely support cross-node tokens in a permissionless network, the consumer's signature MUST bind to the specific issuing node. You must either update the signed message to include the `issuerPeerId` (e.g., `consumer + nonce + command + issuerPeerId`), or introduce a new command (e.g., `createCrossNodeAuthToken`) to implement this securely without breaking backward compatibility for older clients.
• [WARNING][performance] **High Performance Overhead / Potential DoS**

Every HTTP/API request utilizing a cross-node authentication token will trigger a synchronous P2P network request to the issuer node. This adds significant latency to all authenticated operations and creates a multiplier effect where heavy API usage can easily overload both the validating node and the issuer node.

**Recommendation:** 
Implement a short-lived local cache (e.g., LRU cache for 60-300 seconds) for remote token validation results. This will dramatically reduce P2P traffic and latency for subsequent requests using the same token.
• [ERROR][security] **Denial of Service via Unbounded Stream Buffering**

Using `streamToString` to buffer the entire P2P response into memory before parsing makes the node vulnerable to Out-Of-Memory (OOM) attacks. A malicious node could return an infinitely long stream when queried for `VALIDATE_AUTH_TOKEN`, causing the querying node to exhaust memory and crash.

**Recommendation:** 
Enforce a strict byte limit when reading from the untrusted stream (e.g., maximum 1KB, since the expected JSON response is extremely small). If the limit is exceeded, abort the stream and reject the token.
• [INFO][style] Excellent refactoring here. Returning the validated `address` directly from `validateAuthenticationOrToken` significantly simplifies downstream handlers (like Storage handlers) and eliminates the need to redundantly decode tokens later.
• [INFO][logic] If a client passes an empty string for `task.validUntil`, `requested` evaluates to `0`. `Math.min(0, ...)` will return `0`, creating an instantly expired token. Consider ensuring that the requested expiration is explicitly in the future to avoid silent logic traps.

```diff
-      const validUntil = Number.isFinite(requested)
+      const validUntil = Number.isFinite(requested) && requested > createdAt
         ? Math.min(requested, createdAt + MAX_AUTH_TOKEN_TTL_MS)
         : createdAt + MAX_AUTH_TOKEN_TTL_MS
```

alexcos20 (commented):

Overall risk: low

Summary:
Excellent implementation of cross-node authentication token validation. The design is secure, elegantly prevents P2P spam by locally verifying the signature before issuing network requests, and properly bounds incoming stream sizes. The refactoring of `validateTokenOrSignature` to return the `consumerAddress` also cleans up the PersistentStorage handlers perfectly. LGTM!

Comments:
• [INFO][security] Great use of a bounded stream reader (`readBounded`) here to protect against potential memory exhaustion attacks from malicious or misbehaving peers.
• [INFO][style] Using a lazy getter `() => OceanP2P | undefined` for the P2P node is a very smart way to avoid circular dependency issues during the initialization of the `Auth` component.
• [INFO][security] Excellent security measure to verify the inner EIP-191/ERC-1271 signature locally before sending a P2P request. This prevents malicious actors from spamming the P2P network with forged JWTs, since they must possess a valid signature for the given address and nonce in order to trigger the remote query.
• [INFO][bug] Just to be absolutely safe, ensure that `p2p.sendTo` does not throw an exception (e.g., if the peer is unreachable or the network drops). If it can throw, wrapping this call in a `try/catch` would prevent unhandled promise rejections. If the P2P layer already catches network errors and returns a status object instead, this is perfectly fine as is.
• [INFO][style] Returning `consumerAddress` directly from `validateTokenOrSignature` is a clean refactor that eliminates the need for redundant database lookups (`getAddressFromToken`) in the PersistentStorage handlers. Good job!

alexcos20 (commented):

Overall risk: high

Summary:
The PR introduces cross-node authentication token validation by leveraging the P2P network, allowing a token issued by one node to be securely validated by another. The design is mostly solid, utilizing bounded streams to prevent memory exhaustion and enforcing maximum TTLs for tokens. However, there is a critical security vulnerability regarding issuer spoofing due to the lack of cryptographic binding between the consumer's signature and the issuing node. Additionally, there is a performance/DoS risk related to ERC-1271 validation on arbitrary remote tokens.

Comments:
• [ERROR][security] Critical Authentication Bypass via Issuer Spoofing: Because the consumer's signature (`address + nonce + command`) does not cryptographically bind the request to a specific `issuerPeerId`, an attacker can intercept a legitimate user's `CREATE_AUTH_TOKEN` signature and reuse it to construct a forged JWT.

The attacker can set the `issuerPeerId` in this forged JWT to a malicious P2P node they control. When the victim node receives this token, `verifyConsumerSignature` will pass (since the signature legitimately belongs to the consumer), and the victim node will query the attacker's node to validate the token. The attacker's node simply returns `{"valid": true}`, granting the attacker full access to the consumer's resources.

**Recommendation:** 
1. **Ideal Fix:** Update the consumer's signature payload to include the `issuerPeerId` (e.g., `consumer + nonce + command + issuerPeerId`).
2. **Alternative:** If client signatures cannot be changed, cross-node tokens must be signed asymmetrically by the issuer's P2P private key, and the receiving node must maintain a strict whitelist of trusted `issuerPeerId`s allowed to perform cross-node delegations. Otherwise, anyone can spin up a node and validate forged tokens.
• [WARNING][performance] Potential RPC Denial of Service: When validating an unknown remote token, `verifyConsumerSignature` is called prior to the P2P request. If the provided signature fails standard ECDSA verification (which an attacker can easily forge), this falls back to `isERC1271Valid`, triggering blocking RPC calls to the blockchain. An attacker can DoS the node's RPC provider by spamming invalid cross-node tokens. Consider implementing rate-limiting based on IP/PeerID or negative caching for failed signature validations before invoking RPC calls.
• [INFO][security] Excellent use of a bounded stream reader (`readBounded`) with a strict `MAX_VERDICT_BYTES` limit. This securely prevents memory exhaustion (OOM) attacks if a remote node maliciously returns an infinite or excessively large stream.
• [INFO][security] Good practice ensuring that the `validUntil` value is strictly bounded by `MAX_AUTH_TOKEN_TTL_MS`. This effectively prevents malicious clients from requesting non-expiring tokens and limits the lifespan of any compromised sessions.

giurgiur99:
@coderabbitai review

alexcos20 (commented):

Overall risk: low

Summary:
This PR successfully introduces cross-node auth token functionality with excellent architectural design, successfully preventing infinite validation loops. The refactoring of `validateTokenOrSignature` significantly simplifies the PersistentStorage handlers. Security boundaries such as `MAX_VERDICT_BYTES` and `peerIdFromString` are well-implemented to prevent buffer overflow and SSRF/injection attacks. LGTM!

Comments:
• [WARNING][security] While `address`, `nonce`, `signature`, and `issuerPeerId` are explicitly checked to be strings, `chainId` is not. Although an invalid type will be safely caught by the `try/catch` inside `verifyConsumerSignature`, validating it explicitly prevents potential object injection edge cases and fails fast.

```diff
-    if (
-      typeof address !== 'string' ||
-      typeof nonce !== 'string' ||
-      typeof signature !== 'string' ||
-      typeof issuerPeerId !== 'string'
-    ) {
+    if (
+      typeof address !== 'string' ||
+      typeof nonce !== 'string' ||
+      typeof signature !== 'string' ||
+      typeof issuerPeerId !== 'string' ||
+      (chainId != null && typeof chainId !== 'string')
+    ) {
```
• [INFO][performance] If `p2p.sendTo` does not have a strict internal timeout, an attacker could attempt a Denial-of-Service (DoS) by sending multiple token validation requests with unresponsive or blackholed `issuerPeerId`s. Ensure that libp2p stream creation and remote validation enforce a timeout at the network layer.
• [INFO][architecture] Excellent design choice using `auth.getLocalToken(task.token)` here instead of calling the general `validateToken`. This effectively prevents infinite loops across nodes when resolving tokens recursively.
• [INFO][security] Note on Cross-Node Bearer Tokens: Because JWTs act as bearer tokens and do not currently have an audience (`aud`) claim tying them to a specific node, any node that legitimately receives a valid token can theoretically replay it to impersonate the user on other nodes. If nodes within the network are completely trustless, you may want to consider audience restriction in a future PR. However, if this cross-impersonation behavior is intended as a single-sign-on network feature, it is acceptable.

PR #1411, opened 2026-07-09
fix hardcoded event

alexcos20:
# Fix: don't assume the metadata event is the first log in the transaction

### Summary

`DecryptDdoHandler.decryptDDO()` resolved the encrypted document from a `transactionId` by
parsing **`receipt.logs[0]`** and requiring it to be `MetadataCreated`/`MetadataUpdated`.
That assumption only holds for plain EOA transactions. For transactions sent through
**account-abstraction wallets (ERC-4337), multisigs (e.g. Safe) or relayers**, other events
(EntryPoint/wallet/token events) precede the metadata event, so decryption failed with
`Decrypt DDO: Failed to process transaction id` even though the transaction contains a
perfectly valid metadata event.

### What changed

- New helper `findMetadataEventInLogs(logs, dataNftAddress)` in
  `src/components/core/handler/ddoHandler.ts`:
  - scans **all** receipt logs instead of only `logs[0]`,
  - only considers logs **emitted by the data NFT contract itself** (case-insensitive
    address match) — so batched AA transactions that touch multiple NFTs resolve the event
    of the correct contract,
  - reuses the existing `fetchEventFromTransaction()` util from `src/utils/util.ts` for the
    actual log parsing/matching (no shared-util changes needed; logs are pre-filtered by
    NFT address before delegating),
  - returns the parsed `MetadataCreated` or `MetadataUpdated` event, or `null`.
- `decryptDDO()` now uses the helper; also guards against a `null` receipt. Arg extraction
  (`args[3]`/`args[4]`/`args[5]` = flags / encrypted document / document hash) and the
  error contract (HTTP 400 `Failed to process transaction id`, specific reason logged) are
  unchanged.

Note: `getEventFromTx()` from `src/utils/util.ts` could **not** be reused here — it matches
on `log.fragment.name`, which is only populated on receipts obtained from an ethers
`Contract` (`tx.wait()`). `provider.getTransactionReceipt()` returns plain logs without
`fragment`, so it would never match.

### Audit: other hard-coded event positions in the codebase

The rest of the codebase was audited for the same class of bug:

- `src/components/Indexer/ChainIndexer.ts` (`receipt.logs[reindexTask.eventIndex]`) —
  numeric index, but `eventIndex` is optional, never set by its only producer
  (`reindexTxHandler`), and falls back to scanning all logs when undefined. Intentional
  targeted-reindex feature; left as is.
- Everything else is already position-independent: `getEventFromTx` /
  `fetchEventFromTransaction` filter all logs; `BaseProcessor.getEventData` and the main
  indexer loop match `topics[0]` event-signature hashes across all logs; remaining
  `topics[N]` usages are standard indexed-argument access.

### Tests

- New unit suite `src/test/unit/ddoEventLog.test.ts` (6 tests) using synthetic receipts
  built with `Interface.encodeEventLog`:
  - finds `MetadataCreated` when it is **not** the first log (the AA/multisig scenario),
  - finds `MetadataUpdated`,
  - matches the NFT address case-insensitively,
  - ignores metadata events emitted by **other** contracts in the same transaction,
  - returns `null` when no metadata event exists / on an empty logs array.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of metadata events in transaction logs, making decrypt-related flows more reliable.
  * Better handles transactions with multiple logs, different event types, or events emitted from matching contract addresses.
  * Added clearer failures when transaction receipts are missing logs or no relevant metadata event is found.

### DIFF
--- a/src/components/core/handler/ddoHandler.ts
+++ b/src/components/core/handler/ddoHandler.ts
@@ -12,7 +12,12 @@ import {
 } from '../utils/findDdoHandler.js'
 import { toString as uint8ArrayToString } from 'uint8arrays/to-string'
 import { GENERIC_EMOJIS, LOG_LEVELS_STR } from '../../../utils/logging/Logger.js'
-import { sleep, readStream, streamToUint8Array } from '../../../utils/util.js'
+import {
+  sleep,
+  readStream,
+  streamToUint8Array,
+  fetchEventFromTransaction
+} from '../../../utils/util.js'
 import { CORE_LOGGER } from '../../../utils/logging/common.js'
 import { ethers, isAddress } from 'ethers'
 import ERC721Template from '@oceanprotocol/contracts/artifacts/contracts/templates/ERC721Template.sol/ERC721Template.json' with { type: 'json' }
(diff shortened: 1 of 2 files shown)


alexcos20 (commented):

Overall risk: low

Summary:
This PR fixes a bug where the handler blindly assumed the `MetadataCreated` or `MetadataUpdated` event would always be the first log in a transaction receipt. By searching through all logs and filtering by the Data NFT address, the code now correctly handles transactions originating from smart wallets (ERC-4337, multisigs) or other complex contracts that emit multiple events. The implementation is robust and well-tested. LGTM!

Comments:
• [INFO][performance] It's a minor optimization, but you could define `const abiInterface = new ethers.Interface(ERC721Template.abi)` outside of the `findMetadataEventInLogs` function (e.g., at the module level). This avoids re-instantiating the `ethers.Interface` on every call, which parses the ABI JSON every time. Given that this function is not called in a tight loop, the performance impact is negligible, but it's a good practice.
• [INFO][style] Since you added a check for `!receipt`, the error message `'receipt logs 0'` might be slightly misleading if the receipt itself is null (which can happen if the transaction hasn't been mined or doesn't exist on the provider). Consider updating it to something more accurate.

```diff
-            throw new Error('receipt logs 0')
+            throw new Error('Transaction receipt not found or has no logs')
```
• [INFO][other] Excellent unit tests! You clearly mocked the foreign log scenarios (such as those from ERC-4337 entry points) and thoroughly tested case-insensitivity and other edge cases. Great job on the test coverage here.

dnsi0 (approved):

PR #1416, opened 2026-07-14
fix network recreate

### LINKED ISSUE #1415: Services restart:  Failed to relaunch inference service: Error: P2P command error: (HTTP code 409) unexpected - network with name ocean-svc-66a61f88cf370d19485c0f65899e1e9ac7c70aa696fb1e31a95ee0bab2b0f1ec already exists

alexcos20:
Closing #1415 
# Fix: service restart wedged by leaked Docker network (409 "network already exists")

Restarting a Service-on-Demand job could fail permanently with
`(HTTP code 409) unexpected - network with name ocean-svc-<serviceId> already exists`,
leaving the job stuck in `Error` (99) with empty `containerId`/`networkId` — and every
subsequent restart failing the same way, with no way out short of manual `docker network rm`
on the host.

## Root cause

Each service gets a Docker network with a **deterministic name** (`ocean-svc-<serviceId>`),
but every cleanup path (restart teardown, stop, orphan recovery, expiry sweep) removed it
only via the **persisted `job.networkId`** field. That field is written only *after* the
container starts successfully, so a node crash (or error) between `createNetwork` and that
persist leaked the network with `networkId=''` in the DB. From then on, restart skipped
teardown (empty id), hit the leftover named network at `createNetwork` → Docker 409 → job
marked `Error` — a self-perpetuating dead end. Leaked networks were also never reaped by the
expiry sweep (same gate), permanently consuming the Docker daemon's limited IPAM subnet pool.

## What changed

All in `src/components/c2d/compute_engine_docker.ts`:

- **`removeServiceNetwork(serviceId, networkId?)`** (new): removes the service network by
  **both** the persisted id (when set) and the deterministic name, so cleanup never depends
  solely on the DB field. If the network still has stale attached containers (crash after
  `container.start()` but before the persist), they are force-removed first — otherwise
  `network.remove()` fails with "has active endpoints". Benign Docker errors (404/304 =
  already gone) count as success; anything else propagates.
- **`createServiceNetwork(serviceId)`** (new): wraps `createNetwork`; on a 409 name conflict
  it removes the stale network and retries once. This makes already-broken jobs (like the
  reported one) **self-heal on the next restart** with no manual intervention or DB surgery.
  The stale network is removed and recreated rather than reused: reuse saves nothing (a
  leaked network can still hold a stale container bound to the service's host ports, which
  must be inspected and force-removed either way), a fresh network always reflects the
  current code's configuration instead of inheriting options from whatever node version
  leaked it, and remove+recreate matches restart's tear-down-and-rebuild semantics — the
  container is never reused either.
- All four cleanup paths now go through the helper: `restartService` pre-teardown,
  `stopService`, `processServiceStart` orphan recovery, and `cleanupServiceDocker`
  (failure-path cleanup — now works even when `createNetwork` itself threw and no live
  handle exists). Both `createNetwork` call sites (initial start + restart) use the
  self-healing wrapper.
- Since the expiry sweep funnels `Running`/`Error` jobs through `stopService`, leaked
  networks from abandoned jobs are now reaped automatically too — no schema or sweep changes.
- `isBenignDockerError` hoisted from `stopService` to module scope (shared, unchanged).

No handler/API changes: `Error`-status jobs were already restartable; the bug was entirely
engine-side. `markServiceFailed` still deliberately leaves Docker resources in place for
restart to reclaim — reclaiming is just reliable now.

## Tests

- **Unit** (`src/test/unit/service/serviceNetworkCleanup.test.ts`, no Docker daemon):
  removal by deterministic name with empty `networkId` (the exact leak scenario), stale
  attached containers force-removed before network removal, benign 404/304 vs propagated
  errors, 409 → remove + retry-once (and no retry on other errors), and `stopService`
  end-to-end with a leaked network ending `Stopped`.
- **Integration** (`src/test/integration/services.test.ts`): new case simulates the
  post-crash DB state (`containerId=''`, `networkId=''`, `status=Error` while the real
  container + network keep running) and asserts `SERVICE_RESTART` succeeds (pre-fix: 500
  with the 409 message), the service returns to `Running` on the same host port, the stale
  container is gone, and exactly one `ocean-svc-<serviceId>` network remains. The stop test
  now also asserts the network is gone by name.
- Docs: `docs/services.md` notes the self-healing teardown semantics.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved service restart and stop reliability after interrupted starts or node crashes.
  * Automatically removes stale containers and Docker networks that could block service recovery.
  * Added recovery for network-name conflicts during service startup.
  * Ensured failed container starts correctly report an error status.

* **Documentation**
  * Clarified service network cleanup and restart behavior.

* **Tests**
  * Added coverage for stale network recovery, cleanup, and restart scenarios.

### DIFF
--- a/docs/services.md
+++ b/docs/services.md
@@ -68,6 +68,22 @@ cpu/ram/gpu allocation and keeps its host ports held — until it is restarted, 
 stopped, or swept by the same expiry check once `expiresAt` passes (which then fully releases
 everything, same as a normal expiry).
 
+**Restart is self-healing with respect to leftover Docker state.** Each service gets a Docker
+network with the deterministic name `ocean-svc-<serviceId>`. Teardown (restart, stop, expiry
+sweep) removes that network by name — not just by the stored network id — force-removing any
+stale attached container first, so state leaked by a node crash mid-start cannot wedge the
+service. If network creation still hits a name conflict, the stale network is removed and
+creation is retried once.
+
+A leftover network is deliberately **removed and recreated rather than reused**. Reusing it
+would save nothing: a leaked network can still have a stale container attached (crashed after
+`container.start()` but before the job record was persisted), still bound to the service's
+host ports — so the old container must be inspected and force-removed either way, and at that
+point recreating the now-empty network is a single cheap API call. Recreating also guarantees
+the network always reflects the current code's configuration instead of silently inheriting
+whatever options a previous node version created it with, and it matches restart's overall
+tear-down-and-rebuild semantics (the container is never reused either).
+
 ## Configuration
 
 Service-on-demand is configured per Docker connection under `serviceOnDemand`:
--- a/src/components/c2d/compute_engine_docker.ts
+++ b/src/components/c2d/compute_engine_docker.ts
@@ -85,6 +85,10 @@ const C2D_CONTAINER_GID = 1000
 
 const trivyImage = 'aquasec/trivy:0.69.3' // Use pinned versions for safety
 
+// "Already gone" Docker errors (404 missing, 304 already stopped) are the desired end
+// state of a teardown, so treat them as success.
+const isBenignDockerError = (e: any) => e?.statusCode === 404 || e?.statusCode === 304
+
 export class C2DEngineDocker extends C2DEngine {
   private envs: ComputeEnvironment[] = []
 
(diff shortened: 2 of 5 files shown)


alexcos20 (commented):

Overall risk: low

Summary:
This PR effectively addresses the Docker network leak issue caused by node crashes mid-start. The self-healing implementation properly removes network instances by both their deterministic name and stored ID, ensuring robust cleanup. Code structure, error handling, and test coverage are excellent.

Comments:
• [INFO][style] Extracting `isBenignDockerError` to a higher scope is a great refactoring move. It avoids re-defining it inside `stopService` and ensures consistent error checking across different methods like `removeServiceNetwork`.
• [INFO][other] Excellent handling of the edge case where a container might be removed by an external process between `network.inspect()` and the `.remove({ force: true })` call. Leveraging `isBenignDockerError` guarantees teardown stability.
• [INFO][bug] The self-healing logic here effectively catches the `409` conflict and properly recreates the network without halting the process. Well implemented.
• [INFO][other] Great addition of unit tests to mock and assert network operations. The testing of both leak scenarios and non-benign error propagation provides solid assurance for this fix. LGTM!

alexcos20:
@coderabbitai review

alexcos20:
@coderabbitai full review

dnsi0 (approved):

PR #1422, opened 2026-07-15
services updates

### LINKED ISSUE #1415: Services restart:  Failed to relaunch inference service: Error: P2P command error: (HTTP code 409) unexpected - network with name ocean-svc-66a61f88cf370d19485c0f65899e1e9ac7c70aa696fb1e31a95ee0bab2b0f1ec already exists

### LINKED ISSUE #1420: Services:  stopped jobs does not show Expired
Fetched service :
```
 {
        "serviceId": "f1bf16ed06459ec02cabf26bff6eeb220bdf652a0b0f7c088949b53d396b684f",
        "clusterHash": "0xff1004b67de08fc505fbf0a2089010d0f23015338c7def8557697513c4a39935",
        "environment": "0xff1004b67de08fc505fbf0a2089010d0f23015338c7def8557697513c4a39935-0x89f0e2931d4416b91ca85d231c8066906725695de6ba7715804033446015de0d",
        "owner": "0x191126c1C5531b4C50544089991c0aB12D4Fe5cF",
        "image": "vllm/vllm-openai",
        "tag": "latest",
        "dockerCmd": [
            "--model",
            "Qwen/Qwen3-8B",
            "--host",
            "0.0.0.0",
            "--port",
            "8000",
            "--max-model-len",
            "7168",
            "--gpu-memory-utilization",
            "0.2",
            "--served-model-name",
            "Qwen3-8B",
            "--dtype",
            "bfloat16"
        ],
        "containerImage": "vllm/vllm-openai:latest",
        "containerId": "79353facd57b28c8ba1a1058eea43d0a79d743c24c6b17b9fe709b5fccba0254",
        "networkId": "e9904ac810e8ab99610a5580c78a89b36bdf8e3ae0a7c0ae77e79fb30a26e76a",
        "status": 70,
        "statusText": "Stopped",
        "dateCreated": "2026-07-09T12:06:48.314Z",
        "expiresAt": 1783602408314,
        "duration": 3600,
        "exposedPorts": [
            8000
        ],
        "endpoints": [
            {
                "containerPort": 8000,
                "hostPort": 31182,
                "url": "http://test1.oncompute.ai:31182"
            }
        ],
        "resources": [
            {
                "id": "gpu",
                "amount": 1
            },
            {
                "id": "cpu",
                "amount": 1
            },
            {
                "id": "disk",
                "amount": 20
            },
            {
                "id": "ram",
                "amount": 14
            }
        ],
        "payment": {
            "chainId": 8453,
            "token": "0x5494711392a67DA50D3bC7b1fcC2d1877cFaA4d2",
            "lockTx": "0xff78379300ddad18eae19cfa67798c55f301b96618800698633223ebe3d20817",
            "claimTx": "0xd97fef7d4c8520517ebf31e17d26d4b626f0cc1faed0adf4dc21ea8f3e4a55b8",
            "cancelTx": "",
            "cost": 1.1261999999999999
        }
```

### LINKED ISSUE #1421: Services: restart sometimes gives p2p timeout error
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

### LINKED ISSUE #1426: Services & jobs:  have admin endpoint to check what resources are used by who

**Additional context**

alexcos20:
# Fix: service restart/stop race with the InternalLoop (follow-up to #1416)

## Problem

Service restarts still failed on live nodes after #1416, always with:

```
restartService <id> failed: (HTTP code 404) no such container -
failed to set up container networking: network ocean-svc-<id> not found
```

Live-node logs showed the smoking gun — the loop's orphan recovery ran **for the very
service being restarted**, seconds before the failure:

```
16:30:11.984  Performing P2P task: serviceRestart a6aba923…
16:30:16.200  ERROR processServiceStart: orphaned service a6aba923… in state "PullImage" — cleaning up
16:30:18.375  ERROR restartService a6aba923… failed: … network ocean-svc-a6aba923… not found
```

## Root cause

A race between `restartService()` and the engine's `InternalLoop` (2 s tick):

1. `restartService()` runs synchronously from the handler and persists intermediate
   statuses (`Starting` → `PullImage`/`BuildImage`) while it pulls the image.
2. Every tick, the loop fetches all jobs in pending-start statuses
   (`getPendingServiceStarts`) and launches `processServiceStart()` for any serviceId not
   in the in-memory guard set — which only ever covered **loop-launched** starts.
   `restartService()` never registered itself.
3. The loop picked up the mid-restart job; status `PullImage` ≠ `Starting` sent it into
   the **orphan-recovery branch** (crash cleanup), which removes the service network by
   its deterministic name `ocean-svc-<serviceId>`, releases the host ports and flips the
   job to `Error`.
4. The restart then created its container against the now-deleted network →
   `container.start()` → 404 "network not found".

The race predates #1416, but was silent (clobbered status / double-released ports).
#1416's removal-by-name made it deterministic: any restart whose image pull spans a
2-second tick lost the race — i.e. essentially every restart.

`stopService()` had the same family of races in both directions (a stop tearing down an
in-flight start/restart's fresh network; a restart racing an in-flight stop or the expiry
sweep).

## Fix

`src/components/c2d/compute_engine_docker.ts`:

- Generalized the loop's guard set into a **per-service lifecycle lock**
  (`servicesBeingStarted` → `serviceOpsInFlight`): at most one lifecycle operation — the
  loop-driven start pipeline, `restartService`, or `stopService` — runs per service.
- `restartService()` and `stopService()` acquire the lock at entry
  (`acquireServiceLifecycleLock`) and release it in `finally`. If the lock is held they
  throw `Service <id> has a start/stop/restart operation in progress — retry shortly`;
  both handlers already map throws to an HTTP error, so concurrent/duplicate requests get
  a clear rejection instead of corrupting each other.
- The InternalLoop keeps its exact skip-if-locked behavior for starts; its **expiry
  sweep** no longer marks a service `Expired` when the stop was deferred by the lock
  (that would leak the container/ports) — it retries on the next tick instead.
- Crash recovery is unaffected: after a node dies mid-start/mid-restart, the lock set is
  empty at boot, so the loop still orphan-cleans stale `PullImage`-state jobs.
- **Engine `stop()` now drains handler-driven ops too** (review finding): the promise set
  the loop already used for its fire-and-forget starts was generalized
  (`serviceStartPromises` → `serviceOpPromises`) and `restartService`/`stopService`
  register themselves in it, so `stop()` returns only once no lifecycle op is still
  touching Docker/escrow on the shared DB.
- **Health check vs restart hardening** (review finding): `checkRunningServices` skips
  services with a lifecycle op in flight (a restart intentionally kills the container —
  not an "unexpected death"), and `markServiceFailed` bails when the job's containerId
  changed since the health snapshot (a completed restart replaced the container), so a
  healthy restarted service can't be flagged `Error` by a stale check.

`docs/services.md`: documented the "lifecycle operations are exclusive per service"
behavior and the retry semantics.

## Tests

- **New unit suite** `src/test/unit/service/serviceRestartRace.test.ts` (7 tests):
  - lock held during a restart's image pull and released on success **and** failure paths
  - concurrent restart/stop rejected without touching Docker or the DB
  - restart-during-stop rejected; stop releases the lock afterwards
  - regression: the InternalLoop skips a locked service (with a control asserting it
    *would* run orphan recovery when unlocked — the pre-fix behavior)
  - the expiry sweep defers a locked service instead of marking it Expired sans teardown
  - engine `stop()` drains an in-flight handler-driven `stopService` before returning
- **New integration test** `(l3)` in `src/test/integration/services.test.ts`: reproduces
  the live-node scenario end-to-end — delays `pullImageRef` by 5 s so real InternalLoop
  ticks land mid-restart, asserts concurrent SERVICE_STOP / SERVICE_RESTART are rejected
  meanwhile, and that the restart completes to `Running` on the same host port with the
  service answering HTTP. Fails pre-fix exactly like the live node.
- Existing engine-stub tests (`compute.test.ts`, `serviceNetworkCleanup.test.ts`) seeded
  with the new lock field (their `Object.create(prototype)` engines skip field
  initializers).

## Review findings addressed / rejected

- **Addressed**: engine `stop()` not draining direct `restartService`/`stopService` calls
  (see Fix above).
- **Rejected (twice)**: reordering `doRestartService` to persist `Starting` + cleared
  containerId/networkId *before* tearing down the old container/network. That stretches
  the crash window in which the DB holds a bare `Starting` record from milliseconds (two
  consecutive DB writes) to 10+ seconds of Docker teardown I/O. On reboot the loop treats
  `Starting` as a brand-new start and runs the full pipeline — creating **and claiming** a
  **second escrow lock** (double-charging the consumer), allocating **new host ports**
  (silently changing the consumer's endpoints), and — since the crash was mid-teardown —
  potentially leaving the **old container running as an unpaid zombie** no longer
  referenced by the job record. The current order crashes into a benign state instead
  (job still `Running` with stale ids → the pre-existing boot health check flips it to
  `Error` → the consumer re-issues restart, whose teardown handles the missing resources
  via benign 404s on the same ports/payment — no reliance on the new health-check
  guards). The ordering constraint is now documented in a comment at the teardown site.
  The reviewer's underlying concern (the health check racing a restart) is closed by the
  targeted `checkRunningServices`/`markServiceFailed` hardening instead.

## Verification

- `npm run type-check` clean, `npm run lint` green.
- All service + compute unit suites pass (205 tests).
- Integration suite (`npm run test:servicesintegration`) requires Barge — validated in CI.

## Summary by CodeRabbit

* **New Features**
  * Added **Restarting** as a first-class service status and expanded lifecycle state handling for pending/restart scenarios.
  * Introduced cross-process per-service lifecycle locking (with stale-lock stealing) to prevent overlapping start/stop/restart/expiry operations.
* **Bug Fixes**
  * Improved reservation/host-port release behavior: ports/endpoints stay reserved across stop/restart and only release after expiry or successful teardown.
  * Hardened Docker/network lifecycle and orphan/unexpected-death handling; refined restart/stop/expiry race handling.
  * Added best-effort escrow funds pre-check for start (fast “Insufficient escrow funds” failures).
* **Documentation**
  * Updated service-on-demand lifecycle docs with clarified async semantics and polling expectations.
* **Tests**
  * Added restart/stop race coverage and expanded unit/integration assertions for locking, reservations, and state transitions.


alexcos20 (commented):

Overall risk: low

Summary:
This PR effectively addresses race conditions in service lifecycle management by introducing an in-memory per-service lifecycle lock (`serviceOpsInFlight`). The implementation accurately utilizes synchronous lock acquisition (safe in Node.js event loop) and robust `try...finally` blocks to guarantee lock release. The background loop now correctly respects these locks, preventing destructive operations like orphan-recovery from tearing down networks mid-creation. The inclusion of thorough unit and integration tests is highly commendable. LGTM!

Comments:
• [INFO][style] Excellent implementation of the per-service lifecycle lock. Using a synchronous check (`has`) and insertion (`add`) ensures there are no asynchronous yields where a race condition could slip through.
• [INFO][bug] Good use of the `try...finally` pattern. This guarantees that the lock is always released, even if the underlying Docker API throws an unexpected error, avoiding a permanently locked state for the service.
• [INFO][bug] Catching the error here and continuing the loop is a great defensive practice. It ensures that an error in stopping a single locked/expired service does not abort the entire expiry sweep for other services.
• [INFO][other] Great approach to testing the race condition. Intercepting `pullImageRef` to force a targeted delay (`await sleep(5000)`) reliably reproduces the state necessary to test this async bug without introducing excessive test flakiness.

giurgiur99 (approved):

dnsi0 (approved):

alexcos20:
Commit 50af2123ce228620e0d97b4ca3570472d0f57e9c  description:
Closes #1420 and #1421


# Service lifecycle hardening: cross-process locking, async restart, paid-window reservations

### Why

The `network ocean-svc-<id> not found` restart failure reappeared on a live node already
running the lifecycle-lock fix. The in-memory `serviceOpsInFlight` lock serializes
operations **within one process**, but it cannot see a second node process sharing the same
`databases/` directory and Docker daemon (stale container during a redeploy, overlapping
process-manager restart, …): the other process's orphan-recovery deletes the network by its
deterministic name exactly between the restart's `createNetwork` and `container.start()`.
While closing that, this change also fixes the resource-reservation model (a consumer pays
for a **time window**, not for a running container), makes restart non-blocking, and adds
enough DEBUG diagnostics to reconstruct any future lifecycle failure from remote logs alone.

### 1. Cross-process lifecycle lease (`service_locks`)

New SQLite table `service_locks` (`sqliteCompute.ts`, wrapped by `C2DDatabase`). Every
exclusive lifecycle operation — the loop's start pipeline, restart, stop, expiry teardown —
now takes a DB lease alongside the in-memory lock:

- **Atomic acquisition** via a single upsert: insert a fresh row, or steal one whose
  `acquiredAt` is older than 2 minutes (crashed holder). Two processes racing it can never
  both see success.
- **Heartbeat**: a 30 s interval re-stamps every lease the instance holds, so multi-minute
  image pulls/builds are never stolen as stale. Release is holder-scoped (a steal victim
  cannot delete the new owner's row); a crashed process's rows self-expire — no manual
  cleanup ever.
- **Graceful degradation**: if the lock DB call itself fails, the engine falls back to
  in-process-only locking (the previous behavior) instead of bricking single-process nodes.
- The container health check consults the lease too (`isServiceLocked`), so another
  process's mid-restart teardown is no longer misread as "container died" → `Error`.
- A stopped engine instance (config push / engine rebuild) refuses new lifecycle work —
  its replacement may already be running against the same DB and daemon.

### 2. `SERVICE_RESTART` is now asynchronous (new `Restarting (45)` status)

The handler used to block until the restart finished — including a potentially multi-minute
image pull — long enough to time out the HTTP/P2P response. It now mirrors `SERVICE_START`:
fast validations (exists, owner, not expired, payment not refunded), persist the job as
**`Restarting (45)`**, respond immediately; teardown → re-pull/rebuild → new container run
in the background under the same lease. Clients poll `serviceStatus` and watch
`Restarting → PullImage/BuildImage → Running` (or `Error` + reason in `statusText`).

Side benefits:

- `Restarting` is part of the pending-status set (single source of truth:
  `SERVICE_START_PENDING_STATUSES` in `@types/C2D/ServiceOnDemand.ts`), so a crash
  mid-restart is orphan-recovered at boot exactly like a crash mid-start — this also
  removes the old hazard where a bare `Starting` record persisted mid-restart could be
  double-started with a **second escrow lock**.
- New guard: a service whose start payment was **refunded** (lock cancelled, never claimed)
  cannot be restarted — it was never paid for.
- **Compat note:** the restart response now returns the job in `Restarting`, not `Running`.
  Clients that treated the response as "already running" must poll `serviceStatus`.

### 3. Reservations last the whole paid window — `SERVICE_STOP` keeps them

Previously an explicit stop released the resource amounts and host ports, so another
consumer could take a stopped service's GPU/port mid-window and block the restart. Now:

- `Stopped` counts as an **active (resource-holding) status** in `getRunningServiceJobs`,
  which feeds `getUsedResources` → `checkIfResourcesAreAvailable` — a stopped service's
  cpu/ram/gpu still gate both new service starts and compute jobs.
- Stop tears down the container + network but **keeps the host ports reserved**; a restart
  resumes on the same endpoints. Boot-time port re-seeding follows the same query, so the
  reservation survives node restarts. (CPU *pinning* is still freed at stop and re-derived
  at restart — safe, because the amount gate holds the capacity.)
- `Error` keeps its ports now too (paid jobs); previously two failure paths released them.
  Only never-paid (refunded) failures free their ports immediately.

### 4. `Expired` is the **only** release point — and it's enforced

- The expiry sweep now also covers `Stopped` (previously a stopped service sat at `Stopped`
  forever past `expiresAt`, still reading as restartable). Sweep order: teardown → mark
  `Expired` → release ports; amounts stop counting because `Expired` is not active.
- The sweep **refuses to mark `Expired` while teardown failed** (job comes back
  `Error "stop failed: …"`). `Expired` is terminal and never swept again, so stamping it
  over a live container would have leaked the container + ports forever; instead the job
  stays in the expirable set and is retried every tick until Docker recovers.

### 5. Orphan-recovery staleness guard

`processServiceStart` re-reads the job from the DB **after** acquiring the lease and bails
unless it is still in a pending status. The loop's `pendingStarts` snapshot is taken before
the lock, so a row captured mid-restart could previously be processed after that restart
completed — tearing down the freshly created container/network and clobbering
`Running → Error`.

### 6. Service handlers resolve the owning engine by `clusterHash`

restart/stop/extend/getStreamableLogs picked the first engine whose (shared) DB returned
the job — wrong on nodes with several Docker engines, where the first engine's lock and
loop don't protect another engine's job. New shared helper `findServiceJobAndEngine()`
resolves by the job's `clusterHash` and returns an explicit error when no configured engine
owns it (config drift).

### 7. DEBUG diagnostics for remote troubleshooting

- `logServiceDockerState()`: compact `docker ps -a` + all `ocean-svc-*` networks, dumped
  before/after restart and stop teardown, at orphan-recovery entry, on `createNetwork`
  409 conflicts, and **on every start/stop/restart failure right before cleanup** — a
  remote log now shows exactly what the daemon looked like when `container.start()` failed.
- Step-level DEBUG lines across the lifecycle: restart accepted (old ids + expiry), old
  container stop/remove results, per-ref network removal outcome (found / attached
  containers / removed / already gone), network + container created/started, lease
  acquired/released/held-elsewhere (with holder id), stale-snapshot skips, expiry
  "marked Expired — all resources released", INFO on completed start/restart with bound
  host ports.

### Tests

- `serviceRestartRace.test.ts` (+225 lines): async-restart contract (immediate `Restarting`,
  lock held until background op settles, `Error` persisted on failure), refunded-restart
  rejection, DB-lease conflict/steal/holder-scoped release/heartbeat, InternalLoop skipping
  leased jobs, stale-snapshot no-op, stopped-engine rejection, **the user-B scenario**
  (stop → allocator refuses the port → expiry → port free again), expiry-sweep
  teardown-failure retry, Stopped→Expired without touching docker.
- `serviceJobsDatabase.test.ts`: lease semantics on real SQLite; `Stopped` active /
  `Restarting` active + pending; `Stopped` expirable; accounting clean-slate switched to
  `Expired` (the only truly-released status).
- `compute.test.ts`: restart tests reworked to the async contract; failed-start port
  retention proven via the allocator.
- Integration `services.test.ts`: all 17 pass against Barge + real Docker, including the
  mid-restart orphan-recovery race (l3) and the full expiry window (n); stale-job cleanup
  switched to `Expired`.
- `docs/services.md` updated: async restart, reservation-for-the-paid-window contract,
  expiry as sole release point, cross-process lease semantics.

alexcos20:
@coderabbitai full review

alexcos20:
## What was fixed (insufficient-funds follow-up) in https://github.com/oceanprotocol/ocean-node/pull/1422/changes/b3f38e18614538a1702bc81e33253e9822cd558b

**1. Free-restart vulnerability (the serious one).** The restart guard in
`compute_engine_docker.ts` now requires a **claimed** payment:
`if (!job.payment?.claimTx) reject`. Every legitimately restartable job has a `claimTx`
(claiming happens before the first container start), so this closes the vector where a
no-funds start fails at `createLock` with all payment fields empty and a follow-up
`SERVICE_RESTART` would have run the container for free. Rejection message names the
cause: "payment was never claimed (unpaid or refunded) — start a new service".

**2. Unpaid resource squatting.** `getRunningServiceJobs` (`sqliteCompute.ts`) now filters
out `Error`/`Stopped` jobs without a `claimTx` — a never-paid or refunded job no longer
reserves cpu/ram/gpu or gets its ports re-seeded at boot. Mid-pipeline statuses
(`Starting`→`Claiming`) still reserve, since they're en route to payment. So the
reservation contract is now precisely: _paid_ jobs hold their resources for the whole
window; unpaid ones hold nothing.

**3. Fail-fast funds check.** `SERVICE_START` (`startService.ts`) now compares the
consumer's available escrow against the server-side cost before persisting anything,
returning `400 Insufficient escrow funds: available X, required Y wei` instead of a
`200 + serviceId` that dies silently at the Locking step. It's deliberately best-effort:
an RPC failure skips the pre-check (logged at DEBUG) and the background `createLock`
remains the authoritative gate — chain hiccups can't block starts.

New tests: never-paid restart rejection, unpaid-`Error`/refunded-`Stopped` excluded from
the active set (real SQLite), insufficient-funds start → 400 with no job record created,
and RPC-failure-doesn't-block-start. Docs updated with all three.

The end-to-end behavior for a user without funds is now: an immediate `400` in the normal
case; if the balance changes between request and lock, the job lands in `Error` with the
funds message in `statusText`, reserves nothing, and cannot be restarted into a free ride.

alexcos20:
@coderabbitai full review

alexcos20:
### Fixed (5 of 6)  in https://github.com/oceanprotocol/ocean-node/pull/1422/changes/beb981d4efde61b3b9c37c847ff6f91fef339911

1. **`releaseCpus` before teardown** — valid: cores went back to the pinning pool before
   the container was confirmed gone, so a failed teardown left the old container running
   on cores another job could now be pinned to. Moved into the successful teardown branch
   of `doStopService`, alongside the id-clearing; on `cleanupError` the allocation is
   retained.

2. **`markServiceFailed` check-then-write race** — valid: between the advisory
   `isServiceLocked` read and `updateServiceJob`, a restart could acquire the lease and
   persist `Restarting`, which the whole-body Error write would clobber. Now it
   **acquires the actual lifecycle lease** (`tryAcquireServiceLifecycleLock`), re-reads
   the job fresh under it, applies the same Running/containerId guards, writes, and
   releases in a `finally`. Fail-closed is preserved: any failure to acquire skips the
   tick and the next tick re-detects a genuinely dead container.

3. **Extend payment durability** — valid: a crash between `claimLock` and the final write
   took the consumer's money with no record, and a retry charged again. Now the extension
   **intent** (an `extendPayments` entry with `lockTx`, empty `claimTx`/`cancelTx`) is
   persisted **before** claiming; success finalizes it (`claimTx` + `expiresAt`),
   claim-failure closes it with `cancelTx`. A later extend that finds an unresolved
   intent first tries to auto-refund the stale lock (using the _intent's_ chain/token)
   and proceeds; if the refund fails (lock likely already claimed) it returns **409
   without attempting any new charge**. _Partially skipped:_ automatically _completing_ a
   claimed-but-unfinalized extension — that requires on-chain proof the claim landed; the
   durable intent makes the case auditable and operator-recoverable instead of silent.

4. **Refresh-test staleness** — valid: the lock row was milliseconds old, so the "not
   stolen" assertion held even if `refreshServiceLocks` were a no-op. The test now ages
   the original stamp 150 ms, refreshes, and has proc-B probe with a 100 ms staleness
   window — a refusal is only explainable by the re-stamp.

5. **Nitpick** — added the `releaseServiceLock(SERVICE_ID, 'test-holder')` assertion to
   the restart failure-path test.

### Skipped (1)

- **SQLite-backed host-port reservations** — skipped: the end state is already fail-safe
  without it. Each process's allocator combines the in-memory set with a live OS bind
  probe, and if two processes still pick the same port in the millisecond window during
  _simultaneous_ allocations (only possible in the already-degenerate shared-daemon
  two-process topology), Docker rejects the second bind at `container.start()` — a clean,
  retryable failed start, no corruption. A DB-backed allocator needs a schema, TTL/orphan
  cleanup, and a migration of the boot-seeding path; that's disproportionate here and
  noted as follow-up hardening alongside lease fencing.

alexcos20:
@coderabbitai full review

alexcos20:
## New command: `SERVICE_LIST` (`GetServicesHandler`) in https://github.com/oceanprotocol/ocean-node/pull/1422/changes/d6153a0a735cd83991e8dc53cb958d9e6b20f71a

`SERVICE_LIST` (`GET /api/services/serviceList`) is the node-wide service listing, shaped
like `GetJobsHandler` (`GetServicesCommand` / `GetServicesHandler`):

- **Default (no filters)**: only the services **currently holding a resource
  reservation** — exactly what the engines count against the shared pools via
  `getRunningServiceJobs`: Running/Restarting/Stopping, the mid-start pipeline states,
  paid `Error` (container died, restartable), and explicitly `Stopped` (reservation kept
  until `expiresAt`). `Expired` and never-paid jobs hold nothing and are not listed.
- **`status=<n>`**: filter to ONE specific `ServiceStatusNumber` (any status, incl.
  `Expired`); takes precedence over `includeAllStatuses`.
- **`includeAllStatuses=true`**: every status, regardless of reservations.
- **`fromTimestamp`**: only services created at/after that moment — ISO date string or a
  Unix timestamp (seconds or milliseconds).

Authenticated like every other service command (consumerAddress + signature or auth
token) but — deliberately — **not owner-scoped**: any consumer identity can see every
owner's services. The output is therefore **listing-sanitized** (`toListedServiceJob`):
on top of the always-stripped `userData` (the env blob), `dockerCmd`,
`dockerEntrypoint`, `dockerfile` and `additionalDockerFiles` are removed — identity,
status, resources, endpoints and payment metadata are kept. The per-owner full view
remains `SERVICE_GET_STATUS`.

Wiring follows the standard command pattern: `PROTOCOL_COMMANDS.SERVICE_LIST` +
supported-commands entry, `GetServicesCommand` type, `GetServicesHandler`
(`core/service/getServices.ts`), registry registration, REST route with the filter query
params. Documented in `docs/services.md` and `docs/API.md`, and added to the Postman
collection ("List Services" under Service on Demand). Covered by six unit tests
(missing-address 400, invalid status / fromTimestamp 400, default listing cross-owner +
sanitization, status filter incl. Expired, includeAllStatuses + fromTimestamp in both ISO
and Unix forms, empty list) and integration test (e2), which exercises default listing,
the status filter, a future fromTimestamp, sanitization, and unauthenticated rejection
against a live node.

PR #1423, opened 2026-07-15
Feature/use native sqlite

### LINKED ISSUE #1407: Use native node sqlite

alexcos20:
Closes #1407 

# Replace the `sqlite3` native addon with Node's built-in `node:sqlite`

Drops the external `sqlite3` npm dependency and moves every embedded SQLite database onto
Node's built-in `node:sqlite` module (`DatabaseSync` / `StatementSync`), behind one small
shared client. This is a **drop-in engine swap**: no on-disk format change, no schema
change, and no change to any public method signature of the database wrapper classes.

## Motivation

`sqlite3` is a native C++ addon (`hasInstallScript: true` → node-gyp / prebuilt-binary
download at install time) and the source of recurring pain documented in `CLAUDE.md`:
broken bindings after Node version switches (`npm_config_build_from_source=true npm rebuild
sqlite3`), GLIBC mismatches, and a native compile step in the Docker builder.

`node:sqlite` ships **inside the Node binary**: zero install scripts, zero node-gyp, no
prebuilt-binary supply chain — identical behaviour in the slim Docker runner and on dev
machines. The synchronous API also removes ~50% of the code (callback→Promise plumbing) and
fixes one latent bug (see below). SQLite usage here is a perfect fit: small, local,
single-process key-value / job-tracking tables (nonce, config, C2D jobs, auth tokens,
persistent-storage registry) — no heavy queries, no multi-process contention.

## Node version requirement (breaking for very old 22.x)

`node:sqlite` is available **unflagged since Node 22.13.0**. To fail fast instead of
crashing at runtime with `ERR_UNKNOWN_BUILTIN_MODULE: node:sqlite`:

- `.nvmrc`: `22` → `22.22.2` (matches the Dockerfile and CI exactly; any `>=22.13` works).
- `.github/workflows/ci.yml`: `v22.15.0` → `v22.22.2` (all jobs).
- `package.json`: added `"engines": { "node": ">=22.13.0" }`.

Dev (`.nvmrc`), CI, and the Dockerfile (`node:22.22.2`) now all run the same Node. No new
runtime flag is needed — the existing `--experimental-specifier-resolution=node` on `start` is
unchanged. A harmless `ExperimentalWarning: SQLite is an experimental feature` is printed on
stderr at startup (the `start` script already runs with `--trace-warnings`).

## What changed

**New shared client — `src/components/database/sqliteClient.ts`.** A thin synchronous
wrapper around `DatabaseSync` that owns one handle per file and centralizes the
sqlite3-compatibility concerns:

- `run` / `get` / `all` / `exec` mapping the old callback API to synchronous calls.
- `sanitize()` on every binding: `node:sqlite` throws on `undefined` and `boolean`
  bindings where `sqlite3` silently coerced them — so `undefined` → `NULL`, `true/false`
  → `1/0`, applied centrally.
- Eager `mkdirSync(dirname)` in the constructor: `DatabaseSync` opens eagerly and throws if
  the parent dir is missing (the old driver deferred opening), so directory creation moves
  into the client and the previously per-caller `mkdir` blocks are removed.
- `PRAGMA busy_timeout = 5000` for the long-running server process. **WAL is intentionally
  not enabled in this PR** (zero behaviour change / no new `-wal`/`-shm` sidecar files); it
  can be added later as a follow-up knob.

**Engine swap (mechanical) — same public signatures, internals become
`try { …sync… } catch (e) { log; throw }`:**

- `database/sqlite.ts` (`SQLiteProvider`: nonce + config DBs).
- `database/sqliteAuthToken.ts` (`SQLiteAuthToken`).
- `database/sqliteCompute.ts` (`SQLiteCompute`: C2D jobs, docker images, service jobs) —
  the largest file.
- `persistentStorage/PersistentStorageFactory.ts` (bucket registry; schema setup is now
  synchronous in the constructor, so the DB is ready before it returns).

**Redundant `mkdir` removed** from `SQLLiteNonceDatabase.ts`, `SQLLiteConfigDatabase.ts`,
and `AuthTokenDatabase.ts` (now handled by the client).

**Dependency + docs:** `sqlite3` removed from `package.json` / `package-lock.json`
(**−~756 lockfile lines**; `sqlite3` + its unique `node-addon-api`/`bindings`/prebuilds are
gone — `node-addon-api` remains for `keccak`/`secp256k1`). `CLAUDE.md` updated (drop the
`rebuild sqlite3` recovery note; state the ≥22.13 requirement and the `node:sqlite` backing).

### Behavioural fixes / gotchas handled

- **BLOBs come back as `Uint8Array`, not `Buffer`.** Every read-side `.toString()` on a
  blob is now `Buffer.from(blob).toString()` — `generateJSONFromBlob` and `mapServiceRows`
  in `sqliteCompute.ts`. This was the #1 silent-breakage risk; write-side is unchanged
  (`Buffer` *is* a `Uint8Array` and binds as BLOB).
- **`this.changes` → returned value.** `deleteJob` / `updateJob` / `updateServiceJob`
  (compute) and `dbDeleteBucket` / `dbUpdateBucketLabel` (persistent storage) read
  `changes` off the object returned by `run()`. Kept as a JS `number` (no `readBigInts`), so
  the existing `Promise<number>` signatures and `=== 1` comparisons are unaffected.
- **Latent bug fixed:** `sqliteAuthToken.createTable` previously did `await this.db.exec(…)`,
  but `sqlite3`'s `exec` takes a callback and returns the Database object, so the `await`
  never waited — the `CREATE TABLE` raced the `chainId` `ALTER TABLE` (masked by sqlite3's
  internal statement queue). With synchronous `exec` this is now genuinely ordered; the
  migration is wrapped in a `try/catch` that swallows only the expected "duplicate column".
- No code in `src/` matched sqlite3 error codes (`SQLITE_CONSTRAINT` etc.), so the changed
  error shape (`ERR_SQLITE_ERROR`) is a non-issue; errors still land in `DATABASE_LOGGER`.

## Tests

- **New unit** (`src/test/unit/sqliteClient.test.ts`, no daemon): eager parent-dir
  creation, `run` `changes` on hit/miss, `get` `undefined` on no-row, bind sanitization
  (`undefined`→NULL, boolean→1/0), BLOB round-trip (`Buffer` in → `Uint8Array` out →
  `Buffer.from(...).toString()`), and `INSERT … ON CONFLICT` upsert.
- **Updated** `src/test/integration/imageCleanup.test.ts`: its ~5 raw-handle probes
  (`(sqliteProvider as any).db.get/run(…, cb)`) rewritten to the synchronous `SqliteClient`
  API.
- **On-disk compatibility check (the key validation):** booted the migrated providers
  against real `databases/*.sqlite` files **written by the old `sqlite3` driver** — decoded
  5 pre-existing C2D jobs + 141 service jobs from old blobs, read nonces/config version, and
  validated a stored auth token. Confirms the BLOB-decode path and format are compatible in
  both directions (rollback = revert the PR; files stay compatible).

## Summary by CodeRabbit

* **New Features**
  * Added Service-on-Demand capabilities, including templates, asynchronous startup, status, logs, restart, stop, and duration extension.
  * Added sample CPU and GPU service templates, including LLM and embedding deployments.
  * Added cross-node authentication token validation and improved ownership handling.
  * Added support for escrow relock events.
* **Improvements**
  * Simplified compute resource configuration with shared pools and discrete-resource tracking.
  * Improved Docker shutdown and service resource allocation.
  * Migrated SQLite storage to Node’s built-in SQLite support.
* **Documentation**
  * Expanded API, GPU, compute pricing, configuration, and Service-on-Demand documentation.


alexcos20 (commented):

Overall risk: medium

Summary:
Excellent architectural improvement migrating from the native `sqlite3` dependency to Node's built-in `node:sqlite`. The abstraction in `SqliteClient` is extremely well-designed, seamlessly handling type sanitization and path initialization, and the synchronous refactoring correctly maintains backward compatibility. However, there is a critical typo in the pinned Node.js version (`22.22.2`) which does not exist and will break CI pipelines and local setups.

Comments:
• [ERROR][bug] Node.js version `v22.22.2` does not exist (the v22 release line is currently at ~v22.14.x). Pinning to a nonexistent version will cause `actions/setup-node` to fail immediately. Please update to a valid version that satisfies your `>=22.13.0` requirement (e.g., `22.13.1` or `22.14.0`).

```diff
-          node-version: 'v22.22.2'
+          node-version: '22.13.1'
```
*(Note: Ensure you apply this fix to the matrix strategy on line 45 as well)*
• [ERROR][bug] Matches the nonexistent version from CI. Please update this to a valid Node.js version so that `nvm use` does not fail for developers.

```diff
-22.22.2
+22.13.1
```
• [INFO][style] Excellent design choice here. Eagerly ensuring the directory exists (`fs.mkdirSync`) and configuring `PRAGMA busy_timeout` in the central client wrapper removes repetitive boilerplate from the individual DB classes and guarantees consistency.
• [INFO][style] Centralizing the `sanitize` logic for `node:sqlite`'s stricter bindings (gracefully mapping `undefined` and `boolean`) is a fantastic way to ensure a smooth migration without scattering type coercions throughout the consuming code.
• [INFO][other] Good catch handling the `Uint8Array` to `string` conversion here via `Buffer.from(blob)`. This correctly accounts for the change in how `node:sqlite` returns BLOBs compared to the old `sqlite3` addon.

alexcos20:
gemini is wrong:  https://nodejs.org/en/download/archive/v22.22.2

src/components/database/sqliteAuthToken.ts
```diff
@@ -33,16 +34,20 @@ export class SQLiteAuthToken implements AuthTokenDatabaseProvider {
...
-    })
+    // Migration for DBs created before the chainId column existed: add it if missing.
+    // A fresh table already has the column, so ALTER throws "duplicate column name" —
+    // that's the only expected failure, so swallow it. With the synchronous exec above this
+    // now runs strictly after the CREATE TABLE, unlike the old callback-based `await exec`
+    // which never actually waited and let the CREATE race the ALTER.
+    try {
+      this.db.exec(`ALTER TABLE authTokens ADD COLUMN chainId TEXT`)
+    } catch {
+      // column already exists
```
giurgiur99, line 45: Shouldn't we log something here?

giurgiur99 (approved):
LGTM! One small note

src/components/database/sqliteAuthToken.ts
alexcos20: well, the purpose is to add column `chainId` if it does not exists.  if it's already there, good, we are running the latest version

src/components/database/sqliteCompute.ts
```diff
@@ -84,27 +86,25 @@ export function convertStringToArray(str: string) {
...
-        else resolve(this.changes === 1)
-      })
-    })
+    const { changes } = this.db.run(deleteSQL, [jobId])
+    return changes === 1
   }
 
-  createTable() {
+  // eslint-disable-next-line require-await
+  async createTable() {
```
dnsi0, line 111: Would it make sense to add a try-catch to keep things consistent? All the other methods use one

alexcos20: see https://github.com/oceanprotocol/ocean-node/pull/1423/commits/297300aede127c123eb5a234f502a7916de75ad7

dnsi0 (approved):

PR #1424, opened 2026-07-16
remove axios

### LINKED ISSUE #1413: Use native node fetch instead of axios

alexcos20:
Closes #1413 

# Migrate from `axios` to native Node.js `fetch`

### Summary

The codebase was running **two HTTP client stacks side by side**: native `fetch` (undici,
built into Node 22) is already the convention in newer code
(`policyServer/index.ts`, `c2d/compute_engine_docker.ts`, `utils/ip.ts`, `utils/database.ts`,
`P2P/*`), while older code still used `axios`. This PR removes the direct `axios`
dependency and converges every HTTP client call in our own code onto `fetch`.

Why:

- One HTTP stack instead of two — the one that ships with the platform.
- One fewer direct dependency to track for CVEs/upgrades (axios has a steady history of
  advisories: SSRF, ReDoS, `follow-redirects`).
- undici's pooled keep-alive agent is a good fit for the hot Typesense path.

**Honest caveat:** axios does **not** leave `node_modules`. It remains a _transitive_
dependency via `@ipshipyard/libp2p-auto-tls@2.0.1 → acme-client@5.4.0 → axios@1.15.0`
(`npm ls axios` confirms this is now the only path). The win is decoupling _our_ code and
converging on one stack — not disk space or lockfile removal. If acme-client ever drops
axios upstream, it disappears for free.

**Non-goals:** touching acme-client/auto-TLS; changing any public behavior, response
shapes, or the `StorageReadable` contract; adding a wrapper library (`node-fetch`/`ky`/`got`).

### What changed

- **New `src/utils/http.ts`** — a small (~70-line, zero-dependency) helper:
  - `headersToObject(headers)` — `Headers` → plain lowercase-keyed object (the shape axios
    exposed; consumers like `downloadHandler` iterate it with `Object.entries`, which
    yields `[]` on a raw `Headers` instance).
  - `fetchHeadersTimeout(url, init, ms)` — fetch with **axios-like streaming timeout
    semantics**: the timer guards the connect + response-headers phase only, then is
    cleared once `fetch()` resolves, so long-running body streams (downloads, checksums)
    are never aborted mid-flight. This is the one genuinely subtle part of the migration.
  - `fetchStream(url, init, ms)` — adapts a response to the `StorageReadable` contract
    (a Node `Readable` via `Readable.fromWeb`, plus a plain headers object) and **throws on
    non-2xx** to preserve axios's default throw-on-error behavior that every storage caller
    relied on.

- **`src/components/database/typesenseApi.ts`** (the careful one — every metadata op flows
  through `request<T>()`). The class shape, retry loop, node rotation, and `TypesenseError`
  are unchanged; only the transport inside `request()` was rewritten. It got **simpler**:
  axios's `validateStatus: (s) => s > 0` and the manual `transformResponse` were both
  re-implementing exactly what fetch does natively. Query params via `URLSearchParams`
  (scalars coerced to strings), JSON body via `JSON.stringify` + explicit content-type,
  JSON parsing gated on `content-type`. `customError()` signature changed from
  `(AxiosResponse)` to `(status, data)`. Retry-path debug log now reads
  `error.cause?.code ?? error.code` so `ECONNREFUSED`-style codes still appear (fetch wraps
  them in `TypeError: fetch failed` with the real error on `.cause`).

- **`src/components/storage/UrlStorage.ts`**
  - `getReadableStream()` → `fetchStream(input, { method: 'GET', headers }, 30000)`.
  - `upload()` (PUT with a Node `Readable` body) → `fetchHeadersTimeout` with
    `body: Readable.toWeb(stream)` + **`duplex: 'half'`** (mandatory in undici for streaming
    request bodies). Preserves axios's throw-on-non-2xx (the C2D results-upload caller
    relies on the throw to mark the job failed). Uploads stay chunked, same as before.

- **`src/components/storage/ArweaveStorage.ts` / `IpfsStorage.ts`** — `getReadableStream()`
  becomes a one-liner over `fetchStream`.

- **`src/utils/asset.ts`** — `fetchFileMetadata()` uses `fetchHeadersTimeout` and iterates
  the web `ReadableStream` directly (`for await`); chunks are `Uint8Array`, so `.length`
  and `hash.update()` work unchanged, and the early `break` at `MAX_CHECKSUM_LENGTH`
  cancels the stream / releases the socket (same cleanup axios gave). Adds an explicit
  non-2xx throw so error pages aren't checksummed.

- **`src/components/Indexer/purgatory.ts`** — two identical GET-JSON sites → `fetch` +
  `AbortSignal.timeout(2000)`.

- **`src/components/Indexer/processors/BaseProcessor.ts`** — **two** sites (see note below):
  the nonce `GET` and the decrypt `POST`.

- **Dependency + config**
  - `package.json` — `axios` removed from `dependencies` (`package-lock.json` drops the
    direct edge; the transitive edge via acme-client stays).
  - `.eslintrc` — added `RequestInit` to `globals` (see note below).

### Reviewer notes / behavior-preservation

1. **BaseProcessor had a second axios call.** Besides the nonce `GET`, `decryptDDO()` does a
   `POST /api/services/decrypt` wrapped in `withRetrial`, with a custom
   `validateStatus` (allow 2xx/400/403, throw otherwise). fetch never throws on status, so
   the migration relies on the **pre-existing manual status checks** (`400/403 → return`,
   `not 2xx → throw → retry`) to reproduce that contract exactly. The response is adapted to
   `{ status, statusText, data }` (data parsed by content-type, matching axios). The inner
   `ECONNREFUSED` retry check was extended to also inspect `err.cause?.code`.

2. **Dead branches become live (same outcome, better message).** Under axios, purgatory's
   non-2xx and BaseProcessor's nonce failures threw into the `catch`; the `status !== 200`
   branches were unreachable. With fetch they're now reached directly. The end result
   (empty purgatory list / `Date.now()` nonce fallback) is unchanged.

3. **Nonce fallback** now guards on `data?.nonce !== undefined` rather than the old truthy
   `data` check — a small hardening over a latent `String(NaN)` path when the endpoint
   returns a body without `nonce`. Normal responses behave identically.

4. **`.eslintrc` change.** `no-undef` doesn't understand the type-only `RequestInit`
   interface (the `browser` env only supplies runtime globals like `fetch`/`Response`,
   which is why inline-literal fetch callers never tripped it). Registered `RequestInit`
   as a global — one additive line.

5. **Pre-existing bug left untouched (flagging, not fixing):** typesenseApi's retry sleep is
   `await setTimeout(this.config.retryIntervalSeconds)`, which sleeps N **milliseconds**
   despite the "seconds" name. Preserved as-is; worth a separate commit with maintainer
   sign-off.

6. **Env knob for CI/Barge:** default `config.json` points Typesense at `http://localhost:8108`.
   undici's happy-eyeballs resolves `localhost` fine, but if integration ever shows
   `ECONNREFUSED ::1:8108`, switch the config to `127.0.0.1`.

### Files

| Change                      | Files                                                                                            |
| --------------------------- | ------------------------------------------------------------------------------------------------ |
| New helper + unit test      | `src/utils/http.ts`, `src/test/unit/http.test.ts`                                                |
| Buffered JSON swaps         | `Indexer/purgatory.ts`, `Indexer/processors/BaseProcessor.ts`                                    |
| Streaming swaps             | `storage/UrlStorage.ts`, `storage/ArweaveStorage.ts`, `storage/IpfsStorage.ts`, `utils/asset.ts` |
| Careful `request()` rewrite | `database/typesenseApi.ts`                                                                       |
| Dependency + lint config    | `package.json`, `package-lock.json`, `.eslintrc`                                                 |

## Summary by CodeRabbit

* **Improvements**
  * Improved network request reliability with clearer HTTP error handling and configurable timeouts.
  * Enhanced file downloads and uploads, including streaming support for URL, IPFS, and Arweave content.
  * Improved handling of compressed responses and response headers.
  * Added safeguards for unsuccessful or empty responses and better connection cleanup.

* **Bug Fixes**
  * Fixed metadata retrieval and service requests that could mishandle response formats or interrupted connections.

* **Tests**
  * Added coverage for streaming, timeouts, headers, compressed content, and HTTP error responses.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -25,6 +25,7 @@
     "jest": true
   },
   "globals": {
-    "NodeJS": true
+    "NodeJS": true,
+    "RequestInit": true
   }
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -37,7 +37,6 @@
         "@multiformats/multiaddr": "^12.2.3",
         "@oceanprotocol/contracts": "^2.9.0",
         "@oceanprotocol/ddo-js": "^0.4.0",
-        "axios": "^1.15.0",
         "base58-js": "^2.0.0",
         "basic-ftp": "^5.3.1",
         "cors": "^2.8.5",
--- a/package.json
+++ b/package.json
@@ -76,7 +76,6 @@
     "@multiformats/multiaddr": "^12.2.3",
     "@oceanprotocol/contracts": "^2.9.0",
     "@oceanprotocol/ddo-js": "^0.4.0",
-    "axios": "^1.15.0",
     "base58-js": "^2.0.0",
     "basic-ftp": "^5.3.1",
     "cors": "^2.8.5",
--- a/src/components/Indexer/processors/BaseProcessor.ts
+++ b/src/components/Indexer/processors/BaseProcessor.ts
@@ -1,5 +1,4 @@
 import { VersionedDDO, DeprecatedDDO } from '@oceanprotocol/ddo-js'
-import axios from 'axios'
 import {
   ZeroAddress,
   Signer,
@@ -222,12 +221,13 @@ export abstract class BaseEventProcessor {
         INDEXER_LOGGER.logMessage(
           `decryptDDO: Making HTTP request for nonce. DecryptorURL: ${decryptorURL}`
         )
-        const nonceResponse = await axios.get(
+        const nonceResponse = await fetch(
           `${decryptorURL}/api/services/nonce?userAddress=${address}`,
-          { timeout: 20000 }
+          { signal: AbortSignal.timeout(20000) }
         )
-        return nonceResponse.status === 200 && nonceResponse.data
-          ? String(parseInt(nonceResponse.data.nonce) + 1)
+        const nonceData = nonceResponse.status === 200 ? await nonceResponse.json() : null
+        return nonceData?.nonce !== undefined
+          ? String(parseInt(nonceData.nonce) + 1)
           : Date.now().toString()
       } else {
         return Date.now().toString()
@@ -287,17 +287,23 @@ export abstract class BaseEventProcessor {
               nonce
             }
             try {
-              const res = await axios({
-                method: 'post',
-                url: `${decryptorURL}/api/services/decrypt`,
-                data: payload,
-                timeout: 30000,
-                validateStatus: (status) => {
-                  return (
-                    (status >= 200 && status < 300) || status === 400 || status === 403
-                  )
-                }
+              // fetch never throws on HTTP status; the manual status checks below
+              // reproduce the previous axios validateStatus contract (2xx/400/403
+              // handled here, everything else throws and is retried by withRetrial).
+              const fetchRes = await fetch(`${decryptorURL}/api/services/decrypt`, {
+                method: 'POST',
+                headers: { 'content-type': 'application/json' },
+                body: JSON.stringify(payload),
+                signal: AbortSignal.timeout(30000)
               })
+              const decryptContentType = fetchRes.headers.get('content-type') ?? ''
+              const res = {
+                status: fetchRes.status,
+                statusText: fetchRes.statusText,
+                data: decryptContentType.startsWith('application/json')
+                  ? await fetchRes.json()
+                  : await fetchRes.text()
+              }
 
               INDEXER_LOGGER.log(
                 LOG_LEVELS_STR.LEVEL_INFO,
@@ -320,9 +326,11 @@ export abstract class BaseEventProcessor {
               }
               return res
             } catch (err: any) {
-              // Retry ONLY on ECONNREFUSED
+              // Retry ONLY on ECONNREFUSED. fetch wraps the OS error in
+              // `err.cause`, so check both there and on the error itself.
               if (
                 err.code === 'ECONNREFUSED' ||
+                err.cause?.code === 'ECONNREFUSED' ||
                 (err.message && err.message.includes('ECONNREFUSED'))
               ) {
                 INDEXER_LOGGER.log(
--- a/src/components/Indexer/processors/BaseProcessor.ts
+++ b/src/components/Indexer/processors/BaseProcessor.ts
@@ -296,13 +296,22 @@ export abstract class BaseEventProcessor {
                 body: JSON.stringify(payload),
                 signal: AbortSignal.timeout(30000)
               })
-              const decryptContentType = fetchRes.headers.get('content-type') ?? ''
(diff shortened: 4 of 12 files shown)


alexcos20 (commented):

Overall risk: medium

Summary:
This is a high-quality migration from Axios to the native fetch API. The handling of Node Streams and custom timeout semantics via `AbortController` in `utils/http.ts` is exceptionally well done. However, I have identified a critical bug in `BaseProcessor.ts` where a manually checked non-2xx status does not throw an error (breaking the retry/error-handling flow), a potential double-stringification bug in Typesense batch requests, and a stream consumption edge-case in `purgatory.ts`.

Comments:
• [ERROR][bug] The comment mentions that "everything else throws", reproducing Axios's `validateStatus` behavior. However, there is no `throw` statement here. If `res.status` is 500, this block logs the error and then simply returns `res`. This bypasses the intended `catch` block (which would otherwise return `null`), potentially causing callers to misinterpret the failed response as successful.

```diff
          INDEXER_LOGGER.log(
            LOG_LEVELS_STR.LEVEL_ERROR,
            `decryptDDO: HTTP error on decrypt request. Status: ${res.status}`
          )
+         throw new Error(`Request failed with status code ${res.status}`)
        }
        return res
```
• [WARNING][bug] If `bodyParameters` is already a string (e.g., a newline-delimited JSON string for Typesense batch imports), `JSON.stringify()` will double-stringify it (turning it into a string with escaped quotes). Axios automatically detected strings and passed them raw. To prevent potential regressions, conditionally stringify the body based on its type.

```diff
-          ...(bodyParameters !== null && { body: JSON.stringify(bodyParameters) })
+          ...(bodyParameters !== null && {
+            body: typeof bodyParameters === 'string' ? bodyParameters : JSON.stringify(bodyParameters)
+          })
```
• [WARNING][bug] If `INDEXER_LOGGER.log` (with the `true` flag) does not throw an error and terminate execution, `await response.json()` on line 64 will throw a `TypeError: body used already` because the stream was already consumed by `await response.text()` inside this block. It's much safer to return early on non-200 statuses.

```diff
          true
        )
+       return
      }
```
*(Note: This applies to the Accounts Purgatory logic below as well)*
• [INFO][other] Excellent implementation of `fetchHeadersTimeout`. Preserving the Axios stream timeout semantics while migrating to `fetch` is a tricky edge case, and using `setTimeout` to manually abort the controller only during the connection/headers phase is a perfect and robust solution.
• [INFO][style] Great use of the `URL` and `URLSearchParams` APIs for building the request URL. This is much safer and cleaner than manual string interpolation and prevents potential HTTP parameter pollution.

giurgiur99 (approved):

PR #1429, opened 2026-07-23
allow docker specs on service restart

### LINKED ISSUE #1430: ServiceRestart should allow to change docker image

alexcos20:
# Allow `serviceRestart` to change the Docker image / tag

## Motivation

Service-on-Demand let a consumer start a long-running container from their **own** image, but
`serviceRestart` could only recreate the container on the *exact same* image. The realistic
recovery flow was impossible on-node:

> Start a service on your own image → discover a bug → fix it → publish the corrected image
> under a new tag → restart the service on the new tag.

Previously the only option was to stop the service and start a brand-new one, losing the paid
window, the reserved resources and the allocated host ports. This PR lets a restart carry a new
image spec while keeping payment, resources, duration and endpoints intact.

## Behaviour — restart is now atomic (all-old or all-new)

A restart never mixes new request params over the stored job. There are exactly two modes,
chosen by whether the request carries **any** container param:

| Request contains… | Mode | Container rebuilt from |
| --- | --- | --- |
| none of image/tag/checksum/dockerfile/additionalDockerFiles/userData/dockerCmd/dockerEntrypoint | **REUSE** | the stored job, unchanged |
| any of the above (⇒ `image` becomes **required**) | **RESPEC** | the request only — omitted fields are empty, never inherited |

Making `image` mandatory as soon as *any* container param is present is the discriminator that
makes a partial change impossible: you cannot ride a new `userData`/`dockerCmd` on top of the
stored image. In RESPEC mode the image spec is validated exactly like `serviceStart` (exactly one
of `tag`/`checksum`/`dockerfile`; `dockerfile` requires `allowImageBuild` on the environment).

**Always preserved:** identity (`serviceId`, owner, environment), payment/`claimTx`, resources,
duration/`expiresAt`, and the allocated host ports/endpoints. Only the container spec can change —
there is no extra charge, consistent with the pre-existing restart semantics.

This is **not** a breaking change: a restart with no container params behaves exactly as before.

## Error responses

- `400` — a container param was sent without `image` (a rejected partial change), or more than
  one of `tag`/`checksum`/`dockerfile` was supplied.
- `403` — `dockerfile` supplied but the environment has `allowImageBuild=false` (fast-fail in the
  handler; the engine re-checks as the authoritative backstop).

## Changes

**Core**
- `ServiceRestartCommand` (`src/@types/commands.ts`) — added `image`, `tag`, `checksum`,
  `dockerfile`, `additionalDockerFiles`; documented the two-mode contract.
- HTTP route (`src/components/httpRoutes/compute.ts`) — maps the new body fields.
- `ServiceRestartHandler` (`src/components/core/service/restartService.ts`) — RESPEC validation
  (image required + mutual-exclusion) and the `allowImageBuild` fast-fail gate; forwards the full
  spec to the engine.
- Engine (`compute_engine_base.ts`, `compute_engine_docker.ts`) — extended `restartService` /
  `doRestartService` signatures; `doRestartService` now computes the effective spec as a single
  atomic choice (RESPEC ⇒ request values + recomputed `containerImage`; REUSE ⇒ stored values),
  drives the pull-vs-build branch off the effective dockerfile, re-checks `allowImageBuild`, and
  persists the new image spec (`image`/`tag`/`checksum`/`dockerfile`/`additionalDockerFiles`/
  `containerImage`) onto the job so subsequent restarts and orphan-recovery see what is actually
  running. Vulnerability scanning of the new image is inherited from `pullImageRef`/
  `buildImageFromSpec`.

**Tests**
- `src/test/unit/service/serviceHandlers.test.ts` — RESPEC forwarding, REUSE forwards-undefined,
  400 for partial changes (lone tag/userData/dockerCmd), 400 for multiple image modes, 403 for a
  build-disabled dockerfile respec, and 200 for a build-enabled one.
- `src/test/unit/compute.test.ts` — engine-level: RESPEC applies + persists a new tag; RESPEC
  overrides cmd/entrypoint; REUSE reuses the whole stored spec; RESPEC clears cmd/entrypoint with
  empty arrays. (Adjusted the pre-existing tests to the new signature/semantics.)
- `src/test/integration/services.test.ts` — `(l4)` restart on a freshly-published tag applies and
  persists the new image (endpoint still reachable, same hostPort + expiresAt); `(l5)` a container
  param without `image` is rejected `400`.

**Docs**
- `docs/services.md`, `docs/API.md` (request table + 400/403 responses), and the Postman
  collection updated for the two-mode contract and the bug-fix flow.

## Summary by CodeRabbit

- **New Features**
  - Service restarts now support applying a complete new container image specification.
  - Restarts can reuse the existing specification or rebuild from supplied image settings.
  - Updated image tags can redeploy services while preserving expiration times and host ports.
  - Dockerfile-based restarts are supported where image building is enabled.

- **Bug Fixes**
  - Invalid or incomplete restart specifications are now rejected with clearer validation errors.
  - Restart updates consistently persist the newly applied container configuration.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -2084,17 +2084,46 @@ The updated `ServiceJob` (advanced `expiresAt`, new entry in `extendPayments`).
 
 #### Description
 
-Recreate the service container (no extra charge), keeping the same `expiresAt` and host ports.
-Re-checks the environment service gate and access list; rejected if the service has expired.
-Optionally pass `userData` to replace the stored env vars, and/or `dockerCmd`/`dockerEntrypoint`
-to replace the stored CMD/ENTRYPOINT overrides — each is independent: supply it (even as `[]`)
-to replace the stored value, or omit it to reuse what the service already has. This is the
-recommended recovery path after a service lands in `Error` because its container died on its own
-(the background health check leaves host ports/network/container record reserved specifically so
-restart can reuse them), in addition to recovering from an explicit `serviceStop` or any other
-terminal failure.
+Recreate the service container (no extra charge), keeping the same `expiresAt`, resources and
+host ports. Re-checks the environment service gate and access list; rejected if the service has
+expired. This is the recommended recovery path after a service lands in `Error` because its
+container died on its own (the background health check leaves host ports/network/container
+record reserved specifically so restart can reuse them), in addition to recovering from an
+explicit `serviceStop` or any other terminal failure.
+
+**Restart is atomic — either all-old or all-new, never a mix of new params over the stored job:**
+
+- **REUSE mode** — the request carries **none** of the container params below. The service
+  restarts on exactly its stored spec (image, `userData`, `dockerCmd`, `dockerEntrypoint`). Use
+  this to simply bounce a service back to `Running`.
+- **RESPEC mode** — the request carries **any** container param. The container is rebuilt
+  entirely from the request: `image` becomes **required** and exactly one of `tag`/`checksum`/
+  `dockerfile` applies (validated exactly like `serviceStart`). `userData`/`dockerCmd`/
+  `dockerEntrypoint` are taken **as-sent** — anything omitted here is empty/unset, it is **not**
+  pulled from the stored job. This is the bug-fix flow: publish a fixed image under a new tag,
+  then restart re-supplying the full spec (e.g. same `image`, new `tag`).
+
+Because `image` is mandatory whenever any container param is present, you cannot ride a new
+`userData`/`dockerCmd` on top of the stored image — a partial change is rejected (400). Payment,
+resources and duration are always preserved; only the container spec can change. A service whose
+start payment was **never claimed** (escrow lock failed or was refunded) cannot be restarted —
+start a new service instead.
 
 #### Request Body
+
+REUSE mode (bounce the service on its stored spec):
+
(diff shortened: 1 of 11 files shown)


alexcos20 (commented):

Overall risk: low

Summary:
This PR elegantly introduces RESPEC mode for `serviceRestart`, giving users the ability to dynamically update container specifications (e.g., swapping a new image tag to fix a bug) without losing their paid duration or host resources. The distinction between 'all-old' (REUSE) and 'all-new' (RESPEC) prevents fragmented, hard-to-debug container states. The logic is securely implemented with proper edge-case handling, including environment variable decryption validation and Dockerfile build checks.

Comments:
• [INFO][style] Good use of strict validation here. Enforcing an 'all-or-nothing' container specification when any single container parameter is supplied keeps the service state predictable and avoids messy, partial overwrites.
• [INFO][security] Excellent authoritative backstop. Even though the handler fast-fails, having the engine re-validate the daemon's `allowImageBuild` config prevents race conditions if the environment changes right before the engine processes the operation.
• [INFO][other] The tests clearly demonstrate the expected behavior of falling back to default image `CMD` and `ENTRYPOINT` when RESPEC mode omits the command overrides or supplies empty arrays. Very thorough testing.

PR #1443, opened 2026-08-05
fix Negative values in getComputeEnvs

### LINKED ISSUE #1439: Negative values in getComputeEnvs
```
[
  {
    "id": "0xff1004b67de08fc505fbf0a2089010d0f23015338c7def8557697513c4a39935-0x2e308d2161642f4fa082602d60e3a0d88bb6a6a8fd9e6ecbd56c92da067e4cb4",
    "runningJobs": 15,
    "consumerAddress": "0xB54e3DEb3F4E066058F4993156539dB6F2F78eC9",
    "platform": {
      "architecture": "x86_64",
      "os": "linux"
    },
    "access": {
      "addresses": [],
      "accessLists": null
    },
    "fees": null,
    "resources": [
      {
        "id": "cpu",
        "type": "cpu",
        "total": 64,
        "max": 64,
        "min": 1,
        "description": "Intel(R) Xeon(R) CPU @ 2.80GHz",
        "inUse": 15
      },
      {
        "id": "ram",
        "type": "ram",
        "total": 62,
        "max": 62,
        "min": 1,
        "inUse": 15
      },
      {
        "id": "disk",
        "type": "disk",
        "total": 70,
        "max": 38,
        "min": 0,
        "inUse": 18
      }
    ],
    "queuedJobs": 0,
    "queuedFreeJobs": 0,
    "queMaxWaitTime": 0,
    "queMaxWaitTimeFree": 0,
    "runMaxWaitTime": -26786355914.46,
    "runMaxWaitTimeFree": -26786355914.46,
    "enableNetwork": true,
    "storageExpiry": 604800,
    "maxJobDuration": 3600,
    "free": {
      "access": {
        "addresses": [],
        "accessLists": null
      },
      "maxJobDuration": 3600,
      "maxJobs": 15,
      "resources": [
        {
          "id": "cpu",
          "max": 60,
          "inUse": 15
        },
        {
          "id": "ram",
          "max": 55,
          "inUse": 15
        },
        {
          "id": "disk",
          "max": 60,
          "inUse": 18
        }
      ]
    },
    "runningfreeJobs": 15
  }
]
```

alexcos20:
Closes #1439 
# Fix negative `runMaxWaitTime`, queued-job resource under-counting, and silent `getJobs` filters

## Why

A live node returned this from `GET` compute environments:

```json
{
  "runningJobs": 15, "runningfreeJobs": 15, "queuedJobs": 0,
  "maxJobDuration": 3600,
  "runMaxWaitTime": -26786355914.46,
  "runMaxWaitTimeFree": -26786355914.46
}
```

That value is exactly `15 × (3600 − now_in_seconds)`:

```
-26786355914.46 / 15 = -1785757060.96  →  + 3600 = 1785760660.96   ← a valid epoch, the fetch time
```

Every one of the 15 jobs was measuring its elapsed runtime **from the Unix epoch**. Consumers
(ocean.js, the frontend env picker, schedulers) read these fields to estimate when capacity frees
up, so the environment reads as either permanently unavailable or instantly available depending on
which way the client compares. Worse, the same function feeds the strict resource-availability
gate, so a second defect in it lets the engine overcommit the host.

Three independent defects came out of that one payload, plus a fourth found in the neighbouring
job-listing path.

### 1. `'0'` is truthy, so every pull-path job measures from 1970

`buildStartTimestamp` and `algoStartTimestamp` are initialized to the **string** `'0'`, and
`buildStartTimestamp` is only ever overwritten inside `buildImage`. The old code branched on
truthiness:

```ts
const timeElapsed = job.buildStartTimestamp
  ? now - Number.parseFloat(job.buildStartTimestamp)   // parseFloat('0') === 0 → start = epoch
  : now - Number.parseFloat(job.algoStartTimestamp)
maxRunningTime += job.maxJobDuration - timeElapsed
```

`'0'` is truthy, so the first branch is taken for **every job that did not build an image from a
Dockerfile — including jobs mid-algorithm holding a perfectly valid `algoStartTimestamp`**. The
`else` branch is effectively dead for those jobs. Only Dockerfile-build jobs ever reported a
correct remaining time. That matches the payload: all 15 jobs were free, and
`env.free.allowImageBuild` gates builds for free jobs, so all 15 were pull-path.

A genuine overrun cannot reach this magnitude — it would need a job running since 1970. But
overruns *do* make the expression legitimately negative by seconds-to-minutes, between the moment
a job exceeds its budget and the next cron sweep that kills it, so the sentinel and the missing
clamp are two separate fixes.

### 2. `queueMaxWaitTime` is never reset, so released jobs stay "queued" forever

The running/queued discriminator was `job.queueMaxWaitTime === 0`. But `queueMaxWaitTime` is the
caller's **requested** maximum queue wait — assigned once at job creation and never written again.
The queue-release path flips `status` to `BuildImage`/`PullImage` and starts the container, but
leaves `queueMaxWaitTime` at its original positive value.

So for the whole life of any job that was ever queued:

| Consequence | Severity |
|---|---|
| counted in `queuedJobs` / `queMaxWaitTime` forever, even while its container runs | cosmetic |
| never counted in `runningJobs` / `runningfreeJobs` | cosmetic |
| **its cpu/ram/disk/gpu are never added to `usedResources`** | **overcommit** |

The last row is the real hazard. `checkIfResourcesAreAvailable` and
`checkGlobalResourceAvailability` both gate on `envResource.inUse`, so under-counting a running
job's resources lets the engine admit work the host cannot serve — including via the queue-release
path itself, which calls `checkIfResourcesAreAvailable` and can therefore release several queued
jobs against the same phantom-free capacity. It also skews `runningJobs + 1 > maxJobs`, so a node
can exceed its own `maxJobs`.

The queued branch also added a *runtime* budget (`job.maxJobDuration`) to a *queue-wait* metric.

### 3. Free-tier availability gate silently passes on `NaN`

Gate 1 defends itself (`envResource.total - (envResource.inUse ?? 0)`), but the free-tier gate did
not: `envResource.total - envResource.inUse`. Both fields are optional on `ComputeResource`, and a
sparse config (or an unresolved free-resource ref) makes that `NaN`. `NaN < amount` is `false`, so
**the free gate passes** — unlimited free allocation.

This is the same failure mode as #2 (under-counted usage → overcommit) in the same function, and
#2 changes exactly what feeds `inUse`, so the two belong together.

> Side note for whoever owns the reporting node: in that payload the `free.resources` entries carry
> `max`/`inUse` but **no `total`**, which current HEAD cannot produce (`resolveEnvironmentResources`
> spreads the pool resource through). That node is probably running a build predating free-resource
> pool resolution — on which this `NaN` gate would have been live, explaining 15 concurrent free
> jobs saturating the box. Worth checking its commit; the `?? 0` hardening lands regardless so a
> sparse config can never reintroduce it.

### 4. `getJobs` filters return the wrong rows, silently

Separate subsystem, found while tracing the above. `dateCreated`/`dateFinished` are `TEXT` holding
decimal seconds (`"1785760660.961"`), and `getJobs` bound the caller's `fromTimestamp` straight
through as a string — so SQLite did a **memcmp**, not a numeric comparison:

| Input | vs `"1785760660.961"` | Result |
|---|---|---|
| seconds `"1785760660"` | equal prefix, shorter is less | works — **by digit-count accident only** |
| ms `"1785760660961"` | pos 10: `'9'` (0x39) > `'.'` (0x2E) | wrong rows |
| ISO `"2026-08-04T…"` | pos 0: `'2'` > `'1'` | wrong rows |
| garbage `"abc"` | `'a'` (0x61) > `'1'` | wrong rows |

Every failure was `[]` (or a wrong subset) with HTTP 200 and no log line — indistinguishable from
"no jobs in that window". The handler validated only `typeof === 'string'`. `ORDER BY dateCreated
DESC` was lexicographic for the same reason. And `if (status)` dropped the predicate entirely for
`C2DStatusNumber.JobStarted`, which is `0` — filtering for `JobStarted` returned **every** status.

### Capacity reporting — `compute_engine_base.ts`

**One shared numeric guard, not a truthiness fix.** Changing the initializer from `'0'` to
`null` would fix this instance while leaving the identical trap for the next field or caller, and
would not repair rows already in the DB carrying `'0'`. So:

```ts
export function parseJobTimestamp(raw?: string): number {
  if (!raw) return 0
  const parsed = Number.parseFloat(raw)
  if (!Number.isFinite(parsed) || parsed <= 0) return 0
  return parsed
}
```

Callers only ever test `> 0`. Two helpers build on it:

- `getJobRemainingRuntimeSeconds(job, nowSec)` — `buildStartTimestamp` wins when both are set
  (build time counts against `maxJobDuration`, consistent with the runtime-expiry check in the
  docker engine); clamped to `[0, maxJobDuration]`; `maxJobDuration` itself guarded with
  `Number.isFinite`. **A job with no usable timestamp on either field reports its full budget**,
  not `now` — it is allocated but not yet ticking (`PullImage`/`BuildImage`/`ConfiguringVolumes`).
- `getJobRemainingQueueWaitSeconds(job, nowSec)` — mirrors the queue-expiry check
  (`queueMaxWaitTime - (now - dateCreated)`), same clamp, and the same "no usable `dateCreated` ⇒
  report the full requested wait" rule.

**Discriminate on `status`, do not mutate `queueMaxWaitTime`.** Zeroing the stored value on release
would be wrong: it is a contract value, reused for the queue-expiry check and for the escrow lock
duration (`maxJobDuration + queueMaxWaitTime`), and it would need a DB migration. `status` is
already an exact, persisted, migration-free signal — `C2DStatusNumber.JobQueued` is set at creation
**iff** `queueMaxWaitTime > 0` and is replaced on release, and it is the *only* pre-allocation
state:

```ts
const isQueued = job.status === C2DStatusNumber.JobQueued
```

The two branches are inverted accordingly, and the resource tally is now guarded by `!isQueued`, so
every post-queue state contributes its resources. `Date.now()` is hoisted into a single `nowSec` so
all jobs in one response are measured against the same instant.

**Free gate** hardened to `(envResource.total ?? 0) - (envResource.inUse ?? 0)`, matching gate 1 —
a missing value now denies instead of admitting.

### `*MaxWaitTime` semantics: sum kept, documented

All four accumulators still use `+=`. Chosen deliberately over `Math.max` to avoid any contract
change: the only difference now is that each term is non-negative and derived from a valid start
timestamp. Since the sum makes the field names misleading — 15 jobs with 1 s left each report the
same `15` as one job with 15 s left — the four fields in `src/@types/C2D/C2D.ts` now carry doc
comments stating the unit, the summing, and explicitly that **a client cannot derive
time-to-free-capacity from these fields**. A unit test pins the sum so a later refactor cannot
switch it silently.

### Docker engine

`getValidBuildDurationSeconds` refactored onto `parseJobTimestamp` (it already had the correct
`start <= 0` check inline — now one shared rule instead of a second copy, across its three call
sites). The runtime-expiry check no longer does `parseFloat(job.algoStartTimestamp)` raw: a `'0'`
sentinel there would place the start at the epoch and make a healthy container look instantly
expired, so it falls back to `timeNow` and lets the next sweep see the real timestamp.

### `getJobs` — normalize at the boundary, compare numerically in SQL

- **New `src/components/core/utils/timestamps.ts`** — `parseFromTimestamp` moved here from
  `getServices.ts` (which re-exports it, so the service handler and its tests are untouched), plus
  a `parseFromTimestampSeconds()` wrapper. The seconds-vs-ms choice is explicit at the call site
  rather than implicit: the compute-jobs table stores seconds, the service path works in ms.
  A header comment separates this from `parseJobTimestamp` — that one reads *stored* timestamps and
  treats `'0'` as unset; this one normalizes *query parameters* and must distinguish "no filter"
  from "garbage".
- **`GetJobsHandler.validate`** rejects an unparseable `fromTimestamp` with `400` and the same
  message shape as `GetServicesHandler`, instead of `200 []`. Absent/empty still means "no filter",
  as before.
- **DB layer takes `number` (seconds)**, not a string — type-safety over convenience, so a raw
  string can never reach the SQL again. Threaded through `SQLiteCompute.getJobs` and the
  `C2DDatabase` pass-through, and the **stale `ComputeDatabaseProvider.getJobs` declaration**
  (3 of 5 params; not constraining anything today because `C2DDatabase.provider` is typed as the
  concrete class) fixed to the real shape while here.
- **`CAST(dateCreated AS REAL) >= ?`** / `CAST(dateFinished AS REAL) >= ?`, and
  `ORDER BY CAST(dateCreated AS REAL) DESC`.
- **`if (status !== undefined && status !== null)`** so `status = 0` filters.

No schema or migration change — only how existing `TEXT` columns are *compared*.

> A comment on the `ORDER BY` records why this `CAST` must **not** be copied to the `service_jobs`
> table: its `dateCreated` is `toISOString()`, which is fixed-width UTC and sorts correctly as text,
> where `CAST("2026-08-04T…" AS REAL)` would be `2026`. The two tables genuinely disagree on
> timestamp format.

## Summary by CodeRabbit

* **New Features**
  * Added flexible job filtering by Unix timestamps, milliseconds, and ISO-formatted dates.
  * Added validation with clear errors for invalid timestamp filters.
* **Bug Fixes**
  * Improved compute wait-time and runtime calculations for queued, running, completed, and pending jobs.
  * Corrected handling of missing or invalid timestamps to prevent inaccurate expiry and duration results.
  * Improved free-resource capacity checks when usage or capacity information is incomplete.
* **Documentation**
  * Clarified how queue and runtime wait totals are calculated across jobs, including free-job aggregates.

### DIFF
--- a/src/@types/C2D/C2D.ts
+++ b/src/@types/C2D/C2D.ts
@@ -181,9 +181,25 @@ export interface ComputeEnvironment extends ComputeEnvironmentBaseConfig {
   consumerAddress: string // v1
   queuedJobs: number
   queuedFreeJobs: number
+  /**
+   * Seconds of remaining queue wait, **summed across every queued job** in this environment
+   * (not a per-job maximum, despite the name). Each job contributes
+   * `max(0, queueMaxWaitTime - elapsed since it was created)`.
+   */
   queMaxWaitTime: number
+  /** As `queMaxWaitTime`, summed over queued *free* jobs only. */
   queMaxWaitTimeFree: number
+  /**
+   * Seconds of remaining runtime, **summed across every running job** in this environment
+   * (not a per-job maximum, despite the name). Each job contributes
+   * `max(0, maxJobDuration - elapsed since it started)`; a job that has been allocated but
+   * has not started executing yet contributes its full `maxJobDuration`.
+   *
+   * Because it is a sum, a client cannot derive "when does capacity free up" from this field:
+   * 15 jobs with 1 second left each report 15, the same as one job with 15 seconds left.
+   */
   runMaxWaitTime: number
+  /** As `runMaxWaitTime`, summed over running *free* jobs only. */
   runMaxWaitTimeFree: number
 }
 
--- a/src/components/c2d/compute_engine_base.ts
+++ b/src/components/c2d/compute_engine_base.ts
@@ -785,7 +785,15 @@ export abstract class C2DEngine {
 
       // Gate 1 (per-env ceiling) — fungible resources only.
       // envResource.total = env aggregate ceiling (from EnvironmentResourceRef.total).
-      if (isFungible && envResource.total - (envResource.inUse ?? 0) < request.amount)
+      // Both operands are defaulted: a resource object missing either field would make the
+      // subtraction NaN, and `NaN < amount` is false, i.e. the gate would silently admit the
+      // request. Missing must deny, not admit. (`total` is required by the type, but these
+      // objects also arrive from JSON config and older persisted shapes, which the compiler
+      // cannot vouch for — free-resource entries with no `total` have been seen in the field.)
+      if (
+        isFungible &&
+        (envResource.total ?? 0) - (envResource.inUse ?? 0) < request.amount
+      )
         throw new Error(`Not enough available ${request.id} in this environment`)
 
       // Gate 2 (engine-wide pool ceiling) — fungible + exclusive discrete.
@@ -798,11 +806,10 @@ export abstract class C2DEngine {
         if (!env.free) throw new Error(`No free resources`)
         envResource = this.getResource(env.free?.resources, request.id)
         if (!envResource) throw new Error(`No such free resource ${request.id}`)
-        // `total` and `inUse` are both optional on ComputeResource: a sparse config (or an
-        // unresolved free-resource ref) makes the subtraction NaN, and `NaN < amount` is
-        // false, which would silently pass the gate and allow unlimited free allocation.
-        // Default both to 0 so a missing value denies instead of admitting — same shape as
-        // gate 1 above.
+        // Same NaN-admits hazard as gate 1, and more reachable here: `inUse` is optional on
+        // ComputeResource, and free-resource entries carrying `max`/`inUse` but no `total`
+        // have been observed on live nodes (builds predating free-resource pool resolution).
+        // Unguarded, that yields NaN and allows unlimited free allocation.
         if ((envResource.total ?? 0) - (envResource.inUse ?? 0) < request.amount)
           throw new Error(`Not enough available ${request.id} for free`)
       }
--- a/src/components/c2d/compute_engine_base.ts
+++ b/src/components/c2d/compute_engine_base.ts
@@ -17,7 +17,7 @@ import type {
   ComputeEnvFees
 } from '../../@types/C2D/C2D.js'
 import type { ServiceJob } from '../../@types/C2D/ServiceOnDemand.js'
-import { C2DClusterType } from '../../@types/C2D/C2D.js'
+import { C2DClusterType, C2DStatusNumber } from '../../@types/C2D/C2D.js'
 import { C2DDatabase } from '../database/C2DDatabase.js'
 import { Escrow } from '../core/utils/escrow.js'
 import { KeyManager } from '../KeyManager/index.js'
@@ -30,6 +30,24 @@ import { ValidateParams } from '../httpRoutes/validateCommands.js'
 import { EncryptMethod } from '../../@types/fileObject.js'
 import { CORE_LOGGER } from '../../utils/logging/common.js'
 import { DockerRegistryAuthSchema } from '../../utils/config/schemas.js'
+
+/**
+ * Parse a job timestamp stored as decimal seconds in a string column.
+ *
+ * Job timestamps are initialized to the *string* `'0'` (not null/empty) and only get a real
+ * value once the corresponding phase starts. `'0'` is truthy, so a truthiness check happily
+ * accepts it and `Number.parseFloat('0')` then places the job's start at the Unix epoch,
+ * producing ~56 years of "elapsed" time. Always go through this helper: it returns 0 for the
+ * `'0'` sentinel, for null/undefined/empty, and for anything non-finite or negative, so
+ * callers only have to test `> 0` to know whether a real timestamp is present.
+ */
+export function parseJobTimestamp(raw?: string): number {
+  if (!raw) return 0
+  const parsed = Number.parseFloat(raw)
+  if (!Number.isFinite(parsed) || parsed <= 0) return 0
+  return parsed
+}
+
 export abstract class C2DEngine {
   private clusterConfig: C2DClusterInfo
   public db: C2DDatabase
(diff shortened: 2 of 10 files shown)


alexcos20 (commented):

Overall risk: low

Summary:
This is an excellent PR that resolves multiple subtle bugs. It corrects negative `runMaxWaitTime` by correctly identifying unset ('0') timestamps, patches an unbound free-resource allocation flaw caused by `NaN` comparison evaluation, and correctly handles SQLite filtering and sorting semantics (0-index status drops, TEXT vs REAL comparison). The code is well-structured, carefully documented, and accompanied by exhaustive unit tests. LGTM!

Comments:
• [INFO][performance] Using `CAST(dateCreated AS REAL)` in the `WHERE` clause (and later in the `ORDER BY` clause) accurately resolves the string vs. numeric sorting issue. However, please note that applying a `CAST` (or any function) to a column in SQLite usually prevents the query planner from using an index on that column, resulting in a full table scan. This is acceptable for now to ensure correctness, but if this table grows significantly, consider migrating the column data type to `REAL` or `INTEGER` natively in a future schema migration to restore index usage.
• [INFO][security] Excellent catch using the nullish coalescing operator `?? 0` here. A sparse configuration resulting in `NaN < amount` silently yielding `false` and bypassing the free allocation constraint is a subtle but critical logical vulnerability. This reliably patches it.
• [INFO][style] Instantiating `const nowSec = Date.now() / 1000` once per loop iteration is a great practice. It guarantees a consistent reference time for every job in the list and prevents minor calculation drift during the loop execution.

giurgiur99 (approved):

PR #1448, opened 2026-08-14
remove serviceEndpoints

alexcos20:
## Requires ocean.js v9.xx (https://github.com/oceanprotocol/ocean.js/pull/2133 or later)

# Remove `serviceEndpoints` from the root announcement; fix `:job` route param

## Context

[ocean.js PR #2133](https://github.com/oceanprotocol/ocean.js/pull/2133) ("Hardcode http
paths") removes the SDK's dynamic endpoint-discovery step: previously every `HttpProvider`
call did a `GET /` on the node first to read a `serviceEndpoints` map and look up the right
method/path, before making the real request. That PR hardcodes every path directly in the
SDK instead — partly for performance, partly because discovery was silently routing two
calls to the wrong path (`serviceGetStreamableLogs`, Aquarius `querySearch`).

Since ocean.js (the primary consumer) no longer reads `serviceEndpoints`, ocean-node has no
reason to keep computing and exposing it: `GET /` was running a full introspection pass over
the Express router stack on every hit just to build a map nothing external reads anymore.
Separately, while touching this area, the job-status route parameter was named `:job` while
every other reference to a job id in this codebase — `getJobsPool(jobId)`, the compute/escrow
query params, the Postman collection's `{{jobId}}` — uses `jobId`. Fixed for consistency.
Express route param names are internal only, so this is a no-op on the wire for any caller.

## Changes

**Remove `serviceEndpoints` from `GET /`**
- `src/components/httpRoutes/rootEndpoint.ts` — dropped the `serviceEndpoints:
  getAllServiceEndpoints()` field from the response and its now-unused import. This also
  removes a latent circular import (`rootEndpoint.ts` → `index.ts` → `rootEndpoint.ts`).
- `src/components/httpRoutes/index.ts` — deleted `getAllServiceEndpoints()` and the
  `routeUtils.js` import; router setup and all `httpRoutes.use(...)` mounting untouched.
- `src/components/httpRoutes/routeUtils.ts` — deleted entirely (`routesNames`,
  `allRoutesMapping`, `findPathName`, `addMapping`, the regex-based `split()` helper). It had
  no other consumers, and had already drifted from the real routes (e.g. it still mapped a
  `/getOceanPeers` route that doesn't exist).
- `docs/Ocean Node.postman_collection.json` — "Get Node Info" request description no longer
  lists `serviceEndpoints` as a response field.

**Keep the k6 perf scripts working**
- `src/test/performance/util.js` — `stepRootEndpoint()` used to build its target list from
  `data.serviceEndpoints` on the root response. Replaced with a hardcoded `SERVICE_ENDPOINTS`
  list (56 entries) built directly from the routes actually mounted in
  `httpRoutes/index.ts`, keeping a plain `GET /` 200 health check before looping instead of
  using it for discovery.

**Fix the `:job` route param**
- `src/components/httpRoutes/jobs.ts` — renamed `${SERVICES_API_BASE_PATH}/jobs/:job` to
  `.../jobs/:jobId`, and `req.params.job` to `req.params.jobId`, matching
  `indexer.getJobsPool(jobId)` and the Postman collection's existing `{{jobId}}` convention.

**Remove the now-dead `RouteOptions` type**
- `src/@types/express.ts` — deleted. Its only export, `RouteOptions`, existed solely for the
  deleted `routeUtils.ts`; grepped `src/` for `@types/express`/`RouteOptions` and confirmed
  zero remaining importers.

## Summary by CodeRabbit

* **New Features**
  * Updated job routes to use clearer job ID parameters.

* **Bug Fixes**
  * Root and node information responses no longer include service endpoint listings.
  * Health checks now stop immediately when unsuccessful.

* **Documentation**
  * Updated API request documentation to reflect that node responses contain identity information only.

* **Tests**
  * Performance checks now use a predefined set of service endpoints instead of discovering routes dynamically.


alexcos20 (commented):

Overall risk: low

Summary:
This PR successfully removes the dynamic `serviceEndpoints` field from the root endpoint. It deletes the complex and computationally expensive route mapping logic, which improves security (by not exposing the entire API topology dynamically) and performance of the root endpoint. The performance tests were correctly updated to use a static list of endpoints. Additionally, the route parameter in `jobs.ts` was standardized from `:job` to `:jobId`.

Comments:
• [INFO][security] Good removal of the `getAllServiceEndpoints()` call. Removing dynamic API route exposure improves security by preventing automated API topology discovery, and avoids traversing the Express router stack on every request to `/`.
• [INFO][style] Standardizing the path parameter from `:job` to `:jobId` makes the API parameter naming more precise and consistent.
• [INFO][style] Using a static list of endpoints for performance testing is a good tradeoff here to maintain test coverage without relying on the application dynamically leaking its routes. The implementation correctly preserves the behavior of testing all known API boundaries.

alexcos20:
test_system is failing because it's using ocean.js v8.xx, which is looking for serviceEndpoints

PR #1449, opened 2026-08-17
Authenticate `PolicyServerPassthrough` / `initializePSVerification`, and stop logging credentials

alexcos20:
# Authenticate `PolicyServerPassthrough` / `initializePSVerification`, and stop logging credentials

> **Breaking change.** Both `/api/services/PolicyServerPassthrough` and
> `/api/services/initializePSVerification` now require authentication. Existing
> unauthenticated callers will receive `401`.

### 1. The passthrough endpoint was completely unauthenticated

`PolicyServerPassthroughHandler` did rate-limiting and a single presence check, then handed the
caller's object to the PolicyServer verbatim:

```ts
validate(command: PolicyServerPassthroughCommand): ValidateParams {
  if (!command.policyServerPassthrough)
    return buildInvalidRequestMessage('Invalid Request: missing policyServerPassthrough field!')
  const validation = validateCommandParameters(command, []) // all optional? weird
  return validation
}
```

`PolicyServer.passThrough` does no shaping whatsoever:

```ts
async passThrough(request: any): Promise<PolicyServerResult> {
  return await this.askServer(request)
}
```

The only server-controlled fields in the outgoing body were `ddo` (forced to the DB lookup) and
`nodeAddress` (added by `attachNodeAddress`). Everything else — **including `action`** — was
attacker-supplied.

Two consequences:

**No caller identity.** The PolicyServer exists to make authorization decisions, but on this
endpoint it received nothing trustworthy about _who_ was asking. It could not do the additional
checks it is there for.

**Impersonation of the typed actions.** Every other PolicyServer call site builds an explicit
`{action, ...}` payload from an already-authenticated handler:

| method                              | action                    | called from                                             |
| ----------------------------------- | ------------------------- | ------------------------------------------------------- |
| `checkDownload`                     | `download`                | `downloadHandler.ts` (after `validateTokenOrSignature`) |
| `checkStartCompute`                 | `startCompute`            | `startCompute.ts`, `initialize.ts`                      |
| `checkEncrypt` / `checkEncryptFile` | `encrypt` / `encryptFile` | `encryptHandler.ts`                                     |
| `initializePSVerification`          | `initiate`                | `PolicyServerInitializeHandler`                         |
| `passThrough`                       | **caller's choice**       | `PolicyServerPassthroughHandler`                        |

So a client could POST

```json
{
  "policyServerPassthrough": {
    "action": "download",
    "consumerAddress": "0xvictim",
    "documentId": "did:op:…"
  }
}
```

and produce a payload the PolicyServer cannot distinguish from one the node itself built in
`checkDownload` — with an arbitrary `consumerAddress` and no credentials anywhere in the request.

### 2. The sibling `initializePSVerification` route had the same gap

`PolicyServerInitializeHandler` _required_ `consumerAddress` but never verified the caller owned
it, so anyone could trigger a PolicyServer verification flow on behalf of any address.

### 3. `validateCommandParameters` logged every credential, for every command

Found while adding auth to the above. This function logs the whole command object on **every**
request:

```ts
CORE_LOGGER.info(
  `Checking received command data for Command "${commandStr}": ${JSON.stringify(logCommandData, null, 4)}`
)
```

It already redacted `files` and `rawData`, but not credentials — so `authorization` (the JWT),
`signature`, `aes_encrypted_key` and `encryptedDockerRegistryAuth` were written verbatim to the
logs for `download`, `startCompute`, every service command, the persistent-storage commands, and
now these two. The JWT is a bearer credential valid until expiry, so a log reader could replay it
as the user.

Pre-existing and repo-wide, not introduced here — but adding auth to two more endpoints would have
widened it, so it is fixed in the same PR.

### Auth: reuse the existing helper

No new auth code. Both handlers now call the shared
`CommandHandler.validateTokenOrSignature(...)` immediately after `verifyParamsAndRateLimits`,
exactly as `PaidComputeStartHandler.handle` does:

```ts
const authValidationResponse = await this.validateTokenOrSignature(
  task.authorization,
  task.consumerAddress,
  task.nonce,
  task.signature,
  task.command
)
if (authValidationResponse.status.httpStatus !== 200) {
  return authValidationResponse
}
```

That routes to `Auth.validateAuthenticationOrToken`, so the accepted credentials are the same as
everywhere else in the node: an `Authorization` token, **or** `consumerAddress` + `nonce` +
`signature`. `validate()` additionally requires `consumerAddress` and checks it with `isAddress()`,
matching `startCompute`.

Ordering matters and is asserted by a test: parameter validation runs _before_ auth, so a malformed
request gets `400` and never touches the auth component.

### The forwarded address is the _verified_ one, not the claimed one

This is the substantive decision. `Auth.validateAuthenticationOrToken` has two paths:

- **signature path** — safe, because `verifyConsumerSignature` builds its message as
  `consumer + nonce + command + issuerPeerId`, so the address is bound into what was signed.
- **token path** — returns the address encoded in the JWT, which **need not equal**
  `task.consumerAddress`.

Most handlers in this repo then keep using `task.consumerAddress` and inherit the gap: a caller
holding a valid token for A can pass `consumerAddress: B` and still get a 200.
`persistentStorage.ts` is the one that does it right, binding to `isAuthRequestValid.consumerAddress`.

This PR follows persistentStorage. The address forwarded to the PolicyServer is
`authValidationResponse.consumerAddress`, assigned unconditionally so it always replaces whatever
the client supplied — the same way `ddo` was already forced server-side:

```ts
// the passthrough payload is forwarded verbatim, so every identity field has to be
// (re)written here, after validation. otherwise a caller could forge consumerAddress
// and impersonate the typed actions (download, startCompute, ...)
task.policyServerPassthrough.consumerAddress = authValidationResponse.consumerAddress
task.policyServerPassthrough.authorization = task.authorization
task.policyServerPassthrough.nonce = task.nonce
task.policyServerPassthrough.signature = task.signature
```

The injection sits **after** the DDO-resolution block so a client cannot pre-seed these fields.

For `initializePSVerification` the verified address is passed as the `consumerAddress` argument and
the credentials ride in the free-form `policyServer` object — the only field that method forwards
unshaped:

```ts
const policyStatus = await policyServer.initializePSVerification(
  task.documentId,
  ddo,
  task.serviceId,
  authValidationResponse.consumerAddress,
  {
    ...task.policyServer,
    authorization: task.authorization,
    nonce: task.nonce,
    signature: task.signature
  }
)
```

**Credentials are relayed on purpose** so the PolicyServer can run its own independent checks —
it can recompute `consumerAddress + nonce + "PolicyServerPassthrough"` and verify the signature
itself. `docs/PolicyServer.md` notes that this means `POLICY_SERVER_URL` should be an HTTPS
endpoint the operator controls.

### Command shape

`consumerAddress` / `nonce` / `signature` are **top-level** fields on the command, alongside the
inherited `Command.authorization` — consistent with `download`, `startCompute` and the service
commands, and it keeps `policyServerPassthrough` as a purely opaque client payload. The HTTP routes
read the token from the header and the triple from the body, the convention used in `compute.ts`
and `provider.ts`:

```ts
consumerAddress: req.body.consumerAddress,
nonce: req.body.nonce,
signature: req.body.signature,
authorization: req.headers?.authorization,
```

The P2P / `POST /directCommand` path needed no change — it JSON-parses the whole command object, so
the new fields arrive automatically.

### `initializePSVerification` gets its own command

Previously both routes sent `command: PROTOCOL_COMMANDS.POLICY_SERVER_PASSTHROUGH`, so the two
endpoints were indistinguishable at the dispatch layer. Now that the command string is what a
signature is scoped to, sharing it would mean a signature for one endpoint is replayable against
the other. `initializePSVerification` therefore gets its own command:

```ts
POLICY_SERVER_PASSTHROUGH: 'PolicyServerPassthrough',
POLICY_SERVER_INITIALIZE: 'PolicyServerInitialize',
```

Added to **both** `PROTOCOL_COMMANDS` and `SUPPORTED_PROTOCOL_COMMANDS` — the allow-list entry is
mandatory, since `validateCommandParameters` rejects any command not on it
(`Invalid or unrecognized command`), so the constant alone would have broken the route outright.

`PolicyServerInitializeHandler` is also now registered in `CoreHandlersRegistry`. It was previously
HTTP-only and unregistered; leaving it that way while adding its command to the allow-list would
have made `POST /directCommand` with `PolicyServerInitialize` pass validation and then fail with
`501 Unknown command or missing handler`. Registering it follows the add-a-command checklist in
CLAUDE.md and makes the endpoint reachable over P2P like every other command — safe now that it is
authenticated.

### One robustness fix

`typeof command.policyServerPassthrough !== 'object'` → `400`. The handler does
`task.policyServerPassthrough.ddo = null`, which throws a `TypeError` (ESM is strict mode) on a
string payload — a `500` on malformed input, on the path now carrying auth.

### Credential redaction — `validateCommands.ts`

Applied unconditionally, after the existing deep copy, so it covers every command:

```ts
const SENSITIVE_COMMAND_FIELDS = [
  'authorization', // auth token (JWT), usually from the Authorization header
  'signature', // consumer signature authorizing this command
  'aes_encrypted_key', // download: encrypted key material
  'encryptedDockerRegistryAuth' // compute: encrypted docker registry credentials
]
```

`token` needed care and is handled per-command:

| command                                             | `token` means                       | redacted |
| --------------------------------------------------- | ----------------------------------- | -------- |
| `getEscrowEvents`, `computeStart` (`payment.token`) | ERC20 contract address              | no       |
| `invalidateAuthToken`, `validateAuthToken`          | a JWT (`jwt.verify(task.token, …)`) | yes      |

Blanket-redacting `token` would have made escrow and payment logs useless, so only the two
auth-token commands are covered.

Redaction runs on the copy, so handlers still receive the real credentials — pinned by a test.
It also runs after the `structuredClone`/shallow-clone fallback, so a command carrying a
non-cloneable value still gets redacted.

## Tests

Two new unit test files. There were **zero** tests touching PolicyServer or
`validateCommandParameters` before this.

**`src/test/unit/policyServer.test.ts`** — 17 tests. Sinon fakes for the node (no DB, no network),
`PolicyServer.prototype` stubbed to inspect the forwarded payload:

- `400` on missing / non-object `policyServerPassthrough`, missing / malformed `consumerAddress`
- validation runs before auth (the auth component is asserted _not_ called)
- `401` on absent credentials, on an invalid signature, and when the Auth component is unwired
- the auth header and the nonce/signature triple reach `validateAuthenticationOrToken`, scoped to
  the command string
- the forwarded payload carries the verified address, the credentials, the resolved `ddo`, and the
  caller's own fields
- **impersonation guards** — a forged `consumerAddress` inside the payload is overwritten; a
  mismatched address alongside a valid token loses to the token's address
- DDO-not-found still forwards with `ddo: null`; a PolicyServer denial propagates its status/body
- `initializePSVerification`: `401` on bad auth, `404` on a missing DDO, and the verified address +
  credentials in the `policyServer` blob

## Docs

- **`docs/API.md`** — passthrough section documents the auth requirement, the request header, the
  new parameters, the signed-message form and the `401`. Adds a section for
  `initializePSVerification`, which was a live but entirely undocumented route.
- **`docs/PolicyServer.md`** — new "Passthrough and caller identity" section with a table of which
  fields are node-set vs caller-supplied, plus two warnings for PolicyServer implementors: the
  caller controls `action` (so a passthrough request must not be granted node-initiated authority),
  and the caller's token is relayed.

## Migration

Callers of either endpoint must add `consumerAddress` plus either an `Authorization` header or a
`nonce`/`signature` pair.

On the signature path, each endpoint signs over its **own** command string — a signature is not
transferable between them:

| endpoint                                 | signed message                                        |
| ---------------------------------------- | ----------------------------------------------------- |
| `/api/services/PolicyServerPassthrough`  | `consumerAddress + nonce + "PolicyServerPassthrough"` |
| `/api/services/initializePSVerification` | `consumerAddress + nonce + "PolicyServerInitialize"`  |

Nothing else changes: no schema change, no config change, no new env var, and the response shape on
success is unchanged.

## Summary by CodeRabbit

- **New Features**
  - Added authenticated PolicyServer initialization verification.
  - Expanded PolicyServer passthrough and initialization requests with consumer identity and authentication options.
  - Added support for forwarding verified caller credentials and identity.
  - Expanded API documentation for storage, compute, service management, restart, and log retrieval endpoints.

- **Security**
  - Added validation for consumer addresses and authentication credentials.
  - Sensitive credentials are now redacted from command logs.

- **Bug Fixes**
  - Corrected initialization requests to use their dedicated protocol command.

- **Tests**
  - Added coverage for authentication, identity propagation, validation, credential forwarding, and secure logging.


alexcos20 (commented):

Overall risk: low

Summary:
This PR introduces authentication and caller identity verification for the PolicyServer passthrough endpoints. It properly overwrites user-provided identity fields with verified ones to prevent impersonation, and adds excellent security hardening by redacting sensitive credentials from the application logs.

Comments:
• [INFO][bug] In JavaScript, arrays are considered objects (`typeof [] === 'object'`). If the caller maliciously or accidentally passes an array for `policyServerPassthrough`, it will pass this check but could cause unexpected behavior later when properties are injected into it. Consider explicitly ruling out arrays.

```diff
-    if (typeof command.policyServerPassthrough !== 'object')
+    if (typeof command.policyServerPassthrough !== 'object' || Array.isArray(command.policyServerPassthrough))
```
• [INFO][security] Excellent security practice here. Ensuring that sensitive credentials (like JWTs, signatures, and encrypted keys) are redacted before reaching the logger mitigates the risk of credential leakage in server logs.
• [INFO][security] Good job overwriting the caller-supplied `consumerAddress` with the verified one from the auth component. This is a critical security control to prevent users from impersonating other addresses in the passthrough payload. LGTM!

AdriGeorge (approved):

PR #1453, opened 2026-08-21
Dependency and toolchain refresh

alexcos20:
# Dependency and toolchain refresh: 53 → 2 advisories, 424 fewer packages

Aligns `ocean-node` with the toolchain refresh in
[ocean.js#2137](https://github.com/oceanprotocol/ocean.js/pull/2137), clears the security
backlog, and removes eight dependencies the code no longer used.

| | before | after |
| --- | --- | --- |
| npm advisories | **53** (18 high, 14 moderate, 21 low) | **2** (0 high, 0 moderate, 2 low) |
| installed packages | 1454 | 1030 |
| direct dependencies | 56 | 50 |
| devDependencies | 29 | 26 |
| `overrides` entries | 8 | 2 |
| packages behind latest | 63 | 2 (both deliberately held) |

Verified: `npm run lint` (eslint + type-check) **0 errors**, `tsc --noEmit` **0 errors**,
`npm run build` clean, `npm run test:unit` **532 passing** (up from 525) with the one
pre-existing failure noted under [Testing](#testing). Integration tests need Barge and could
not run here — see [What still needs verifying](#what-still-needs-verifying).

> **No API or config changes.** No env var added or removed, no schema change, no response
> shape change. The one externally-visible string — `friendlyName` in the node status
> response — is proven byte-identical.

---

## 1. The finding that motivated this: version skew, not staleness

The most consequential problem was invisible to `npm outdated`. Three direct dependencies
were pinned **below** the version `libp2p@3.x` itself depends on, so the tree carried two or
three copies of each:

| package | declared | tree held | libp2p 3.3.8 requires |
| --- | --- | --- | --- |
| `@multiformats/multiaddr` | `^12.2.3` | 12.5.1 **+ 13.0.1** | `^13.0.3` |
| `uint8arrays` | `^4.0.6` | 4.0.10 **+ 5.1.0 + 5.1.1** | `^6.1.1` |
| `multiformats` | *undeclared* | 12.1.3 + 13.4.2 | `^14.0.0` |
| `@libp2p/interface` | — | 1.7.0 **+ 2.11.0 + 3.2.2** | `^3.2.5` |

Multiaddr, key and CID objects are version-branded: one built through the v12 API is not
accepted by libp2p code holding v13 types. This matters across repos, because #2137 promotes
libp2p from a bundled dev dependency to a real runtime dependency of `@oceanprotocol/lib`,
shipping `multiaddr@13` and `uint8arrays@6` to every client. Clients would have been dialling
a node that built multiaddrs with v12.

All now deduplicated to a single copy each. `@libp2p/interface` collapsed to one version once
the dead `@libp2p/peer-id-factory` (a libp2p-v1-era package) was dropped.

A fourth instance of the same pattern surfaced during the upgrade and was **not** in the
original audit: `datastore-level@12` depends on `interface-datastore@9` while libp2p 3.3.8
needs `@10`. It produced a hard type error on the libp2p init options — `Key` has "separate
declarations of a private property `_buf`". Fixed by `datastore-level@13`.

## 2. Removed: eight unused dependencies

The first sweep used a plain grep and let four through, because it counted **commented-out
imports** and same-named local variables as usage. The second pass strips comments before
matching module specifiers.

| package | why |
| --- | --- |
| `it-pipe` | zero references |
| `@libp2p/peer-id-factory` | zero references; sole cause of `@libp2p/interface@1.7.0` in the tree |
| `@libp2p/pubsub` | zero references (only `pubsub-peer-discovery` was imported) |
| `delay` | **never imported** — the 20 apparent hits were local variables named `delay`, plus comments |
| `stream-concat` | no import; only a `src/@types/stream-concat.d.ts` shim, also removed |
| `hyperdiff` | import at `P2P/index.ts:1` is commented out; shim `P2P/hyperdiff.d.ts` removed |
| `@libp2p/pubsub-peer-discovery` | import at `P2P/index.ts:16` is commented out |
| `humanhash` | replaced by an inlined port — see §5 |

The commented-out import lines are left in place as a record of intent.

Worth noting: `hyperdiff` was earlier bumped 2.0.23 → 2.0.27 to clear a **high** nested-lodash
advisory — an advisory only ever reachable through a package the code does not use. `lodash` is
now a single copy in the tree with nothing attached to it.

Also removed as dev-only dead weight: `@types/ip` (no `ip` package anywhere; `ipaddr.js` ships
its own types), `tsx` (no npm script referenced it — this repo runs mocha against compiled
`dist/`), `concurrently` (no script referenced it), and `@types/node-cron` (redundant, node-cron
4 bundles types).

### Declared: two phantom dependencies

Both were imported but undeclared, resolving only by hoisting luck — a dedupe or lockfile
refresh could have broken the build with no `package.json` change.

- `multiformats@^14.0.5` — imported by `src/utils/conversions.ts`
- `tar-stream@^3.2.0` — imported by `compute_engine_docker.ts:39`, where only
  `@types/tar-stream` was declared

### Two that look unused but are not

`koffi` is loaded as `await import(specifier)` from a **non-literal** variable in
`c2d/gpu/nvml.ts`, deliberately, so the optional FFI dependency is only required at runtime on
NVIDIA hosts — no static scan can see it. `eslint-config-prettier` is never imported by name,
but `eslint-plugin-prettier/recommended` `require`s it at load and it is an *optional* peer, so
npm will not reliably install it transitively. `eslint.config.js` now carries a comment saying
so, since a future cleanup would otherwise flag it.

## 3. `overrides`: eight entries down to two

Six pinned packages no longer anywhere in the tree (`elliptic`, `tough-cookie`, `tmp`,
`base64url`, and `secp256k1` under both `eth-crypto` and `eccrypto` — `eth-crypto@4` dropped
`eccrypto` entirely). `semver` was satisfied naturally. Worse, the `xml2js` pin was actively
**downgrading** upstream from 0.6.x to 0.5.0; the advisory it was added for is fixed in both, so
it was holding two libp2p packages back for nothing.

The two that remain do real work:

```json
"overrides": {
  "serialize-javascript": "^7.1.0",
  "ws": "^8.21.3"
}
```

- `serialize-javascript` — mocha 11.8.0 is the latest release and still pulls a vulnerable
  version; there is no fixed mocha to upgrade to.
- `ws` — `eth-crypto@4.1.0` pins `ethers@6.16.0` *exactly*, dragging in `ws@8.17.1` and two high
  advisories. A scoped `{"eth-crypto": {"ethers": "…"}}` override did not take effect, so the
  vulnerable package is targeted directly.

## 4. ESLint 8 → 10: flat config, held at the old severity

ESLint 8 is end-of-life and was the root of five of the 18 high advisories via
`@typescript-eslint@6`. The blocker was real: `eslint-config-oceanprotocol@2.0.4` is the latest
release and depends on `eslint@^8`, and its own base `eslint-config-standard@17.1.0` is likewise
capped at `^8`. Neither can follow ESLint to flat config. As in #2137, the shared preset is gone
and composed in-repo instead.

- `.eslintrc` and `.eslintignore` deleted; `eslint.config.js` added
- `@typescript-eslint/{eslint-plugin,parser}` → `typescript-eslint`, plus `@eslint/js`,
  `globals` and `eslint-plugin-security@4`
- `--ext` / `--ignore-path` dropped from the `lint` scripts (both removed in ESLint 10)

**The rule set is deliberately pinned to the old gate.** A naive migration is not neutral:
extending `typescript-eslint`'s `recommended` produced **1358 errors**, almost all
`no-explicit-any` — a rule the old config never enabled. Backing that out still left **203**,
from rules ESLint 9/10 and the newer plugins added on top (`no-useless-assignment`,
`preserve-caught-error`, promise's recommended set, and `security/detect-object-injection` at 128
warnings).

Rather than guess, the old gate was **measured**: a throwaway git worktree at unmodified `HEAD`
with the old dependencies installed, then `eslint --print-config`. That produced **0 errors and
40 warnings**, and revealed the two details that mattered — `no-unused-vars` carried
`args: 'none', caughtErrors: 'none'` from `eslint-config-standard` (which is why it reported
nothing), and only `promise/param-names` was enabled, not promise's recommended set. The flat
config reproduces that, and every rule switched off carries a comment explaining why.

### Stale `eslint-disable` directives

ESLint 10 reports unused directives by default; the tree had **56**. Rather than delete all of
them, each was checked for *why* it was dead. 34 were dead only because this PR dropped the rule
they referenced, so re-enabling was tested first:

| rule | cost to re-enable | outcome |
| --- | --- | --- |
| `no-new`, `no-self-compare`, `no-unmodified-loop-condition` | **0 new errors** | re-enabled — 8 directives become meaningful again |
| `no-await-in-loop` | 264 violations | left off |
| `camelcase` | 111 violations | left off |
| `no-use-before-define` | 54 violations | left off |

That the last three are expensive tells us the old gate never enforced them either; they are
noted in `eslint.config.js` as their own piece of work. The remaining 48 directives are removed
and `reportUnusedDisableDirectives` is now `'error'` (not `warn`), so a stale directive fails the
build rather than accumulating.

Two mechanical notes: `--fix` correctly trimmed *compound* directives, keeping the live half
(`no-unused-vars, require-await` → `require-await`); and it left 15 stray blank lines, several
splitting a doc comment from its method, which were removed by mapping the diff back to exact
line numbers.

## 5. `humanhash` replaced by an inlined port

`humanhash@1.0.4` is unmaintained and depends on `uuid@3`, whose advisory has **no fix at any
version** — the last moderate in the audit. The `uuid` dependency is reachable only from
`humanhash`'s `uuid()` method, which this repo never calls; the only thing used was
`humanize()`.

`src/utils/humanHash.ts` ports `humanize()` and its 256-word list. This is a **compatibility-
critical** output: it produces `friendlyName` in the node status response, which operators and
monitoring use to identify a node, so a different algorithm would rename every node in the fleet.

Equivalence was proven before removing the package: **7943 of 7944 cases byte-identical**,
covering 3000 real compressed-secp256k1 public keys, every hex length from 4 to 80 (including odd
lengths, which exercise the original's `/(..?)/g` trailing-character quirk), all 256 byte values,
and every word-count and separator combination. The single difference is the empty-string input,
where both implementations throw and only the message differs (`Cannot read properties of null`
→ `Fewer input bytes than requested output`); `publicKeyHex` is never empty.

Confirmed end-to-end on a running node: the same key that reported
`friendlyName: mexico-high-tennessee-gee` with the original package still reports
`mexico-high-tennessee-gee`.

`src/test/unit/humanHash.test.ts` (7 tests) pins golden values captured from `humanhash@1.0.4`,
including the odd-length and uppercase cases, so a future refactor cannot silently rename nodes.
`uuid` is now absent from the tree entirely.

## 6. TypeScript 5.9 → 6.0.3

`typescript-eslint@8.67` peers at `>=4.8.4 <6.1.0`, so 6.0.3 is in range and **7.0.2 is not** —
one of two reasons TS 7 is blocked. #2137 also deliberately stopped at 6.0.3.

The headline change: **TS 6 defaults `strict` to true**, and this tsconfig never set it.

```
strict on (TS 6 default)                1716 errors
  − strictNullChecks: false              392
  − useUnknownInCatchVariables: false      3   ← landed here
```

Turning off exactly two flags reaches 3 errors while **gaining** `strictFunctionTypes`,
`strictBindCallApply`, `noImplicitThis` and `alwaysStrict` — none of which were active before.
The alternative, a blanket `strict: false`, would have been a 2-line diff with zero safety gain.

The 3 errors were all genuine improvements:

- two `this: OceanP2P` annotations in `handleProtocolCommands`, making the `.bind(this)` contract
  explicit rather than implicit
- one `this: MochaContext` in `src/test/utils/hooks.ts`

Both deferred flags are documented in `tsconfig.json` with their error counts, as is the other
TS 7 blocker: `moduleResolution: node10` is deprecated and stops working in TS 7.
`ignoreDeprecations` moved `"5.0"` → `"6.0"` to defer it; the real migration to `nodenext` costs
~123 errors (~31 relative imports still missing their `.js` extension, plus
`@oceanprotocol/ddo-js` not exposing types under a modern resolver).

## 7. Express 4 → 5

Smaller than expected: no bare `*` wildcards in routes, no `res.send(<number>)`, no
`req.param()`, no `req.query` mutation, no `app.del`, no `res.redirect('back')`. Body parsers are
attached per-route and pass their options explicitly, so Express 5's changed
`urlencoded({extended})` default does not apply.

**Three real changes.**

**a. `path-to-regexp` 8 dropped `:param?`.** Two routes in `aquarius.ts` migrated to brace
groups:

```diff
- `${AQUARIUS_API_BASE_PATH}/assets/ddo/:did/:force?`
+ `${AQUARIUS_API_BASE_PATH}/assets/ddo/:did{/:force}`
```

**b. `express.static.mime` was removed.** `src/index.ts` called
`express.static.mime.define({ 'image/svg+xml': ['svg'] })` — configuring MIME types for a static
file server this app never mounts (there is no `express.static(...)` or `sendFile` anywhere), and
Express 5's `mime-types` already maps `.svg` → `image/svg+xml`. Deleted as dead code.

**c. The one that would have failed silently.** Express 4's `express.json()` left
`req.body = {}` when there was nothing to parse; Express 5 leaves it **`undefined`**. This repo
has 7 `const {…} = req.body` destructures and dozens of `req.body.x` reads — including a GET
route (`/api/admin/config`) that destructures a body — and every one throws a `TypeError` on
`undefined`, converting clean 400s into 500s.

Fixed with one app-level normalizer rather than 100+ call-site edits:

```ts
if (req.body === undefined) {
  req.body = {}
}
```

This is safe *and verified against `body-parser@2.3.0` source*, which is the crux: body-parser
skips only on `onFinished.isFinished(req)` — never on `req.body` already being set — its
`if (!('body' in req)) req.body = undefined` reset is skipped when the property is present, and
`read.js:162` assigns unconditionally after a successful parse. Real bodies still parse and
overwrite the `{}`.

Also: `accessList.ts` now names its route params explicitly, because a bare `Request` in
`@types/express` 5 types `req.params` values as `string | string[]` (path-to-regexp 8 allows
repeats).

### Verified against a running node

Unit tests do not exercise HTTP, so the node was booted and the routes exercised directly:

| check | result |
| --- | --- |
| `POST /directCommand` `{"command":"status"}` | full status returned — bodies parse |
| `POST /api/services/auth/token` with a body | `"nonce: 1 is not a valid nonce"` — destructured fields were read |
| `POST /api/services/decrypt` with a partial body | names the missing field — property reads work |
| `GET /assets/ddo/:did` and `/assets/ddo/:did/true` | both match and reach the handler |
| `GET /assets/ddo/notadid` | **400** — route matched *and* handler validated |
| bodyless `GET /api/admin/config`, `POST /directCommand`, `/logs`, `/auth/token` | **400, not 500** |
| `//api//services//nonce`, `/api/services/nonce/` | 200 — `removeExtraSlashes` still works |
| unmatched path | clean 404 |

Every error in the server log was database-absence in the DB-less test config (`dbType`,
`retrieve`, `searchByWallet`). Zero `req.body` TypeErrors, zero path-to-regexp failures.

### `/getP2PPeers` and siblings returned 500 when P2P is disabled

Three of the four P2P handlers dereferenced `getP2PNode()` without a null check, so with
`hasP2P: false` they threw and the catch-all reported
`500 "Unknown error: Cannot read properties of null (reading 'getAllPeerStore')"`.

`GetP2PNetworkStatsHandler` in the same file already had the correct guard, returning
`503 "P2P Interface is disabled"` — that pattern is now applied consistently to
`GetP2PPeersHandler`, `GetP2PPeerHandler` and `FindPeerHandler`. Fixed in the **handlers**, not
the routes, so the P2P and `POST /directCommand` paths are covered too.

Confirmed on a running node with `hasP2P: false`:

```
GET /getP2PPeers                       503  P2P Interface is disabled   (was 500)
GET /getP2PPeer?peerId=…               503  P2P Interface is disabled   (was 500)
GET /findPeer?peerId=…                 503  P2P Interface is disabled   (was 500)
GET /getP2pNetworkStats                400  Not enabled or unavailable  (unchanged)
```

### `refreshServiceLocks` test was load-sensitive

`serviceJobsDatabase.test.ts` aged a lock stamp by 150 ms and then asserted a second process
could not steal it using a 100 ms staleness window — leaving only ~100 ms for
`refreshServiceLocks` + `acquireServiceLock`. It failed once on a loaded machine, and with
`bail: true` in `.mocharc.json` a flake there **truncates the whole suite** (that run reported
409 tests instead of 525).

The window is now 500 ms with the aging scaled off it, a 5× margin. It also asserts afterwards
that the refresh-to-acquire gap actually fell inside the window, so a machine slow enough to
invalidate the premise reports *"machine too slow to exercise the refresh window"* rather than a
false pass or a confusing failure. 6 consecutive runs pass.

## 9. `postinstall` patch removed

`scripts/fix-libp2p-http-utils.js` rewrote `@libp2p/http-utils` to default a missing URL port to
443/80. On a completely fresh `node_modules` it printed "Already patched", which should be
impossible — so every published tarball was checked: upstream shipped the fix in **2.0.3**
(absent through 2.0.2, present 2.0.3–2.0.6) as
`port === '' ? getDefaultPort(protocol) : parseInt(port, 10)`, where `getDefaultPort` is a
superset of the patch (same 443/80 for https/http, plus `wss:`/`ws:`). All three consumers
request `^2.0.0`, so npm always resolves the newest 2.x.

The script, the `postinstall` hook, and the CLAUDE.md line saying not to remove it are all gone.
`npm install` is now plain, which also affects `RUN npm ci` in the Dockerfile.

## 10. Notable no-ops

These majors needed **no** source changes: `uint8arrays` 4→6, `multiformats`→14, `chai` 4→6
(all 48 test files already used named `{ expect, assert }` imports), `sinon` 19→22, `node-cron`
3→4, `dockerode` 4→5, `eth-crypto` 2→4, `dotenv` 16→17, `base58-js` 2→3, `basic-ftp` 5→6,
`winston-daily-rotate-file` 4→5, `koffi` 2→3, `@types/node` 25→26.

`multiaddr` 13 did need one change: it removed `nodeAddress()`. `OceanP2P.shouldAnnounce()` now
takes the host from the leading `ip4`/`ip6`/`dns*` component. One behavioural subtlety was
preserved deliberately — `nodeAddress()` used to *throw* on a circuit-relay address, and the
surrounding `try/catch` turned that into `return true`. Reading a missing component instead
yields `undefined`, which would have fallen through to `ipaddr.isValid('')` and silently flipped
the answer to `false`, **suppressing relay address announcements**. The no-host case now returns
`true` explicitly.

`zod` 4 needed two small fixes: `ZodError.errors` → `.issues`, and 2 of the 10 `z.record()` call
sites needed an explicit key schema (the other 8 already passed two arguments).

## Testing

```
532 passing, 1 failing        (was 525 passing, 1 failing)
```

The `+7` are the new `humanHash` tests. **The 1 failure is pre-existing** — an
`"after all" hook … Invalid PRIVATE_KEY` teardown hook that fails identically on unmodified
`HEAD`. This was confirmed by installing the *old* dependencies in a separate worktree and
running the same suite, which produced the same single failure.

`npx eslint .` reports 0 errors and 34 warnings, against a measured pre-change baseline of 0
errors and 40 warnings. All 34 remaining are pre-existing
`security/detect-non-literal-fs-filename` and `prefer-destructuring` warnings in unrelated files.

## Deliberately held

| package | held at | why |
| --- | --- | --- |
| `@elastic/elasticsearch` | 8.19.2 | client 9 requires an Elasticsearch **9 server**. `elasticsearch-compose.yml` still pins `elasticsearch:8.5.1`, and every operator on that backend would have to migrate their cluster. Server first, client second, release-noted. |
| `typescript` | 6.0.3 | `typescript-eslint@8.67` peers `<6.1.0`; TS 7 also requires the `node10` → `nodenext` resolver migration first. #2137 stopped here too. |
| `mocha` / `diff` | 11.8.0 | 11.8.0 is the latest release and still carries the low `diff` advisory. Nothing to upgrade to. |

## What still needs verifying

1. **`eciesjs` 0.4.18 → 0.5.0 interop.** The only change with cross-repo blast radius, and it
   matches #2137. 0.5.0 always returns `Uint8Array` instead of `Buffer`, and
   `symEncrypt`/`symDecrypt` take an explicit `Config` rather than reading the global. The default
   config and wire format are unchanged and the unit suite passes, but an encrypt/decrypt
   round-trip against a live ocean.js `9.0.0-next` client is the only real proof. **Do this before
   merging.**

## Known follow-ups, not in this PR

- **Type-safety debt**, documented in `tsconfig.json`: `strictNullChecks` (~1324 errors) and
  `useUnknownInCatchVariables` (~389). Worth doing subsystem by subsystem.
- **Lint rules the old gate never enforced either**, noted in `eslint.config.js`:
  `no-await-in-loop` (264), `camelcase` (111), `no-use-before-define` (54).
- **Before TS 7:** the `nodenext` resolver migration (~123 errors, mostly missing `.js`
  extensions) — and `typescript-eslint` must support TS 7 first.
- **A second pre-existing bug found while testing, not fixed here:** `config.json` with
  `dbConfig: null` fails schema validation (`expected object, received null`) instead of simply
  disabling the metadata DB. Omitting the key works; setting it to `null` does not.
- **`lodash` is *not* removable**, contrary to an early read of this codebase. It has four call
  sites, and only the two `cloneDeep` ones in `P2P/index.ts` are trivial. The other two are
  load-bearing: `lodash.set` in `utils/config/builder.ts:36` is the entire env→config mapping
  (nested paths from dotted strings), and `lodash.merge` at line 239 decides env-var-vs-
  `config.json` precedence for the whole node config, including `ALLOWED_ADMINS` and
  `AUTHORIZED_DECRYPTERS`. It is already at the latest 4.18.1 with no advisory.

## Note on the lockfile

`package-lock.json` was regenerated from scratch: the stale lock pinned `eslint@8` and blocked
resolution outright (`ERESOLVE`). Expect a large diff there.

## Summary by CodeRabbit

* **New Features**
  * Added human-readable node naming without requiring an external naming utility.
  * Added clearer handling for unavailable P2P functionality, returning HTTP 503 with an explanatory message.

* **Bug Fixes**
  * Improved request handling when no request body is provided.
  * Strengthened configuration validation for environment variables and additional Docker files.
  * Updated route handling and network address processing for improved compatibility.

* **Refactor**
  * Modernized linting, type checking, and project tooling.
  * Removed obsolete workarounds and tightened automated test coverage.


alexcos20 (commented):

Overall risk: low

Summary:
This is an exemplary pull request that correctly executes a major dependency upgrade (ESLint flat config, TypeScript 6, Express 5, Zod 4, Libp2p, and Node 22+) while meticulously resolving breaking API changes and ensuring behavioral parity. The removal of the outdated `humanhash` dependency is a great security and maintenance win.

Comments:
• [INFO][style] Excellent work migrating to the new ESLint Flat Config. Keeping the severity rules exactly as they resolved previously ensures this dependency update doesn't accidentally snowball into a linting rewrite.
• [INFO][security] Inlining `humanizeHex` to drop the old vulnerable `uuid@3` dependency is a great architectural choice. Supplying a deterministic test suite against the old outputs ensures zero backwards-compatibility drift.
• [INFO][bug] Good catch on the Express 5 `req.body` change! Pre-seeding it as an empty object will prevent a lot of `TypeError`s in the route handlers.
• [INFO][bug] Updating the Express path params to use the strict regex matching `/:did{/:force}` properly handles the Express 5 route parser update.
• [INFO][other] Enabling `strict: true` while explicitly suppressing the remaining rules is a solid strategy to prevent regressions on the newly enforced checks. Great forward progress for TypeScript 6.

.github/workflows/ci.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
         with:
-          node-version: 'v22.22.2'
```
giurgiur99, line 20: Should we migrate to node 24?

giurgiur99 (approved):

.github/workflows/ci.yml
alexcos20: I will open another PR

PR #1455, opened 2026-08-26
p2p_improve

### LINKED ISSUE #1438: Libp2p stability  enhancements
- [ ] DHT 
- [ ] Compute capabilities announcements

alexcos20:
# P2P: fix silent data loss, bound every budget, make discovery work

## Why

Three independent defects in the same path, each of which returns wrong data rather than an
error:

1. **Truncation read as success.** The frame reader could not distinguish a peer closing cleanly
   between frames from a peer vanishing mid-frame, so a cut-short download or compute result was
   returned with `httpStatus: 200`.
2. **A large body was silently discarded.** `lpStream` took the library's 4 MiB buffer default;
   on overflow the library discards the *entire* buffer and its rejection is a no-op when no read
   is pending — i.e. exactly while a consumer is working. Measured: 10 MB of a 40 MB payload
   delivered, plus corrupt out-of-sequence frames, and in the request direction **no error at
   all**.
3. **Addresses expired 48× too fast.** DHT provider records are valid 48 h; libp2p's peer store
   drops addresses after 1 h, and re-storing an address it already holds does not refresh it
   (`toPeerPB` carries the old `observed` forward). Provider records routinely pointed at peers
   whose addresses had been discarded.

Compute discovery was separately broken: `ram` and `disk` announced **nothing at all** (the
engine publishes GB, the announce loop stepped in bytes), and a resource type the code did not
know — `fpga`, `pcie` — was silently dropped by a `switch` with no `default`.

### Framing and truncation
`LpFrameReader` (now in `handleProtocolCommands.ts`, shared with the client) counts transport
bytes against bytes accounted for by complete frames. EOF with nothing pending is clean;
anything else propagates. The responder tracks whether the status frame has been written — after
that point a failure **aborts the stream** instead of writing another complete frame, which the
client would otherwise read as a valid end.

Overflow is fixed by four coordinated limits plus `pause()`/`resume()` backpressure around each
yield: explicit `maxDataLength`, a raised `stream.maxReadBufferLength` (yamux grows its window to
16 MiB while the read buffer stopped at 4 MiB), a resume water mark, and an explicit
`maxBufferSize`. Derivation is in the comments at each constant.

### Peer store lifetime
`peerStore: { maxAddressAge, maxPeerAge }` both **48 h**, env-overridable. This node already uses
a `LevelDatastore`, so the lifetime survives a restart. Note the values are read **once at node
construction** — changing the env vars on a running process has no effect until restart.

### Budgets — `src/components/P2P/timeouts.ts` (new)
Every scattered timeout literal is now one named budget with one coercion rule shared by the zod
schema and the getters, so the validated config and the running code cannot disagree. A
malformed value is **ignored and falls back**, never fatal. Clamped at both ends: below a 50 ms
floor or above the 32-bit timer ceiling falls back (a value over `2^31-1` made `setTimeout` fire
after **1 ms**, aborting every call instantly).

`sendTo` now always creates its own overall deadline for the setup phase (worst case 105 s → 45 s;
the indexer's two sequential calls 210 s → 90 s), and the response body has a separate budget —
previously a peer trickling one frame every 250 ms streamed indefinitely.

### Observability — `src/components/P2P/counters.ts` (new)
`resolve:peerstore-hit` / `dht-hit` / `miss`, and `sendTo:ok` / `fail` by reason, exposed through
the existing network-stats surface. These are the before/after instrument for the peer store
change.

### Compute discovery
A generic announce loop replaces the `switch`: any resource with an integer `max` is announced,
in its own unit, as **bucket values only** (powers of two `<=` max). GPU model is no longer part
of the content string — which deletes a bug where the same object was pushed twice after
mutation, so the plain GPU capability was never announced at all. Deduplication is keyed on the
serialized string rather than object identity, which never matched.

`POST /getProvidersForStrings` translates an exact C2D string to its bucket before lookup, so the
examples published in `docs/API.md` keep resolving.

The serializer is a deliberate second implementation of a format frozen with ocean.js (this repo
cannot import that package). Golden vectors are mirrored in both repos so a format change fails a
test in both instead of silently returning zero providers. Verified byte-identical across 368
combinations including four resource types neither codebase knows.

### Provide storm
All four bulk-provide paths now share one limiter. Previously the two crons had separate bounds,
`storeAndAdvertiseDDOS` created a **per-invocation** limiter (unbounded across concurrent calls),
and four overlapping publishes ran 20 concurrent provides — roughly 400 DHT PUTs at k=20.
Measured peak now exactly 5.

### Correctness fixes worth calling out
- **`p2pAnnounceC2D` had no outer `try`/`catch`** and is invoked fire-and-forget by the
  scheduler. `src/index.ts` installs an `unhandledRejection` handler that calls `process.exit(1)`,
  so any throw in that cron **killed the node**. Guarded at both the job and the scheduler.
- `advertiseString` now returns a boolean and rejects on a failed provide; the crons previously
  reported `N/N` success regardless of how many provides failed.
- Connection teardown after a failed `newStream` used `conn.streams.length === 0` to mean "nobody
  else is using it". The muxer splices closed streams out, so **zero is the normal state of a
  healthy idle connection** — this was closing connections identify, ping and the DHT were using.
  Ownership is now established by snapshotting connection ids before the dial.
- Byte caps on the accumulating stream helpers and the indexer read paths (previously unbounded
  memory on a path a hostile peer can hold open).
- `Auth` called `stream.destroy()` on a bare async iterable, so exceeding its size limit threw
  `TypeError` instead of the intended error.
- Retry is suppressed when a request body has been partially consumed — resuming mid-stream sent
  valid framing with leading chunks missing, i.e. a silently corrupt upload.

## Where to look hardest

- **`handleProtocolCommands.ts`'s `statusWritten` handling.** If a failure after the status frame
  can still write a complete frame, truncation reads as `200` again.
- **The four framing constants.** Three are load-bearing; the resume water mark is currently
  redundant (`maxReadBufferLength` caps any resume flush below `maxBufferSize`) and kept as margin.
- **The shared budget coercion.** Schema and getters must agree; that has drifted twice before.
- **The connection-ownership snapshot.** Two concurrent calls to the same peer that both see an
  empty snapshot will both consider the connection theirs — narrowed, not eliminated. Closing it
  needs a per-peer dial mutex.

## Summary by CodeRabbit

- **New Features**
  - Added configurable P2P timeouts, retries, stream limits, discovery settings, readiness thresholds, and peer-store retention.
  - Improved C2D capability announcements with broader resource support, bucketing, deduplication, and controlled concurrency.
  - Added detailed P2P networking status and statistics.
  - Added persistent storage support for node databases.

- **Bug Fixes**
  - Improved cancellation, deadlines, retries, and provider lookups.
  - Added safeguards against oversized responses and incomplete network frames.
  - Improved error reporting for missing logs and failed streaming responses.
  - Improved remote DDO validation and concurrent provider queries.

- **Documentation**
  - Expanded P2P configuration documentation with defaults and validation behavior.
  - Removed the obsolete pubsub peer-discovery interval setting.


alexcos20 (commented):

Overall risk: low

Summary:
The PR introduces exceptional improvements to the P2P layer, memory safety, and DHT announcement mechanisms. It resolves critical stream buffering issues, introduces global backpressure, and fixes potential OOM vectors when parsing unbounded streams. Excellent use of `p-limit` and `AbortController`. LGTM!

Comments:
• [INFO][security] Excellent custom `LpFrameReader` implementation. Properly tracking frame boundaries and managing backpressure prevents malicious or slow peers from exhausting memory via infinite varints or unbounded backlog attacks.
• [INFO][performance] Great architectural decision to use a module-level `p-limit` for global concurrency control of DHT provide operations. This effectively caps the fan-out and protects the connection manager from stream exhaustion.
• [INFO][security] Good enforcement of `MAX_ACCUMULATED_STREAM_BYTES` in stream readers. Adding a hard boundary and safely destroying streams prevents Out-Of-Memory (OOM) vulnerabilities when parsing untrusted payloads.
• [INFO][style] Constructing the JSON string manually (`{"c2d":{"free":...}}`) to enforce deterministic field order for hashing is a smart and highly robust approach to avoid hash drift.

alexcos20:
### DHT mode, fan-out and dial budgets

**`clientMode: false` was suppressing the very behaviour its comment claimed.** kad-dht registers
its auto-switch listener only `if (init.clientMode == null)` — passing the option *at all*, even
`false`, prevents the node ever re-evaluating its own reachability. The hardcoded value is
removed; `P2P_DHT_FORCE_SERVER` (default `false`) restores it for operators who know they are
reachable. Demonstrated with two real nodes: with the option omitted, giving the node a public
address flips `client → server` and taking it away flips it back; with `clientMode: false` the
mode never moves.

Mode transitions are now logged in both directions (kad-dht emits no event, so `setMode` is
wrapped) and `getNetworkingStats()` reports the live mode.

`dhtFilter` now defaults to `filterPrivate`, matching kad-dht's own upstream default, with
`announcePrivateIp === true` forcing `passthroughMapper` back for local and test networks.

`disjointPaths: 4` added. Undeclared, it defaults to `alpha` (10), so a single query could put
**10 × 10 = 100** concurrent lookups in flight against far fewer dial slots; this caps it at 40.
`kBucketSize` untouched — it is simultaneously the provider-record replication factor and the
per-provide write amplification.

Dial budgets: `connectionsMaxParallelDials` 15 → 50, `connectionsDialTimeout` 30 s → 15 s,
`maxPeerAddrsToDial` 5 → 30, and `maxDialQueueLength` exposed at libp2p's own default of 500. The
address cap mattered: a single bootstrap peer now has ~6 addresses and the budget is spent
*before* the transport check, so even instantly-skipped addresses counted against it. Both
`config.json` files hardcoded the old values and would otherwise have silently defeated the
schema change.

Bootstrap peers are now tagged `keep-alive-bootstrap` with no TTL. libp2p's reconnect queue only
redials tags starting with `keep-alive`; the plain `bootstrap` tag shields a peer from trimming
but **never reconnects it**. A finite TTL would be worse than none — the tag expires, the peer
becomes prunable, and it is never redialled.

alexcos20:
@coderabbitai full review

src/components/P2P/timeouts.ts
```diff
@@ -0,0 +1,590 @@
...
+ * crippling a budget - an unbounded DHT query, and a 1ms one, are both failure modes these
+ * constants exist to stop.
+ */
+function envPositiveNumber(
+  name: string,
+  fallback: number,
+  max?: number,
+  min?: number
+): number {
+  return normalizeP2pBudget(process.env[name], max, min) ?? fallback
```
giurgiur99: This reads only .envs, should we read from config.json?

src/components/P2P/index.ts
```diff
@@ -801,110 +1267,604 @@ export class OceanP2P extends EventEmitter {
...
+              countSendToFailure(SENDTO_FAIL_REASONS.stream)
+              return {
+                status: { httpStatus: 500, error: `P2P error: ${failure.message}` }
+              }
+            }
+            P2P_LOGGER.warn(
+              `Retrying ${peerId} after a ${failure.name} error: ${failure.message}`
+            )
+            try {
+              await connection.close()
```
giurgiur99: Shouldn't we   ```await releaseUnusedConnection()``` here too?

src/components/P2P/index.ts
```diff
@@ -923,83 +1883,100 @@ export class OceanP2P extends EventEmitter {
...
-          'Could not find any Ocean peers. Nobody is listening at the moment, skipping...'
-        )
-        // save it for retry later
-        // https://github.com/libp2p/js-libp2p-kad-dht/issues/98
-        if (!this._pendingAdvertise.includes(did)) {
-          this._pendingAdvertise.push(did)
-        }
+      const dht = (this._libp2p.services as Record<string, any> | undefined)?.dht as
+        { routingTable?: { size?: number } } | undefined
+      const routingTableSize = dht?.routingTable?.size
```
dnsi0: `getDhtRoutingTableSize` could be used to establish the peers in DHT routing table

src/components/P2P/index.ts
```diff
@@ -923,83 +1883,100 @@ export class OceanP2P extends EventEmitter {
...
+      if (typeof routingTableSize === 'number' && routingTableSize > 0) {
```
dnsi0: Would it make sense to return false when `routingTableSize = 0`? If we rely on getConnections, it wouldn't work anyway when `allowQueryWithZeroPeers: false`

src/components/core/handler/ddoHandler.ts
```diff
@@ -48,10 +48,90 @@ import { Asset, DDO, DDOManager } from '@oceanprotocol/ddo-js'
...
+  const now = Date.now()
+  for (const [id, expiresAt] of notFoundDdos) {
+    if (expiresAt <= now) {
+      notFoundDdos.delete(id)
+    }
+  }
+}
+
+/** True when `id` was searched for recently and found nowhere. Prunes as it reads. */
+function isDdoKnownMissing(id: string): boolean {
```
dnsi0, line 97: A ddo id only if it is queried again, if it is not queried again, it will be kept forever. Is this the desired result?

alexcos20: pruneExpiredNotFoundDdos will clear it from the list

src/components/P2P/timeouts.ts
alexcos20: see https://github.com/oceanprotocol/ocean-node/pull/1455/commits/ad1cf1a5c2269a75241b1d5e95d602cbeb07c9c4, thanks

PR #1458, opened 2026-08-27
Fix: indexer silently skipping DDOs on transient provider errors

### LINKED ISSUE #1459: FRE & Dispenser tests failing from time to time
```
2026-08-27T09:21:12.4178754Z
2026-08-27T09:21:12.4178869Z   320 passing (15m)
2026-08-27T09:21:12.4179091Z   1 failing
2026-08-27T09:21:12.4179217Z
2026-08-27T09:21:12.4179481Z   1) **********         Publish pricing scehmas and assert ddo stats - FRE & Dispenser
2026-08-27T09:21:12.4179918Z        should get stats for fre:
2026-08-27T09:21:12.4180363Z      TypeError: Cannot read properties of undefined (reading 'indexedMetadata')
2026-08-27T09:21:12.4181332Z       at Context.<anonymous> (file:///home/runner/work/ocean-node/ocean-node/dist/test/integration/pricing.test.js:159:28)
2026-08-27T09:21:12.4182152Z       at callFn (/home/runner/work/ocean-node/ocean-node/node_modules/mocha/lib/runnable.js:366:21)
2026-08-27T09:21:12.4182879Z       at Runnable.run (/home/runner/work/ocean-node/ocean-node/node_modules/mocha/lib/runnable.js:354:5)
2026-08-27T09:21:12.4183634Z       at Runner.runTest (/home/runner/work/ocean-node/ocean-node/node_modules/mocha/lib/runner.js:809:10)
2026-08-27T09:21:12.4184317Z       at /home/runner/work/ocean-node/ocean-node/node_modules/mocha/lib/runner.js:959:12
2026-08-27T09:21:12.4184963Z       at next (/home/runner/work/ocean-node/ocean-node/node_modules/mocha/lib/runner.js:724:14)
2026-08-27T09:21:12.4185600Z       at /home/runner/work/ocean-node/ocean-node/node_modules/mocha/lib/runner.js:734:7
2026-08-27T09:21:12.4186583Z       at next (/home/runner/work/ocean-node/ocean-node/node_modules/mocha/lib/runner.js:595:14)
2026-08-27T09:21:12.4187961Z       at Immediate.<anonymous> (/home/runner/work/ocean-node/ocean-node/node_modules/mocha/lib/runner.js:702:5)
2026-08-27T09:21:12.4189132Z       at process.processImmediate (node:internal/timers:534:21)
2026-08-27T09:21:12.4189627Z
2026-08-27T09:21:12.4189636Z
2026-08-27T09:21:12.4189644Z
2026-08-27T09:21:12.5262830Z ----------|---------|----------|---------|---------|-------------------
2026-08-27T09:21:12.5265700Z File      | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s
2026-08-27T09:21:12.5266309Z ----------|---------|----------|---------|---------|-------------------
2026-08-27T09:21:12.5269190Z All files |     100 |       75 |     100 |     100 |
2026-08-27T09:21:12.5271772Z  hooks.ts |     100 |       75 |     100 |     100 | 46
2026-08-27T09:21:12.5272358Z ----------|---------|----------|---------|---------|-------------------
```

alexcos20:
# Fix: indexer silently skipping DDOs on transient provider errors

## Problem

Under transient RPC conditions (provider failover, node restart, socket timeout, or an
overlapping indexer teardown destroying a provider mid-crawl), the indexer could **silently drop
an asset and advance its cursor past it** — the DDO would never be indexed and would only reappear
after a manual chain reindex.

The mechanism:

1. Every event processor wraps its work in a `try/catch` that logs and returns `null`/`undefined`.
   A `provider destroyed` / network error inside `processEvent` was therefore indistinguishable, to
   the caller, from a legitimate "nothing to do".
2. `processChunkLogs` / `processBlocks` never threw, `processBlocks` returned
   `lastBlock = lastIndexedBlock + count`, and `updateLastIndexedBlockNumber` **advanced the cursor**
   past the un-stored asset.
3. The crawl loop's existing `catch` (which correctly holds the cursor, sleeps, and retries) was
   never reached, because the error was swallowed one level deeper.

This surfaced as an intermittent integration-test failure (`should get stats for fre` in
`pricing.test.ts`) on the Node 24 / ethers 6.17 timing change, where a `provider destroyed;
cancelled request` during a crawl caused the DDO to be lost — but it is a genuine production
robustness bug independent of the tests.

## Approach

Two complementary fixes:

1. **Make the indexer self-healing** so it is correct under _any_ transient provider/RPC error,
   from whatever source. The guiding rule: **a transient provider/RPC error must fail the chunk
   (retry, cursor held); a legitimate business result — including a genuine `null` — must be left
   exactly as before.** A provider error is now *distinguished* from a business skip and rethrown so
   it lands in the crawl loop's existing retry path, instead of being swallowed.
2. **Close the overlap at its source** so the specific race that exposed the bug — a dying
   indexer's provider being destroyed while a crawl is still using it — can no longer happen: a
   newly-created indexer waits for the previous `OceanNode` instance's teardown to finish before it
   starts crawling.

The self-healing layer (1) keeps `OceanNode.getInstance`'s synchronous signature untouched and
makes overlap _safe_; the teardown-await (2) makes overlap _not happen_. Together they are
defence-in-depth.

### 1. Error classifier — `src/components/Indexer/utils.ts`

- New `isProviderError(err)` — narrowly matches transient ethers failures only: codes
  `NETWORK_ERROR`, `SERVER_ERROR`, `TIMEOUT`, `UNSUPPORTED_OPERATION` with a `provider destroyed`
  message, plus the `provider destroyed` / `cancelled request` message fragments.
- Guards the **8 contract-call catches** in this file (`isValidFreContract`,
  `isValidDispenserContract`, `getPricesByDt` ×2, `getPricingStatsForDddo` ×4) with
  `if (isProviderError(e)) throw e` as the first statement — otherwise a provider death mid-pricing
  would save a DDO with silently-empty `stats` and still advance the cursor.

### 2. Shared guard — `src/components/Indexer/processors/BaseProcessor.ts`

- New `protected rethrowIfProviderError(err)` helper (imports `isProviderError`).
- Called as the first statement of the top-level `catch` in **all 15 event processors**
  (`MetadataEventProcessor`, the four `Exchange*`, three `Dispenser*`, `Order{Started,Reused}`,
  `MetadataState`, `Address{Added,Removed}`, `NewAccessList`, `Escrow`). In
  `MetadataEventProcessor` the guard is placed **before** the `ddoState.update(..., false, ...)`
  call, so a transient blip is not persisted as a permanent DDO failure.

### 3. Self-healing crawl loop — `src/components/Indexer/ChainIndexer.ts`

- `provider` / `signer` changed from `const` to `let` and **re-acquired at the top of each loop
  iteration**. If the provider was destroyed by an overlapping teardown, `getProvider()` recreates
  it and `getSigner()` rebuilds against it; when healthy, both return the cached instances at
  negligible cost.
- Rethrows provider errors from the `retrieveChunkEvents` catch, so a failed `eth_getLogs` doesn't
  fall through to `processBlocks([])` and advance the cursor over an empty result. Propagated errors
  land in the loop's outer `catch`, which sleeps `interval` and retries the same range.

### 4. Reset cached signer on stop — `src/utils/blockchain.ts`

- `Blockchain.stop()` now clears the cached `signer` after destroying the provider. The signer is
  bound to the destroyed provider and `getSigner()` only rebuilds when it is falsy.

### 5. Close the overlap window — `src/OceanNode.ts` + `src/components/Indexer/index.ts`

The root cause of the exposing race: `getInstance(..., newInstance=true)` fired the previous
instance's `tearDownAll()` **without awaiting it**, while a freshly-constructed `OceanIndexer`
auto-starts crawling from its constructor — so a dying indexer briefly ran alongside the new one on
the same DB/chain, and the teardown's `provider.destroy()` could land mid-crawl.

- `OceanNode` now **retains the teardown promise** in a private static `pendingTeardown` (set in
  `getInstance` where the old instance is torn down) and exposes
  `static async awaitPendingTeardown()`, which awaits it once and clears it (guarding against a
  newer teardown replacing it while awaiting). The `getInstance` signature is unchanged — it still
  returns synchronously and does not await.
- `OceanIndexer.startAllChainIndexers()` **awaits `OceanNode.awaitPendingTeardown()`** as its first
  step, before `checkAndTriggerReindexing()` and before starting any `ChainIndexer`. Since
  `getInstance` (which sets the promise) always runs before the new `OceanIndexer` is constructed,
  the new indexer cannot begin crawling until the previous instance's indexer has stopped and its
  provider has been destroyed.

This introduces an import edge `Indexer/index.ts → OceanNode.ts` (on top of the existing
`OceanNode.ts → Indexer/index.ts`). The cycle is safe: `OceanNode` is referenced only inside a
method body at runtime, never at module-evaluation time — verified by loading the built module
graph in both import orders.

## Why it's safe

- **Idempotent reprocessing.** `createOrUpdateDDO` upserts via `ddoDatabase.update`, and
  `MetadataEventProcessor` short-circuits same-txid reprocessing — no duplication/corruption on retry.
- **No new unbounded loop.** Recovery replaces a dead provider on the next iteration; a genuinely
  down RPC now stalls that chain (retry every `interval`) instead of dropping assets, bounded by the
  existing `catch → sleep(interval)`.
- **Narrow classifier.** Only the listed ethers codes / message fragments count; every other path
  keeps its previous log-and-continue behaviour, so no legitimate business outcome (including a
  genuine `null`, e.g. an exchange not approved by the Router) changes.
- **No deadlock from the teardown-await.** `awaitPendingTeardown()` waits on the _old_ instance's
  teardown (stops the _old_ indexer, destroys the _old_ provider); it never depends on the new
  indexer, and an already-resolved promise is awaited instantly. On the DB-reconnect path there is
  no pending teardown, so it is a no-op.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved indexing reliability during temporary blockchain provider or network failures.
  * Indexing now retries affected block ranges instead of skipping data when provider errors occur.
  * Prevented stale blockchain connections from being reused after shutdown.
  * Ensured previous indexing activity fully stops before a new indexing run begins.
* **Reliability**
  * Improved recovery from timeouts, cancelled requests, unavailable providers, and other transient connection issues.

### DIFF
--- a/src/OceanNode.ts
+++ b/src/OceanNode.ts
@@ -114,16 +114,28 @@ export class OceanNode {
         if (!blockchainRegistry)
           blockchainRegistry = new BlockchainRegistry(keyManager, config)
       }
-      // teardown old instance if needed; retain the promise so a newly-created
-      // OceanIndexer can await it before crawling (see awaitPendingTeardown).
-      OceanNode.pendingTeardown =
-        this.instance?.tearDownAll().catch((err: unknown) => {
-          OCEAN_NODE_LOGGER.warn(
-            `Failed to tear down previous OceanNode instance: ${
-              err instanceof Error ? err.message : String(err)
-            }`
-          )
-        }) ?? null
+      // Teardown the old instance if needed, and retain the promise so a newly-created
+      // OceanIndexer can await it before crawling (see awaitPendingTeardown). Chain onto any
+      // still-pending teardown from an earlier replacement instead of overwriting it, so the
+      // barrier keeps covering every outstanding teardown, not just the most recent one.
+      const previousTeardown = OceanNode.pendingTeardown
+      const instanceToTearDown = this.instance
+      OceanNode.pendingTeardown = (async () => {
+        if (previousTeardown) {
+          await previousTeardown
+        }
+        if (instanceToTearDown) {
+          try {
+            await instanceToTearDown.tearDownAll()
+          } catch (err: unknown) {
+            OCEAN_NODE_LOGGER.warn(
+              `Failed to tear down previous OceanNode instance: ${
+                err instanceof Error ? err.message : String(err)
+              }`
+            )
+          }
+        }
+      })()
       OCEAN_NODE_LOGGER.debug('Creating new OceanNode instance')
       this.instance = new OceanNode(
         config,
--- a/src/OceanNode.ts
+++ b/src/OceanNode.ts
@@ -36,6 +36,10 @@ export interface RequestDataCheck {
 }
 export class OceanNode {
   private static instance: OceanNode
+  // In-flight teardown of the previous instance (set by getInstance). A newly-created
+  // OceanIndexer awaits this before crawling, so it never overlaps a dying indexer on
+  // the same DB/chain while that instance's provider is being destroyed.
+  private static pendingTeardown: Promise<void> | null = null
   // handlers
   private coreHandlers: CoreHandlersRegistry
   // compute engines
@@ -110,14 +114,16 @@ export class OceanNode {
         if (!blockchainRegistry)
           blockchainRegistry = new BlockchainRegistry(keyManager, config)
       }
-      // teardown old instance if needed
-      this.instance?.tearDownAll().catch((err: unknown) => {
-        OCEAN_NODE_LOGGER.warn(
-          `Failed to tear down previous OceanNode instance: ${
-            err instanceof Error ? err.message : String(err)
-          }`
-        )
-      })
+      // teardown old instance if needed; retain the promise so a newly-created
+      // OceanIndexer can await it before crawling (see awaitPendingTeardown).
+      OceanNode.pendingTeardown =
+        this.instance?.tearDownAll().catch((err: unknown) => {
+          OCEAN_NODE_LOGGER.warn(
+            `Failed to tear down previous OceanNode instance: ${
+              err instanceof Error ? err.message : String(err)
+            }`
+          )
+        }) ?? null
       OCEAN_NODE_LOGGER.debug('Creating new OceanNode instance')
       this.instance = new OceanNode(
         config,
@@ -132,6 +138,22 @@ export class OceanNode {
       OCEAN_NODE_LOGGER.debug('Return cached OceanNode instance')
     }
     return this.instance
+  }
+
+  /**
+   * Await a previous instance's in-flight teardown (set by getInstance) so callers such as a
+   * freshly-constructed OceanIndexer don't start crawling while the old instance is still
+   * shutting down its indexer and destroying its provider. No-op when nothing is pending.
+   */
+  public static async awaitPendingTeardown(): Promise<void> {
+    const pending = OceanNode.pendingTeardown
+    if (pending) {
+      await pending
+      // Only clear if a newer teardown didn't replace it while we were awaiting.
+      if (OceanNode.pendingTeardown === pending) {
+        OceanNode.pendingTeardown = null
+      }
+    }
   }
 
   // in the future we should remove these 'add' methods as well
--- a/src/components/Indexer/ChainIndexer.ts
+++ b/src/components/Indexer/ChainIndexer.ts
@@ -475,6 +475,13 @@ export class ChainIndexer {
           })
         }
       } catch (error) {
+        if (isProviderError(error)) {
+          // Transient provider failure: preserve the task for a later retry (do not emit
+          // REINDEX_QUEUE_POP) and rethrow so the loop's outer catch holds the cursor,
+          // sleeps, and re-acquires a healthy provider before retrying.
+          this.reindexQueue.unshift(reindexTask)
+          throw error
+        }
         INDEXER_LOGGER.log(
           LOG_LEVELS_STR.LEVEL_ERROR,
           `REINDEX Error for tx ${reindexTask.txId}: ${error.message}`,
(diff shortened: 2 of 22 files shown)


alexcos20 (commented):

Overall risk: low

Summary:
This PR hardens the `OceanIndexer` by introducing synchronization for provider teardowns, preventing race conditions where multiple indexers could concurrently crawl the same chain/DB. It also significantly improves error handling by identifying transient provider/RPC errors (`isProviderError`) and rethrowing them, which ensures the indexer correctly retries instead of swallowing errors and erroneously advancing the block cursor. The changes are robust and substantially increase the resilience of the node.

Comments:
• [INFO][other] The implementation of `awaitPendingTeardown` handles race conditions elegantly by checking `if (OceanNode.pendingTeardown === pending)` before clearing the reference. Good job anticipating overlapping teardowns.
• [INFO][other] The `isProviderError` function is a great heuristic to accurately capture `ethers.js` network, timeout, and destroyed provider errors, keeping business-logic errors cleanly separated.
• [INFO][style] Using `undefined as unknown as Signer` bypasses TypeScript's type safety. While acceptable as a quick fix, it is usually better to update the class property definition to explicitly allow `undefined` or `null` (e.g., `private signer?: Signer` or `private signer: Signer | null`).

```typescript
-      this.signer = undefined as unknown as Signer
+      this.signer = undefined // (assuming property type is updated to allow undefined)
```

PR #1464, opened 2026-09-01
add metadata for services

### LINKED ISSUE #1463: Add metadata for services

alexcos20:
# Feat: user-defined `metadata` on Service-on-Demand jobs

## Problem

Compute jobs already support an arbitrary, node-opaque `metadata` bag
(`DBComputeJobMetadata` — a `{ [key]: string | number | boolean }` map) that the node stores
verbatim, echoes back on status, and never interprets. It's a lightweight annotation channel
for client-side bookkeeping (run names, experiment IDs, correlation tokens, tags).

Service-on-Demand jobs had no equivalent. This PR adds the same field to services, reusing the
compute type, so a consumer can label a service at start, read those labels back on status, and
optionally replace them on restart.

## Behavior (as designed)

- **`SERVICE_START`** accepts an optional `metadata` object — arbitrary keys with scalar values
  (`string` / `number` / `boolean`), capped at **1 KB** when JSON-stringified. Over the cap is
  rejected **fail-fast with a `400`** in the handler's `validate()`, before any escrow/job work.
  The node stores it verbatim and never interprets it.
- **`SERVICE_GET_STATUS`** returns it. This path is **authenticated and owner-scoped**, so
  service metadata is effectively **owner-only** — a stronger privacy posture than compute,
  where status-by-jobId is unauthenticated and the metadata is readable by anyone holding the
  jobId.
- **`SERVICE_LIST`** (node-wide, authenticated but **not** owner-scoped) **keeps** metadata,
  but strips the other config-revealing fields it already removes (`userData`, `dockerCmd`,
  `dockerEntrypoint`, `dockerfile`, `additionalDockerFiles`).
- **`SERVICE_RESTART`**: if `metadata` is supplied it **replaces** the stored bag; if omitted
  the original metadata is **kept**. It is **not** a container param — it's independent of the
  REUSE/RESPEC discriminator, so sending only `metadata` does not force RESPEC mode (i.e. it
  never makes `image` suddenly required).

## Persistence

No SQL migration. A `ServiceJob` is persisted as a whole-object JSON blob in the
`service_jobs.body` column (`newServiceJob` / `updateServiceJob` write `{ ...job }`), so the new
field round-trips losslessly.

### Types

- `src/@types/C2D/ServiceOnDemand.ts` — import `DBComputeJobMetadata` from `./C2D.js`; add
  `metadata?: DBComputeJobMetadata` to `ServiceJob`.
- `src/@types/commands.ts` — add `metadata?: DBComputeJobMetadata` to `ServiceStartCommand`;
  add it to `ServiceRestartCommand` (documented as non-container-param, replace-on-present).

### Shared size guard (dedupe)

- `src/components/c2d/index.ts` — new `validateJobMetadataSize()` + `MAX_JOB_METADATA_SIZE`
  (`1024`), a single source of truth for the metadata cap. `startComputeJob`'s previously inline
  1 KB check was refactored to call it, so the compute and service paths can't drift.

### Engine

- `src/components/c2d/compute_engine_base.ts` — added `metadata` param to the `createServiceJob`
  signature and `newMetadata` to `restartService`.
- `src/components/c2d/compute_engine_docker.ts`:
  - `createServiceJob` — new `metadata` param (placed **before** `outputBucketId`, which stays
    last), validates size, and assigns `metadata` onto the constructed `ServiceJob`.
  - `restartService` — new `newMetadata` param; validates size up front, then applies
    replace-if-present (`if (newMetadata !== undefined) job.metadata = newMetadata`) right before
    the job is persisted as `Restarting`. Independent of the REUSE/RESPEC container grouping.

### Handlers

- `src/components/core/service/startService.ts` — `validate()` rejects an oversized `metadata`
  bag **fail-fast with a `400`** (using the shared `MAX_JOB_METADATA_SIZE`); passes
  `task.metadata` into `createServiceJob`.
- `src/components/core/service/restartService.ts` — `validate()` applies the same `400`
  fail-fast, placed **outside** the RESPEC block so a lone oversized `metadata` is still caught;
  passes `task.metadata` into `restartService`; deliberately **not** added to the RESPEC
  discriminator list. The engine-level `validateJobMetadataSize` remains as defense-in-depth.

### Sanitizers

- `src/components/core/service/utils.ts`:
  - `toPublicServiceJob` — unchanged; metadata flows through via `...rest` (owner status path).
  - `toListedServiceJob` — now keeps `metadata` (added to the `Omit<...>` return type and
    the destructure)

### HTTP routes

- `src/components/httpRoutes/compute.ts` — `serviceStart` and `serviceRestart` routes map
  `req.body.metadata`.

### Tests

- `src/test/unit/service/serviceUtils.test.ts` — `toPublicServiceJob` keeps metadata;
  `toListedServiceJob` strips it (plus null-safety).
- `src/test/unit/service/serviceHandlers.test.ts` — `ServiceStartHandler` forwards metadata to
  `createServiceJob` (before `outputBucketId`) and returns a `400` (no job created) for
  oversized metadata; `ServiceRestartHandler` forwards metadata (arg 10, no RESPEC forced) and
  returns a `400` (engine not called) for oversized metadata.
- `src/test/unit/compute.test.ts` — `restartService` replaces metadata when supplied, keeps the
  original when omitted, and rejects metadata >1 KB.

### Docs

- `docs/API.md` — `serviceStart` request field + example; `serviceStatus` (metadata kept, owner-
  scoped) and `serviceList` (metadata stripped) response notes; `serviceRestart` field with its
  replace/keep semantics.
- `docs/services.md` — metadata described in the intro, and added to the `SERVICE_LIST`
  sanitization note.

## Why it's safe

- **Additive & backwards-compatible.** `metadata` is optional everywhere; existing callers and
  the wire shape are unchanged. Persisted as a JSON blob field, so no schema migration.
- **Single source of truth for the cap.** Compute and service metadata now share
  `validateJobMetadataSize`, so the 1 KB limit and error message can't diverge.
- **Restart stays atomic.** Metadata is orthogonal to the container REUSE/RESPEC grouping, so it
  can't produce a partial container respec.

## Notes

- The 1 KB cap is enforced **fail-fast** in each handler's `validate()` (a `400` before any
  escrow/job work), with `validateJobMetadataSize` in the engine layer kept as
  defense-in-depth. Both use the shared `MAX_JOB_METADATA_SIZE` constant.

## Summary by CodeRabbit

* **New Features**
  * Service start and restart requests support optional user-defined metadata labels.
  * Metadata is preserved across restarts unless replacement metadata is provided.
  * Owners can view metadata in service status responses; it is excluded from node-wide listings.

* **Bug Fixes**
  * Metadata must contain scalar values and fit within 1 KB.
  * Invalid or oversized metadata is rejected with a clear validation error.

* **Documentation**
  * Updated API and service documentation with metadata limits, visibility, validation, and replacement rules.


alexcos20:
@coderabbitai full review

alexcos20 (commented):

Overall risk: low

Summary:
The PR successfully implements the `metadata` feature for service start and restart operations. The architectural design is sound, correctly segregating owner-scoped metadata from the public service list. Data privacy is maintained, and appropriate size safeguards are enforced to prevent storage bloat. A few minor optimizations for code reuse and robustness have been suggested, but the implementation is excellent overall. LGTM!

Comments:
• [INFO][style] Good use of ES6 rest parameter destructuring here to ensure `metadata` (and other configuration fields) are safely and elegantly omitted from the public listing.
• [INFO][other] Using `Object.keys(metadata).length > 0` correctly skips empty objects, but if a primitive type (like a number or boolean) is somehow passed as metadata, `Object.keys()` will return an empty array, bypassing the size check entirely. While numbers and booleans will never exceed 1KB when stringified anyway (meaning no security risk), you can simplify the logic and remove type assumptions by checking for null/undefined directly.

```diff
-  if (metadata && Object.keys(metadata).length > 0) {
+  if (metadata !== undefined && metadata !== null) {
     if (JSON.stringify(metadata).length > MAX_JOB_METADATA_SIZE) {
```
• [INFO][style] There is a slight duplication of the `JSON.stringify` size validation logic here (and in `restartService.ts`) versus the `validateJobMetadataSize` function in `src/components/c2d/index.ts`. To keep this completely DRY, consider extracting a boolean helper (e.g., `isJobMetadataSizeValid(metadata)`) that both the handlers and the engine can use, though the current approach works fine for returning the specific 400 response.

giurgiur99 (approved):

giurgiur99 (commented):
If the metadata retrieval is owner-scoped we can't index it in monitoring

giurgiur99 (approved):

PR #1470, opened 2026-09-04
Bugs/gpu metrics fix

alexcos20:
# Host-wide GPU telemetry (idle GPUs included) + telemetry dashboard/import fixes

## Context

The **Ocean Node — Compute Resources** Grafana dashboard showed **"No data"** on every GPU
panel (utilization, memory, temperature, power), even on nodes that have GPUs. The cause was
scope, not a missing metric: GPU device telemetry was only sampled **for a GPU currently
allocated to a running compute job**. The aggregate's `gpus` array was built from the sampled
running jobs, so an idle box produced nothing — `ocean_compute_gpu_devices_in_use` reported `0`
and none of the per-device series existed. Confirmed live: the GPU-capable nodes advertise
capacity (`ocean_compute_env_resource_total{resource=~"gpu.*"}` = 1 and 8 devices) while
`ocean_compute_jobs_running` / `ocean_compute_gpu_devices_in_use` were `0` — idle, so nothing to
sample.

This PR makes GPU health **host-wide**: every GPU visible to the node process is sampled on the
metrics cadence, idle ones included, so a card's utilization/memory/temperature/power show up
even when no job holds it. A device is tagged `in_use` so "held by a job" stays distinguishable.

While in the telemetry area, it also hardens the shared-Grafana import tooling and the two
existing dashboards against a datasource-resolution bug and an empty-variable failure mode
(details below).

### GPU: sample all host GPUs, not just job-held ones

- **`src/components/c2d/gpu/types.ts`** — added `sampleAll()` to the `GpuVendorCollector`
  contract; added `uuid`/`index` to `GpuDeviceMetrics` (set by enumeration, used to map a device
  back to its configured resource id).
- **`src/components/c2d/gpu/nvml.ts`** — bound the enumeration calls (`nvmlDeviceGetCount_v2`,
  `nvmlDeviceGetHandleByIndex_v2`, `nvmlDeviceGetUUID`), extracted the per-device reads into
  `applyDeviceReadings`, and added `readUuid` + **`sampleAll()`**: enumerate every visible device
  and sample each, carrying its NVML UUID and index. Same best-effort discipline as before — a
  failed read yields nulls for that device and never throws into the C2D loop.
- **`src/components/c2d/gpu/index.ts`** — added **`GpuMetricsService.sampleHost()`** and the pure,
  NVML-free **`mapHostGpuDevices()`** that aligns each enumerated device's UUID to its configured
  resource id (`gpu0`) so host series share the `gpu` label with `ocean_compute_env_resource_*`;
  a device with no configured match falls back to `vendor+index`.
- **`src/components/c2d/compute_engine_docker.ts`** — new `hostGpuSnapshot` field +
  **`refreshHostGpuSnapshot()`**, called from the InternalLoop, throttled to
  `C2D_METRICS_INTERVAL_SECONDS`, strictly best-effort, and a **no-op on nodes that declare no GPU
  resources** (a pure-CPU node never touches NVML).
- **`src/telemetry/computeGauges.ts`** — `ocean_compute_gpu_devices_in_use` now reflects the GPUs
  **allocated to jobs**; the per-device `utilization/memory/temperature/power` series are emitted
  from the **host-wide snapshot** for every device, each carrying `in_use="true|false"` (derived
  from the env resource snapshot). The job-scoped array is kept only as a **dedup fallback** for a
  device the host enumeration could not read (e.g. NVML unavailable), keyed by resource id so a
  device is never emitted twice per tick.
- **`src/test/unit/c2d/gpuHostMetrics.test.ts`** *(new)* — unit-tests `mapHostGpuDevices`: UUID →
  resource-id mapping, both fallbacks (unmatched UUID, no UUID), non-GPU-resource filtering, and
  the null-vs-number preservation (idle `0%` is kept, not treated as "missing").
- **`docs/env.md`**, **`docs/compute.md`** — document the host-wide export, the `in_use`
  semantics, and that it honours the same `GPU_METRICS` switch and NVML requirements.

### Telemetry dashboards + import tooling

- **`deploy/telemetry/scripts/import-dashboard.sh`** — replaced the naive
  `datasources.find(d => d.type === 'prometheus')` (which blindly took the first Prometheus
  datasource — wrong on a Grafana with several, silently blanking every panel) with a resolver
  that goes `PROM_UID` → exact `PROM_NAME` (case-insensitive) → the org default (`isDefault`) →
  the sole candidate → **fail loudly listing candidates**. Also normalizes a trailing slash on
  `GRAFANA_URL` and documents the precedence in the header.
- **`deploy/telemetry/grafana/dashboards/ocean-node-p2p.json`**,
  **`ocean-node-compute.json`** — added `"allValue": ".*"` to the `instance`/`role` template
  variables so "All" always interpolates to a match-everything regex; a momentarily-empty
  `label_values(...)` can no longer collapse a panel to `label=~""` (No data).
- **`ocean-node-compute.json`** GPU panels — surfaced `in_use` in the per-device legends, and made
  the **"GPU memory used ratio"** panel compute the ratio inline
  (`sum(...gpu_memory_used_bytes) / sum(...gpu_memory_total_bytes)`) so it no longer depends on
  the `ocean_node:compute_gpu_mem_used_ratio` recording rule being loaded on the target Prometheus.

## Metric surface (labels)

- **Now emitted for every visible GPU, idle included:**
  `ocean_compute_gpu_{utilization_percent,memory_used_bytes,memory_total_bytes,temperature_celsius,power_watts}`
  with labels `gpu` (configured resource id, else `vendor<index>`), `vendor`, and `in_use`.
- `ocean_compute_gpu_devices_in_use` — unchanged meaning: count of GPUs allocated to running jobs.
- No new instruments; the existing observable gauges in `src/telemetry/metrics.ts` are reused.

## Operational notes

- The node process must have **NVML visibility** to enumerate host GPUs — `NVIDIA_VISIBLE_DEVICES`,
  the driver capabilities, plus the optional `koffi` dependency and `libnvidia-ml.so.1` reachable
  in the container. This is the same requirement as the pre-existing per-job GPU metrics; a
  containerized node does not inherit the host driver libraries automatically. Documented in
  `docs/env.md` / `docs/compute.md`.
- **Dashboard re-import to the shared Grafana is intentionally deferred**: the host-wide GPU series
  only exist once nodes run this build, so importing the updated compute dashboard beforehand would
  still show "No data". Re-import with
  `PROM_NAME=Prometheus ./deploy/telemetry/scripts/import-dashboard.sh compute` after deploy.

## Summary by CodeRabbit

* **New Features**
  * GPU health metrics now cover every GPU visible to the node, including idle devices, with an `in_use` indicator.
  * Added host-wide GPU utilization, memory, temperature, and power telemetry.
  * Dashboard GPU panels now show device usage status and calculate memory ratios directly.

* **Bug Fixes**
  * Improved dashboard imports by reliably resolving the intended Prometheus datasource.
  * “All” dashboard filters now match every instance and role.

* **Documentation**
  * Expanded GPU telemetry documentation, including sampling behavior and device allocation indicators.


alexcos20 (commented):

Overall risk: low

Summary:
The PR introduces host-wide GPU telemetry collection using NVML, extending the existing job-scoped GPU monitoring. It enumerates all visible GPUs and matches their NVML UUIDs with the configured node resources to properly tag and emit metrics (`in_use`, `utilization`, `memory`, `power`, etc.) to Prometheus and Grafana. The Grafana dashboard deployment script was also significantly refactored for better Prometheus datasource resolution.

Comments:
• [INFO][performance] The comment mentions that a "hung NVML can never stall the loop". While `refreshHostGpuSnapshot()` handles throttling and catches JavaScript exceptions, the underlying Koffi FFI calls in `nvml.ts` are synchronous by default, meaning they execute directly on the main thread. If the NVML C-library function blocks or hangs, it will block the Node.js event loop entirely. If non-blocking resilience against a hanging C-library is strictly required, you might want to look into using Koffi's asynchronous bindings (e.g., `func.async`) to offload the FFI work to a background worker thread.
• [INFO][security] Excellent implementation here. Creating a temporary file with `mktemp`, securely binding it to an `EXIT` trap for cleanup, and streaming the JSON via `stdin` avoids common shell quote-mangling and injection issues that often plague inline `node -e '...'` scripts. This is highly robust and secure.
• [INFO][implementation] Good handling of the C string buffer. Passing a pre-allocated Node.js `Buffer` to the Koffi pointer and correctly searching for the null-terminator with `buf.indexOf(0)` ensures memory safety and avoids reading garbage data.

PR #1477, opened 2026-09-23
PoC service status

### LINKED ISSUE #620: UX: real readiness signals for service startup, and a download button for buckets
Three related gaps in the "I launched something, is it usable yet?" story, plus the missing half of the bucket file manager. Today a user launches an inference service and gets an App URL the moment the container reports Running. For vLLM that's minutes before the engine can answer a request, so the link works but every call 503s.

1. Gate the access link on actual readiness (vLLM / llama is the source of truth)
2. Image download progress bar
3. Download button in storage buckets

andreip136:
Fixes https://github.com/oceanprotocol/nodes-dashboard/issues/620 .

https://claude.ai/artifact/DWCUkk2ZqdFn6XjbCtWGi4?sk=aFZ55n1i89QtMlPtA92bDQ#f3217cd2-2e97 

## Summary by CodeRabbit

* **New Features**
  * Service status now reports readiness for supported workloads, showing whether they are waiting, ready, or failing.
  * Image pulls display download and extraction progress. Supported model downloads show downloaded bytes and file progress when available.
  * Listed service information includes readiness and model-download details when available.
  * Services using unrecognized images continue running without readiness reporting or model-download details.
  * Readiness checks run in the background, so they do not hold up service monitoring.


alexcos20 (commented):

Overall risk: medium

Summary:
Overall, this is an incredibly well-designed PR. The architecture handles complex state synchronizations decoupling gracefully, especially leveraging `BEGIN IMMEDIATE` for SQLite concurrency, managing state transitions efficiently, and adding smart optimizations for Docker I/O. However, there is a critical functional bug regarding how Docker image references are matched against the engine profiles. Since tags and digests are not stripped from the image string prior to matching, the regex will fail on typical production images, breaking the feature completely.

Comments:
• [ERROR][bug] The `matchesImage` profile functions are anchored to the end of the string (`$`), and the JSDoc explicitly states `Matches the image reference (without tag/digest)`. However, `job.image` is passed raw and will likely contain tags or digests in production (e.g., `vllm/vllm-openai:v1.0.0` or `...@sha256:...`). This will cause the regex to fail, completely disabling the new readiness features for production images. Also, it is safer to fall back to `job.containerImage` if `job.image` is absent.

```diff
-  const image = (job.image ?? '').trim()
-  if (!image) return null
-  return SERVICE_ENGINE_PROFILES.find((profile) => profile.matchesImage(image)) ?? null
+  const rawImage = (job.image ?? job.containerImage ?? '').trim()
+  if (!rawImage) return null
+
+  // Strip digest and tag to match the profile contract
+  const withoutDigest = rawImage.split('@')[0]
+  const parts = withoutDigest.split('/')
+  const lastPart = parts.pop() || ''
+  const nameOnly = lastPart.split(':')[0]
+  const baseImage = parts.length > 0 ? `${parts.join('/')}/${nameOnly}` : nameOnly
+
+  return SERVICE_ENGINE_PROFILES.find((profile) => profile.matchesImage(baseImage)) ?? null
```
• [INFO][style] The promise chaining logic (`pullWrites = pullWrites.then(...)`) for sequential database writes is extremely elegant. It perfectly ensures that writes are sequentially resolved without causing SQLite locking issues while avoiding race conditions with state transitions. Great job on this design!
• [INFO][security] Using `redirect: 'manual'` in the `fetch` call for the readiness probe is an excellent security measure. It effectively neutralizes potential SSRF vectors if a compromised or malicious workload attempts to redirect the node's internal fetch request elsewhere. Good attention to detail here!
• [INFO][performance] Aborting the `tarStream` and Docker `archive` socket early when the directory depth exceeds what is needed, rather than draining the whole stream, is a fantastic performance optimization. It saves huge amounts of unnecessary bandwidth and I/O overhead.

alexcos20 (changes requested):

alexcos20 (commented):

Overall risk: low

Summary:
This PR introduces excellent, non-intrusive monitoring for service readiness, Docker image pull progress, and model download progress. The architecture is clean, segregating engine-specific probing logic and abstracting the Docker daemon interactions efficiently. The changes degrade gracefully for unknown engines and handle transient failures well.

Comments:
• [WARNING][bug] If `containerIps` ever contains an IPv6 address (e.g., in IPv6-enabled Docker networks), formatting it directly as `http://${ip}:${containerPort}` will produce an invalid URL like `http://fe80::1:8000/path` which `fetch` will fail to parse. It is safer to bracket the IP if it contains a colon.

```diff
-    if (ip) urls.push(`http://${ip}:${containerPort}${path}`)
+    if (ip) {
+      const host = ip.includes(':') ? `[${ip}]` : ip
+      urls.push(`http://${host}:${containerPort}${path}`)
+    }
```
• [INFO][performance] Chaining database writes in a tight loop via `.then()` works, but if `pullImageRef` emits progress frequently (e.g., every 1s) and the image takes 10 minutes to pull, you could queue ~600 sequential DB write promises. If SQLite writes lag behind the emission rate, the queue could grow indefinitely and delay the final `await pullWrites`. Since `ImagePullTracker` throttles to 1s, it's mostly fine for SQLite, but consider a "drop intermediate updates" approach if the queue gets backed up in the future.
• [INFO][style] Excellent implementation of `patchServiceJobBody`. Using `BEGIN IMMEDIATE` to acquire the write lock upfront and wrapping the read-modify-write cycle prevents race conditions when dealing with concurrent telemetry updates and lifecycle transitions. Very robust pattern.
• [INFO][logic] Falling back to 2 bytes (`BYTES_PER_PARAM[dtype] ?? 2`) for unrecognized datatypes is a smart, pragmatic way to handle unexpected/future HuggingFace parameter metadata without breaking the progress bar completely.

